PR #12, opened 2015-10-22
[IMP] point_of_sale: make POSBox spelling consistent

jorenvo:
@mart-e good idea. I went with POSBox, because that's what we use in the shop (odoo.com/shop) and it also makes the most sense (POS abbrevation, Box camelcase).


mart-e:
merged at 5563869a84c442ece26f453c6710cfc3a339ce9a, thanks

PR #24, opened 2015-10-29
[FIX] css: rendering of inline images

mart-e:
Bootstrap forces the img-responsive images to be a block.
IMHO bootstrap should be killed and burned with fire but them it appears I know
nothing (John Snow) about CSS and making things looks nice so they have probably
their reason to do so.
Anyway, not that I care, let me use inline images for Flupke sake!

Before (:-1: :cry:)
screenshot from 2015-10-29 10-11-24

After (:+1: :tada:  )
screenshot from 2015-10-29 10-12-29


rim-odoo:
image

PR #51, opened 2015-11-03
Enable translations

mart-e:
New command make i18nhtml to add parameter LANG=...

e.g. `make LANG=fr i18nhtml`

Problem: need to make one make per language.


KangOl:
seems correct

PR #100, opened 2015-11-19
[IMP] PoS_loyalty: - improving description type of rewards

lla-odoo:
Improvements in PoS_loyalty with printscreen for rewards
Adding how to add a customer from a barcode in how to invoice


tivisse:
Landed in 9.0. Thank you.

PR #129, opened 2015-12-16
[ADD] point_of_sale: belgian fiscal data module documentation

jorenvo:


point_of_sale/belgian_fdm/setup.rst
```diff
@@ -0,0 +1,82 @@
...
+
+Required hardware
+=================
+
+* A government certified `Fiscal Data Module
+  <http://www.systemedecaisseenregistreuse.be/systemes-certifies#FDM%20certifiés>`_
+  per POS, all of them should work, but the Cleancash SC-B is
+  recommended, you will also need:
+
+  * Serial null modem cable per FDM (`example
```
mart-e: somehow sphinx complains for `Duplicate explicit target name: "example".` not sure why

point_of_sale/belgian_fdm/setup.rst
```diff
@@ -0,0 +1,82 @@
...
+government-certified device called a Fiscal Data Module (also known as
+a blackbox). This device works together with the POS application and
+logs certain transactions. On top of that, the used POS application
+must also be certified by the government and must adhere to strict
+standards specified by them. `Odoo 9 is a certified application
+<http://www.systemedecaisseenregistreuse.be/systemes-certifies>`_. More
+information concerning the Fiscal Data Module can be found on `the
+official website <http://www.systemedecaisseenregistreuse.be/>`_.
+
+**Currently our government certificate is only valid for Odoo
```
mart-e: probably add a date on that statement, in case somebody keeps a copy of an old doc somewhere

point_of_sale/belgian_fdm/setup.rst
```diff
@@ -0,0 +1,82 @@
+===================================================
+Setting up the Fiscal Data Module with the Odoo POS
+===================================================
+
+Introduction
+============
+
+The Belgian government requires certain businesses to use a
+government-certified device called a Fiscal Data Module (also known as
```
mart-e: bold for "Fiscal Data Module" and "blackbox"

point_of_sale/belgian_fdm/setup.rst
```diff
@@ -0,0 +1,82 @@
...
+the *Hardware status page* via the POSBox homepage.
+
+.. image:: media/posbox_fdm_hardware_status.png
+    :align: center
+
+Odoo
+----
+
+An Odoo POS is turned into a certified POS by installing the Belgian
+Registered Cash Register module (technical name:
```
mart-e: name of the module in bold (and we tend to use "app" instead of "module")

point_of_sale/belgian_fdm/setup.rst
```diff
@@ -0,0 +1,82 @@
...
+``pos_blackbox_be``). This installation cannot be undone. After this,
```
mart-e: Can you elaborate why can not be undone? What does it implies if I try to uninstall the app?

point_of_sale/belgian_fdm/setup.rst
tivisse: Cannot reproduce the issue.

jorenvo: I already fixed it. It was because you can't have two 'Named hyperlink references' with the same name. I turned them into 'Anonymous hyperlink references' which is the same thing but with two underscores at the end.

jorenvo:
Thanks for the quick review Martin. Fixed according to your feedback.

tivisse:
Landed in 9.0 at https://github.com/odoo/documentation-user/commit/19e1eceef0f40dbe94e4af865eb48d9b487ca94e

Thanks @jorenvo

mart-e:
@jorenvo no problem, thanks to write doc pages :-)
The "cannot be undone" this is a legal or technical restriction? I mean if I just want to test the blackbox (or I just click on the wrong module) will it be stuck on my db or can I still remove it?

jorenvo:
@mart-e it's legal restriction. It is illegal to turn a certified POS into a non-certified one. So we cannot allow uninstallation of the module, even for testing purposes. If you click on the wrong module then I don't know what to do. The 'correct' legal way would be to create a new db with a different uid and migrate all your data to that.

PR #132, opened 2016-01-04
doc sale_ebay

mart-e:


crm/ebay/manage.rst
```diff
@@ -0,0 +1,31 @@
+======================
+How to list a product?
+======================
+
+Listing without variation
+=========================
+
+In order to list a product, you need to check the **use eBay** field on a product form. Then the eBay tab will be available.
```
mart-e: We usually avoid to start sentence by "then" or such not very formal keywords.

mart-e:
You can use the tool `pngquant` to reduce your image sizes without too much losses.

crm/ebay/manage.rst
```diff
@@ -0,0 +1,31 @@
...
+
+.. image:: ./media/ebay_manage_1.png
+  :align: center
+
+If you check the **Use Stock Quantity** field, the quantity sets on eBay will be the Odoo **Forecast Quantity**.
```
mart-e: we try to avoid using "you" when possible. So in this case, you rephrase "When the **Use Stock Quantity** field is  checked..."

crm/ebay/manage.rst
```diff
@@ -0,0 +1,31 @@
...
+
+The **Description Template** allows you to use templates for your listings. The default template does nothing except using the **eBay Description** field of the product. You can use html inside the **Description Template** and in the **eBay Descritpion**.
```
mart-e: "does nothing except" is not very positive, "does only" probably better ;-)

crm/ebay/setup.rst
```diff
@@ -0,0 +1,28 @@
+==============================
+How to configure eBay in Odoo?
+==============================
+
+Create eBay tokens
+==================
+
+In order to create your tokens, you need to create a developer account `there <https://go.developer.ebay.com/>`_.
```
mart-e: instead of "there" or "here" link, you should use one that describe the page you are going to (e.g. "developer ebay website")

mart-e:
Thanks merged in 9.0

PR #139, opened 2016-01-27
Add Google Analytics tutorial

mge-odoo:
This tutorial is currently in `res_config` of `website_version` under the checkbox "Show me tutorial".
The user documentation is certainly a better place to show this kind of stuff.
This tutorial will also be used in another module later (in master) so instead of copy/pasting it, we will just refer to this user documentation.


mge-odoo:
Should I do something for translations @mart-e ?

PR #149, opened 2016-03-08
Remove misleading statement

cst-odoo:
Removed the sentence from line 39:

**Don't be afraid of messing up the data**,
it's very easy to duplicate your database, clean it or start a new one.

Reason: It is possible to create records that are difficult to reverse, such as accounting entries and delivery records.

RSH was consulted and approved the change.


mart-e:
Thanks

PR #150, opened 2016-03-16
Improve inventory documentation

Yenthe666:
Without adding the information about marking a transfer as TO DO you will not see the icon show up and this is really confusing.


mart-e:
Thanks, I merged it after a few changes (split line, remove uppercase, regenetate translations)

PR #155, opened 2016-03-22
[IMP] point_of_sale: reflect new certified POS approach

jorenvo:
The certification restriction is gone. Some things have changed in the
process because of this. The main change is that we'll now provide
registered POSBoxes instead of registering Odoo instances.


mart-e:
Thanks, merged at 7f0036ca2f5ac626bcac30eabd512bace3ab87ab

PR #164, opened 2016-04-26
[ADD] base_import FAQ

mart-e:
see #157 for comments
all responsability on @csnauwaert


mart-e:
damn it wrong target !

PR #170, opened 2016-06-16
Corrected udev rule name

asantos82:
The udev rule must be 99-usbusers.rules instead of 99-usbusers.rule because otherwise it would not work


tivisse:
Hello @asantos82 

This issue has already been fixed in https://github.com/odoo/documentation-user/commit/fc6906073b6616d0d4f7899c50563141c114e1d0

Thanks for the feedback.

PR #172, opened 2016-06-20
[IMP] pos: certification and on-premise

mart-e:
Specifiy explicitely that the on-premise module is for Enterprise customers
and that bying a POSbox is not enough.


jorenvo:
Looks good to me

PR #184, opened 2016-09-03
[IMP] grammar

fhe-odoo:
The en_US police has your back.

### DIFF
--- a/getting_started/documentation.rst
+++ b/getting_started/documentation.rst
@@ -111,7 +111,7 @@ must:
     departments. This person must have the support, or be part, of
     the top management.
 
-Integrate 95% of your business, not 100%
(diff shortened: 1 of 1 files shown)


fhe-odoo:
@mart-e Can you check this.

mart-e:
Sure! Thanks for the changes.
Is the 95 -> 90% change necessary?
I don't have problems changing it but it will break the translations so I want to avoid if not necessary.

fhe-odoo:
Up to you. Not mandatory.

mart-e:
Ok, so I will just keep the part with grammar errors.
Merged at 243560be77db5ad842ee3b2a8c498b8ab8a686a4

PR #187, opened 2016-09-30
[ADD] Shipping DHL: How to get dhl credentials

lla-odoo:
As it was already asked why the customer receive a "wrong SiteID" message I decided to create a page to explain of to get dhl credentials. Because it's because the customer doesn't have correct credentials

Issue: 685598


tivisse:
Landed in 9.0 at https://github.com/odoo/documentation-user/commit/e3bb738337bb2ad501468072e660b587b123487f and forward ported in 10.0

PR #201, opened 2017-01-06
[FIX] Typo into "Set up stages probabilities"

ebirbe:
There is a typo in the **Won** stage description.


tivisse:
Hello @ebirbe 

Typo fixed in 10.0 at https://github.com/odoo/documentation-user/commit/ba5511d7357bb5ebbe5956b93c02ff5c1689d2dc

Thanks for the feedback.

PR #203, opened 2017-02-02
[FIX] point_of_sale: correct reprint documentation

jorenvo:
The documentation incorrectly implied that the Reprinting related
configuration affects the 'reprint' button in the backend on
pos.order. This button is always available in Odoo 9.

Instead, the Reprinting feature (which is implemented by pos_reprint)
allows a cashier to reprint the last printed receipt from the POS
frontend.

opw-703848


jorenvo:
@mart-e thanks!

PR #204, opened 2017-02-10
Update setup.rst

multiwebinc:
I believe this makes the documentation clearer. I was wondering what to do with the image file, but from what I gather, that's only needed when being installed on a raspberry pi. See https://github.com/odoo/odoo/issues/15420

### DIFF
--- a/point_of_sale/overview/setup.rst
+++ b/point_of_sale/overview/setup.rst
@@ -160,25 +160,14 @@ POSBoxless Guide (advanced)
 
 .. image:: media/posboxless_setup.png
 
-If you are running your Point of Sale on a Debian-based Linux
-distribution, you do not need the POSBox as you can run its software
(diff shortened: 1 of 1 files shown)


Yenthe666:
@mart-e documentation PR for you.

mart-e:
Hello,

Thanks for the corrections.
Actually, I think we should not talk at all about the image build in the posbox-less guide. It is a "make your own" section that could be inside the "POSBox Technical Documentation" chapter.

[The image](https://www.odoo.com/documentation/user/10.0/_images/posboxless_setup.png) from the posbox-less chapter is clear: there is no raspberry pi involved.

tivisse:
@multiwebinc 

According to @mart-e 's answer I'm closing the PR. 

Don't hesitate if you have any question/request about this.

Have a nice day,

Yannick.

PR #211, opened 2017-03-08
fix in the negative reminder delay example

manitra:
If we have -3 as reminder delay and the reminder occures on the 17th, the correct due date is 20 and not 30


tivisse:
Hello @manitra 

The example has been fixed in 10.0 at the revision https://github.com/odoo/documentation-user/commit/beb156ed50ed374d2762a739ccfe78ae0ea5ba54

Thanks for the feedback. Have a nice day,

Yannick

PR #216, opened 2017-05-05
Fix error with order used instead of invoice

zeroheure:
In chapter `eCommerce Order --> Invoice` : the sentence  «An eCommerce order will also trigger the creation of the order when it is fully paid.»  means obviously 
«An eCommerce order will also trigger the creation of the invoice when it is fully paid.»

Same error in 11 version


tivisse:
Hello @zeroheure 

Thanks for the feedback. The issue as been fixed at https://github.com/odoo/documentation-user/commit/0c354a19c8c5c3b6bf0ed2e8a637576410b22af1

Have a nice day,

Yannick.

PR #223, opened 2017-05-16
Update Credit Card Documentation

lap-odoo:
From V10, there is a new account type "Credit Card". That's the one that should be used for the credit card account. Information added in the documentation.


tivisse:
Fixed in 10.0 at https://github.com/odoo/documentation-user/commit/04e4ced6b4d7cb0a2475c7d85e85fb3550d68720

PR #244, opened 2017-10-11
Mention single-device only!!!

exidoncom:
This document should state, that it only works using the same device. No multi-device sync at all!


mart-e:
Hello,

I guess you have been disappointed because somebody did not worked as you expected but I am not sure what.
Could you please explain what you were trying to achieve. I agree about adding warning but it should be clarified.

PR #257, opened 2017-12-23
[IMP] vat-anti-fraud: better guidelines to install/update the module

fgi-odoo:


fgi-odoo:
@tivisse quick update of anti-fraud fr doc to help users install or update the module. Could you check to update terms to translate in transifex afterwards? That way we can update the fr translation asap, still providing wrong links. Thanks!

tivisse:
Fixed at https://github.com/odoo/documentation-user/commit/e5e9a476de6725b695c1dabbb619a11d3680cc9e

PR #276, opened 2018-04-11
[9.0] FIX price included => tax included and so on

zeroheure:
price included => tax included
price excluded => tax excluded


mart-e:
Thanks, merged. We will port it to upper versions, no need to make several pull requests

PR #296, opened 2018-09-19
[IMP] getting_started: Replace old doc

jubodoo:
Replacing the old doc by new ones for QS methodology.


jubodoo:
Just had a small issue, it's not in any toctree, but the previous one (documentation.rst) was also not included, so I thought was okay to leave as it. Can you check YTI? Thanks.

PR #298, opened 2018-10-02
[ADD] accounting: add colombian localization docs

jorenvo:
Text by REN.


jorenvo:
@tivisse hi Yannick, not sure if you're the responsible for documentation, but here are some docs for the new v12 Colombian electronic invoicing module.

tivisse:
The PR was good. Thanks.

Just a note, do not forget to optimize your pictures, for example with pngquant.

Have a nice day,

Yannick.

jorenvo:
Oh right, that is a good idea. Thanks for merging and optimizing the pictures!

PR #299, opened 2018-10-04
[FIX] update urls to v12

mreficent:
Point old urls to v12.


mreficent:
@mart-e

mart-e:
Merged at ade8bf3, thanks

PR #306, opened 2018-11-16
[ADD] legal: add license section for v12

fhe-odoo:


fhe-odoo:
@odony this should be straightforward. It's exactly the same way as you did it last year.

odony:
Thanks!

PR #307, opened 2018-11-19
[IMP] unsplash: new guidelines negociated with Unsplash

rdeodoo:
Unsplash CEO asked us to stop exposing our key client side (even if it was
only in the browser network and allowed by Unsplash API team leader).

Also, all our Odoo instance, on premise or not should share the same
production key, which is not possible for us obviously.

After multiple exchange, it has been negociated that:
1. Our non saas user won't be able to apply for a production key and should
   prefix they Unsplash application name by 'Odoo:'. Their key will remain in
   test mode.
   Indeed, Unsplash won't be able to review every Odoo application and ensure
   it respects the API terms and it is a real application.
2. Our saas user will query Unsplash server side so the key remain hidden.
3. The documentation should explain it all

### DIFF
--- a/general/unsplash/unsplash_access_key.rst
+++ b/general/unsplash/unsplash_access_key.rst
@@ -2,6 +2,12 @@
 How to generate an Unsplash access key
 =======================================================
 
+.. tip::
+  **As a SaaS user**, Unsplash is ready to use. You won't need to follow this guide to set up Unsplash informations as you will use our own Odoo Unsplash key in a transparent way.
+
+Generate an Unsplash access key for **non-Saas** users
+======================================================
+
 - Create an account on `Unsplash.com <https://unsplash.com/join>`_.
 
 - Go to your `applications dashboard <https://unsplash.com/oauth/applications>`_ and click on **New Application**.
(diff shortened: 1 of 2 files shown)


rdeodoo:
@jubodoo Not sure if we should merge it already or wait the changes in production
It will probably need one or two day before Odoo's code reflect this doc changes

jubodoo:
Sure we can wait wednesday or friday to merge it if you want.

*Julien Bertrand* – Product Marketing Officer
T. +32497353485 jub@odoo.com
Odoo - Rue des Bourlottes 9, 1367 Ramillies - Belgium

JKE-be:
please can you merge now. 
Thx

JKE-be:
Just to be sure... Now in the sense : not postpone in several day ;)
Take all the time that you need (k)

But since it will be merged this week and that doc is build only once by day... it is not critical to have the doc 12h before in stead to 12h after ;)

jubodoo:
Same for me, so, merged. :)

rdeodoo:
Linked to https://github.com/odoo/odoo/pull/28804 and https://github.com/odoo/internal/pull/500

PR #318, opened 2019-01-17
[FIX] Inventory: remove wrong pdf

lap-odoo:
The pdf with the barcode actions is not up-to-date, it's removed here. Now it's easily accessible from the inventory settings so it's not necessary to have in the documentation anymore (and it's easier to avoid any issue across versions).


lap-odoo:
@tivisse Could you please check this ? Thanks

rim-odoo:
Merged https://github.com/odoo/documentation-user/commit/4f421f45bbcfb970b0d4ca45e57b45358ea0f4a2

PR #329, opened 2019-02-22
[ADD] auth: Azure OAuth is not compatible

jev-odoo:


general/auth/azure.rst
```diff
@@ -0,0 +1,5 @@
+====================================
+Azure Oauth not compatible with Odoo
```
rim-odoo: OAuth ;)

general/auth/azure.rst
```diff
@@ -0,0 +1,5 @@
...
+====================================
+
+Microsoft Azure Oauth identification is NOT compatible with Odoo at the moment.
```
rim-odoo: ... due to specific requirements in Azure's OAuth implementation.

general/auth/azure.rst
rim-odoo: with a capital A

jev-odoo: Sorry

rim-odoo: no problem :D

mart-e:
https://www.odoo.com/documentation/user/14.0/general.html
Shouldn't we call this page "Azure OAuth" instead? It's a bit confusing on the index page.
Especially the fact that we don't have a "OAuth" documentation :D

rim-odoo:
yes, you're right, you can do it @mart-e

jev-odoo:
@rim-odoo @mart-e 
If I remember well, it was the first entry about an OAuth documentation.
It contains only Azure at the moment, but we could(should) improve it with other OAuth topics ...
Should we then create 1 page per OAuth doc, or one OAuth page with multiple OAuth topics?

mart-e:
It depends if somebody is working on the rest of documenting general OAuth flow (cc @AntoineVDV ?). If not, we can simply rename the page from now.
Got a bit of "Yeah a doc page for OAuth :tada:  ... hmmm or not :disappointed:  " moment

rim-odoo:
> Nope, there is nothing planned on that topic. And I think that we can classify this attempt of creating an incentive in AL for new features as failed

Not at all.

Just an attempt to document a recurring support question

PR #332, opened 2019-03-01
Explain overlay fs features

rim-odoo:
opw-1945086


beledouxdenis:
This is already partly noted there:
https://www.odoo.com/documentation/user/11.0/odoo_sh/getting_started/online-editor.html#edit-the-source-code

Maybe the only thing missing would be the part
```
Please note that these changes won't be propagated to a new build, you must commit them in your source code if you want to make them persist.
```
which can be added in that place.

PR #335, opened 2019-03-08
[FIX] base: Correct state in res_partner template

Abridbus:
State from the template was 'CA' for California - not recognized on import
Changed it to 'California' to match with the db state

Already changed on the V12 template of the Contact app

opw 1943904


rim-odoo:
Not relevant anymore since https://github.com/odoo/odoo/commit/215c4a6b5d71c56315f09ce388a8316b76a3a23e

PR #336, opened 2019-03-20
[IMP] Restructure of the tutorial in order to put useful todos at the…

fgi-odoo:
… top


fgi-odoo:
@JKE-be @rdeodoo Coucou! I updated the SEO doc. I moved most important sections at the top (what the user have to do). Would be great if you could review it and add or suggest missing information. Thanks!

PR #341, opened 2019-03-29
[FIX] accounting: Fix mexican view due to a mistake with special characters

CarmenMiranda:
In the mexican documentation we have an errror like the one shown bellow: 
Mexico: Documentation Error
([UNKNOWN NODE title_reference])

In order to make those errors disappear and that the correct information is displayed, what this PR do, is change some special characters.

### DIFF
--- a/accounting/localizations/mexico.rst
+++ b/accounting/localizations/mexico.rst
@@ -271,34 +271,34 @@ process in Odoo, this considerations to understand the behavior are important.
 
    **1.1. How can I generate an invoice with payment term `PUE`?**
 
-   `According to the SAT documentation`_ a payment is classified as `PUE` if
+   `According to the SAT documentation`_ a payment is classified as ``PUE`` if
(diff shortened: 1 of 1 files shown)


luistorresm:
@moylop260

Could you approve this PR, please?

Based on the last comment [here](https://github.com/odoo/documentation-user/pull/315)

Regards

xmo-odoo:
robodoo r+

jga-odoo:
@mart-e @xmo-odoo Hello; are these changes automatically merged with later versions? The v11 page was updated, but not the v12 page, please see attached screenshots. Also, this page is not very useful in English, as it is mainly for Mexican customers; how is the translated content handled? Directly in Github, or over Transifex? We are currently in talks with Vauxoo to share duties related to maintaining this page through all its versions and languages.

v11

v12

mart-e:
@jga-odoo I have just forward ported it in 12.0, it will be in the next build

PR #346, opened 2019-04-15
[FIX] Push Notif OCN - Mobile

Abridbus:
No API key needed anymore. Cleaning the doc regarding this.


mobile/firebase.rst
```diff
@@ -24,7 +24,7 @@ Firebase Settings
 Create a new project
```
adr-odoo: In 12.0, it's not required anymore. You can remove this section.

mobile/firebase.rst
```diff
@@ -56,24 +56,11 @@ not work.
...
-that you need to set in Odoo General Settings.
-
-.. image:: media/firebase05.png
-   :align: center
 
 Settings in Odoo
 ================
 
-Simply paste the API key and the Sender ID from Cloud Messaging.
+Simply select the Odoo Cloud Notifications (OCN) in the General Settings.
```
adr-odoo: I think you can't change it in 12.0, you can remove it too.

PR #349, opened 2019-05-02
show where each sales settings is explained

zeroheure:
Add a note to show where each sales settings option is explained.  And minor edits for clarity.


rim-odoo:
Thanks @zeroheure for your contribution.

PR #354, opened 2019-05-12
Update budget.rst

ray-odoo:
https://www.odoo.com/forum/help-1/question/i-don-t-understand-the-purpose-of-budgetary-positions-why-do-we-use-them-149442

### DIFF
--- a/accounting/others/adviser/budget.rst
+++ b/accounting/others/adviser/budget.rst
@@ -5,9 +5,12 @@ How to manage a financial budget?
 Overview
 ========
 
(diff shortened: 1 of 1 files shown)


rim-odoo:
Thanks @ray-odoo

Himilo2021:
Can you setup  in Odoo a budgetary control where the PO amount is validated against available budget for each account in the PO line. And if available budget is less than the PO amount then the PO will fail to validate.

Thanks for your feedback

PR #366, opened 2019-07-23
[FIX] Grammar and punctuation in my_odoo_portal.rst

nKlapatch:
I added some punctuation to the documentation about the Odoo portal as well as some grammar improvements.


mart-e:
Thanks !

PR #391, opened 2019-09-04
[IMP] odoo_sh: recurring question about crons

rim-odoo:
opw-[2058254](https://www.odoo.com/web?debug#id=2058254&model=project.task&view_type=form&menu_id=5200)

### DIFF
--- a/odoo_sh/advanced/frequent_technical_questions.rst
+++ b/odoo_sh/advanced/frequent_technical_questions.rst
@@ -16,6 +16,21 @@ This is due to the fact that there might be multiple customers on the same serve
 .. warning::
     Do not expect any automated action to be run more often than every 5 min.
 
+Are there "best practices" regarding automated actions?
+-------------------------------------------------------
+
+**Odoo.sh always limits the execution time of automated actions (*aka* crons).**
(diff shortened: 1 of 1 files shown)


odoo_sh/advanced/frequent_technical_questions.rst
```diff
@@ -6,6 +6,16 @@
 Frequent Technical Questions
 ==================================
 
+"Automated actions do not run at the exact time they were expected"
+-------------------------------------------------------------------
+
+On the Odoo.sh platform, we cannot guarantee an exact running time for automated actions.
+
+This is due to the fact that there might be multiple customers on the same server, and we must guarantee a fair share of the server for every customer. Automated actions are therefore implemented slightly differently than on a regular Odoo server, and are run on a *best effort* policy.
+
```
amigrave: we might take the opportunity to add a short message explaining best practice for crons such as batches explicitly commited written in an idempotent way (maybe it's already present somewhere else in the docs in such case a link is enouf)

rim-odoo: I'll add it, good idea!

PR #393, opened 2019-09-05
[FIX]  accounts unbalanced in continental accounting for stock transa…

jpp-odoo:
…ctions

In the documentation of continental accounting for stock transactions.
In the perpetual inventory valuation, when we choose the option 'Vendor
Invoice (PO €48, Invoice €50)', the purchase goods must be €50 and not
€48 as is the case for the moment.

opw-2052725


jpp-odoo:
@rim-odoo This Fix must be fowardported. Thanks.

PR #394, opened 2019-09-06
[FIX] Remove screenshot with non-visible field

JPML94:
The field carrier_price is no longer viewable in the delivery order by default and the screenshot could bring confusion to people that are following the documentation.


rim-odoo:
Merged at https://github.com/odoo/documentation-user/commit/49434c2dae5be334799c3cdd0a0cd6110e83834f

PR #406, opened 2019-09-18
Adapt Build Process to new structure

24601:
The new layout of directories and file structure was freaking out the sphinx html build target, tweaks to fix this.


24601:
Sorry, PR against the wrong repo

PR #413, opened 2019-10-03
[FIX] v13 urls

mreficent:
Outdated references

Point old urls to v13.


mreficent:
ping: @mart-e

mart-e:
Thanks, merged at efcb1a4

PR #450, opened 2019-10-19
Update taxcloud.rst

taxcloud:
Minor updates regarding use of street address (not just zip code) and TIC description.


rim-odoo:
Hello @taxcloud and thank you for your contribution.
Does this change also apply to Odoo 110 and 12.0?
Thanks

PR #455, opened 2019-10-28
'[FIX] Update Deactivating Users Section'

jop-odoo:
-updated screenshots and text to match version 13


rim-odoo:
Thanks @jop-odoo, this should be published tomorrow.

PR #458, opened 2019-11-04
[IMP] ecommerce: authorize.net setup

dbo-odoo:
Changes needed since there are now more keys to set up.


dbo-odoo:
TODO: specifiy that response URLs need to be set in authorize.net backend

PR #460, opened 2019-11-07
[IMP] taxcloud: coupon nightmare

dbo-odoo:
to merge once module is validated + backport 11/12


len-odoo:
Documentation???
We'll review it by the end, for now the information is a bit scattered in the commit message/PR/manifest...
But your first draft seems already pretty good :+1: 

(Todo: maybe note that without the extra modules, the discount are simply not applied?)

PR #486, opened 2019-12-09
Inventory replenishment doesn't change the uom

obayit:
The statement:
    "When doing a replenishment via the Replenish button on the product form, you have the possibility to change the unit of measure."
Implies that when a replenishment is done the purchase uom will be changed when actually it just creates a stock.move with a calculated amount based on the product's uom.

### DIFF
--- a/inventory/management/products/uom.rst
+++ b/inventory/management/products/uom.rst
@@ -82,7 +82,7 @@ Replenishment
 -------------
 
 When doing a replenishment via the *Replenish* button on the product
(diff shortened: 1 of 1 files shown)


obayit:
Actual replenishment code:
`
def launch_replenishment(self):
        uom_reference = self.product_id.uom_id
        self.quantity = self.product_uom_id._compute_quantity(self.quantity, uom_reference)
        try:
        .... creating the procurement.group and the moves ..etc ....
`

PR #487, opened 2019-12-10
[IMP] SEO stuff

mart-e:
- [x] add `links` with `hreflang`
- [x] add `canonical`
- [ ]  meta tags `description` `keywords`
- [ ] sitemap
- [ ] backport
- [ ] technical doc

Task id: 2121586
Pad: https://pad.odoo.com/p/r.83090664d0c2a9f20ff54edc93e80d70

### DIFF
--- a/_extensions/odoo/layout.html
+++ b/_extensions/odoo/layout.html
@@ -25,6 +25,13 @@
   {{ super() }}
 {%- endblock -%}
 
+{%- block linktags -%}
+  {% for code, url in language_codes %}
+    <link rel="alternate" hreflang="{{ code }}" href="{{ url }}" />
+  {%- endfor %}
+  {{ super() }}
+{%- endblock -%}
+
 {%- block sidebar1 -%}{%- endblock -%}
 {%- block sidebar2 -%}{%- endblock -%}
 {%- block relbar1 -%}{%- endblock -%}
--- a/_extensions/odoo/layout.html
+++ b/_extensions/odoo/layout.html
@@ -29,6 +29,7 @@
   {% for code, url in language_codes %}
     <link rel="alternate" hreflang="{{ code }}" href="{{ url }}" />
   {%- endfor %}
+    <link rel="canonical" href="{{ canonical }}" />
   {{ super() }}
 {%- endblock -%}
 
--- a/conf.py
+++ b/conf.py
@@ -337,7 +337,9 @@ def setup(app):
     app.add_javascript('coa-valuation-continental.js')
     app.add_javascript('coa-valuation-anglo-saxon.js')
 
+    app.connect('html-page-context', canonicalize)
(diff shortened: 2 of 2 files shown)


_extensions/odoo/layout.html
```diff
@@ -25,6 +25,13 @@
   {{ super() }}
 {%- endblock -%}
 
+{%- block linktags -%}
+{% for code, url in language_codes %}
+  <link rel="alternate" hreflang="{{ code }}" href="{{ url }}" />{% endfor %}
+  <link rel="canonical" href="{{ canonical_rel }}" />
```
xmo-odoo: endfor should probably be on the newline [possibly using %- / -% to suppress whitespace](https://jinja.palletsprojects.com/en/2.10.x/templates/#whitespace-control), cf the block/endblock statements which make use of this feature though possibly out of cargo-culting rather than understanding)

And canonical should be un-intended to make it clear it's not part of the loop, also possibly *before* the loop?

conf.py
```diff
@@ -403,9 +405,25 @@ def localize(app, pagename, templatename, context, doctree):
         for la in app.config.languages.split(',')
         if la != current_lang
     ]
+    context['language_codes'] = [
+        (la.split('_')[0] if la != 'en' else 'x-default', _build_url(app.config.canonical_root, (la != 'en' and la or ''), pagename))
+        for la in app.config.languages.split(',')
+    ]
```
xmo-odoo: Could be worth merging languages and languages code in a single traditional loop? Not really sure:
```python
context['languages'] = []
context['language_codes'] = []
for la in app.config.languages.split(','):
    url = _build_url(app.config.canonical_root, (la != 'en' and la or ''), pagename)
    context['langage_codes'].append((la.split('_')[0] if la != 'en' else 'x_default', url))
    if la != current_lang:
        context['langages'].append((LANGAGES.get(la, la.upper()), url)
```

_extensions/odoo/layout.html
```diff
@@ -25,6 +25,13 @@
   {{ super() }}
 {%- endblock -%}
 
+{%- block linktags -%}
+{% for code, url in language_codes %}
+  <link rel="alternate" hreflang="{{ code }}" href="{{ url }}" />{% endfor %}
+  <link rel="canonical" href="{{ canonical }}" />
```
xmo-odoo: endfor should probably be on the newline [possibly using %- / -% to suppress whitespace](https://jinja.palletsprojects.com/en/2.10.x/templates/#whitespace-control), cf the block/endblock statements which make use of this feature though possibly out of cargo-culting rather than understanding)

And canonical should be un-intended to make it clear it's not part of the loop, also possibly *before* the loop?

conf.py
mart-e: not sure two loops improve the readability IMHO, we only have a few items anyway

PR #491, opened 2019-12-12
[REM] Removal of mention of Storno accounting

RamblingFiddler:
There was a mention that Odoo supports Storno accounting since v9.0.
It is not true and this change must be forward ported on all branches.

Thanks


AntoineVDV:
@mart-e branch 9.0 is write-protected (as well as 10.0 I guess) so I let you handle the merge and the forward-ports ;)

PR #496, opened 2019-12-17
[ADD] : l10n_ar Localization user documentation

ren-odoo:
Functional documentation for the Argentinean localization, it includes:  
- Reference and explanation of new models. 
- Configuration
- General workflow
- Testing guidance.


ren-odoo:
> Hi @ren-odoo, I made some minor changes to the structure of the doc (not the content) to fix issues so make sure to update your local branch with this one if you still have changes to make.
> 
> Also, I was wondering if it was intended to leave some parts of the text in Spanish although the rest of it is written in English ? For instance : "Invoice for a customer IVA Responsable Inscripto, prefix A"

Hi @AntoineVDV the parts in Spanish are related to official government terms in the case you mentioned in the example is the taxpayer type, I can do a literal translation but any l10n_ar understand better if we use the Spanish term. 

Regarding the change you made in the structure is it something I need to take into consideration for future PR?

Thank you.

AntoineVDV:
> Hi @AntoineVDV the parts in Spanish are related to official government terms in the case you mentioned in the example is the taxpayer type, I can do a literal translation but any l10n_ar understand better if we use the Spanish term.

Alright, let's keep it that way then :)
I was asking because we plan to clean a bit the l10n documentations. In particular, we should only keep pages written in English and let the translations from Transifex handle the other languages, so that each documentation displays in the right language when the latter is changed from the menu. This means no more pair of pages like `columbia` and `columbia_ES`.
In your case, it's already good as it is since it's written in English and the images are in Spanish (which is what we want), so there is nothing to change ;)

> Regarding the change you made in the structure is it something I need to take into consideration for future PR?

No. If you follow the Google Doc (Build & Push the doc) that Martin sent you, you will be prompted to delete your local branch or at least no reuse it. As my changes are already merged with yours (I wrote them in your commit), you will see them in the main branch `13.0` as soon as this PR gets merged.
If you want to make extra sure that you are up-to-date with this branch, or if you want to keep working on it, you can use the following command : `git checkout 13.0-l10n_ar-ren && git fetch origin && git reset origin/13.0-l10n_ar-ren`. This will bring you local branch up-to-date with the remote one, in which my changes are present.

PR #499, opened 2019-12-26
[IMP] conf: add sitemap generation

mart-e:
Well need to resolve issue about jdillard/sphinx-sitemap#22

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -33,6 +33,7 @@ extensions = [
     'odoo',
     'demo_link',
     'embedded_video',
+    'sphinx_sitemap'
     'github_link',
     'html_domain',
     'redirects',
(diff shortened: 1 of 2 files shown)


xmo-odoo:
@mart-e aside from the conflicts is there anything blocking this?

mart-e:
Not really, just need to install the dependency

Feyensv:
Still needed/useful ? :thinking:

xmo-odoo:
As much as it ever was: sitemaps are useful to SEO.

mart-e:
Restarted at  #1067

mart-e:
Is a sitemap really useful anyway? The documentation is correctly indexed, it's not like the forum where you need to point to every existing question, here every page is accessible from the index through the navbar and crawled by google.

PR #500, opened 2019-12-27
[IMP] sale_amazon: add information about compliance with Amazon's sec…

AntoineVDV:
…urity standards

Depending on whether the seller made a request to get access to Personally Identifiable Information of its customers, and on other factors, Amazon may require him to fill out a second form about his data protection policy.
This commit makes it clear that compliance with Amazon's security standards is not guaranteed if the seller applying for Amazon MWS access keys is hosted on Odoo (SaaS or .sh).

Additionally, it add a section in which the questions asked to Amazon sellers are listed and answered for the Odoo Online and Odoo.sh platforms.

### DIFF
--- a/sales/sale_amazon/apply.rst
+++ b/sales/sale_amazon/apply.rst
@@ -1,0 +1,146 @@
+================================
+Apply for Amazon MWS Access Keys
+================================
+
+.. _amazon/developer-form:
+
+Submit the Amazon MWS Developer Registration and Assessment Form
+================================================================
+
+In order to synchronize your Amazon orders with Odoo, Amazon MWS access keys are required.
+They can be obtained by submitting the **Amazon MWS Developer Registration and Assessment form** to
+register as a developer. Once recognized by Amazon as a developer (i.e. you make use of an
+application connecting to MWS), you will be granted Amazon MWS access keys.
+
+First, visit the `Amazon Marketplace Web Service documentation
+<http://docs.developer.amazonservices.com/en_US/dev_guide/DG_Registering.html>`_ and follow the
+instructions to register as a developer. Take care to choose the form "I represent a seller
+organization integrating with Amazon MWS for its own selling account only.".
+
+Fill out the *Developer Registration and Assessment form* as suggested below and provide your own
+contact information in the **Developer contact information** section. In the **Business use
+information** section, select the correct region of your seller account. For the other sections,
+adapt your responses in accordance with your business case.
+
+Give a particular attention to **Merchant Fulfilled Shipping**. It should only be checked if you
+ship your products yourself. You should uncheck it if you sell exclusively with the *Fulfillment by
+Amazon* service. Please note that requesting this function is not recommended if you are hosted on
+*Odoo Online* or *Odoo.sh* as the additional security requirements asked by Amazon may not be met by
+Odoo.
+
+.. warning::
+   Depending on several factors (your region, whether you checked the **Merchant Fulfilled
+   Shipping** function (i.e. you request access to Personally Identifiable Information of your
+   customers), etc.), Amazon may request you to fill out a second form before granting you MWS
+   access keys. As that form depends on the data protection policy of the region of your seller
+   account (e.g. GDPR in Europe), we cannot provide you with a pre-filled form. Instead, the answers
+   of questions related to Odoo are listed in the `Answer the Additional Form`_ section.
+
+.. tip::
+   If you need assistance for your application for Amazon MWS access keys, `submit a support ticket
+   to Odoo <https://www.odoo.com/help>`_.
+
+.. image:: ./media/dev_form.png
+
+Answer the Additional Form
+==========================
+
+This section lists all questions asked by Amazon in additional forms. The answers are tailored for
+*Odoo Online* and *Odoo.sh*. If you did not receive any additional form after your :ref:`initial
+application for Amazon MWS keys <amazon/developer-form>`, you may disregard this section.
+
+.. warning::
+   If you are not hosted on Odoo.com (*online*) or on Odoo.sh, you should adapt the answers related
+   to hosting according to your own infrastructure and data protection policy.
+
+- **Describe all functionalities in your application where Personally Identifiable Information (e.g.
+  customer name, street address, billing address) is required.**
+
+  ► If you did not apply for the *Merchant Fulfilled Shipping* function:
+    | - Generation of customer invoices
+
+  ► If you applied for the *Merchant Fulfilled Shipping* function:
+    | - Generation of delivery orders
+    | - Generation of customer invoices
+
+- **List all outside parties with whom your organization shares Amazon Information (e.g. information
+  exposed by Amazon through Amazon MWS, Seller Central, or Amazon's public-facing websites) and
+  describe how your organization shares this information.**
+
+  ► If you do not share Amazon Information with outside parties:
+    Odoo does not share any information with outside parties.
+
+  ► If you share Amazon Information with outside parties:
+    [Description of your organization's policy regarding Amazon Information]
+
+- **List all non-Amazon MWS sources where you retrieve Amazon Information.**
+
+  Odoo only relies on MWS to retrieve Amazon Information.
+
+- **Describe how your organization restricts public access to databases, file servers, and
+  desktop/developer endpoints.**
+
+  | - Access to the postgreSQL database through the network is disabled and standard ports are
+  |   closed. The database is only accessible through a socket on the server itself.
+  | - The reverse proxy only serves whitelisted directories that are only from sources controlled by
+  |   Odoo S.A.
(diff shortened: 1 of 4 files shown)


sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,124 @@
...
+    [Description of your organization's policy regarding Amazon Information]
+
+- **List all non-Amazon MWS sources where you retrieve Amazon Information.**
+
+  Odoo only relies on MWS to retrieve Amazon Information.
+
+- **Describe how your organization restricts public access to databases, file servers, and
+  desktop/developer endpoints.**
+
+  TODO
```
dbo-odoo: To check with saas office:
- http access to postgres is disabled and standard ports are closed, the db is only accessible through a socket on the server itself
- the reverse proxy only serves whitelisted directories that are from sources only controlled by Odoo S.A.
- API endpoints are password protected (PBKDF2+SHA512 encryption, salted + stretched for thousands of rounds)
- Login credentials are always transmitted securely over HTTPS

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,124 @@
...
+
+- **Describe how your organization uniquely identifies employees and restricts access to Amazon
+  Information on a need-to-know basis.**
+
+  TODO or organization dependent ?
```
dbo-odoo: org dependent imho (e.g. do they share a user or not? we can't tell)

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,124 @@
...
+
+- **Provide details on your organization's privacy and data handling policies (a link to your policy
+  is also acceptable).**
+
+  [Description of your organization's privacy and data handling policies]
+
+- **Describe where your organization stores Amazon Information and provide details on how you
+  encrypt this information (e.g., algorithm).**
+
+  TODO storage
```
dbo-odoo: To check with saas office:
- amazon information is stored in unencrypted database. Direct access to the database is not possible for the customer outside of UI interactions or API calls, granular access rights control ensure access is not shared to all users of the database.

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,124 @@
...
+
+  Odoo does not encrypt Amazon Information at rest. However, all data transfers are encrypted.
+
+- **Describe how your organization backups or archives Amazon Information and provide details on how
+  you encrypt this information (e.g., algorithm).**
+
+  TODO backups
```
dbo-odoo: saas office, i honestly don't know our exact plans for that

AntoineVDV:
@odony Could you go over the answers (starting from [here](https://github.com/odoo/documentation-user/pull/500/files#diff-36e6fea0236619f0546af60df84f1cd7R45)) to check that everything is correct ? In the case where the answer would not be the same for two platforms, let's answer the question in the context of the least secure platform.
Thanks :)

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,148 @@
...
+<http://docs.developer.amazonservices.com/en_US/dev_guide/DG_Registering.html>`_ and follow the
+instructions to register as a developer. Take care to choose the form "I represent a seller
+organization integrating with Amazon MWS for its own selling account only.".
+
+Fill out the *Developer Registration and Assessment form* as suggested below and provide your own
+contact information in the **Developer contact information** section. In the **Business use
+information** section, select the correct region of your seller account. For the other sections,
+adapt your responses in accordance with your business case.
+
+Give a particular attention to the **Merchant Fulfilled Shipping** function. It should only be
```
dbo-odoo: You should pay particular attention to the **Merchant Fulfilled Shipping** ~function~ section/field (?)

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,148 @@
...
+   to Odoo <https://www.odoo.com/help>`_.
+
+.. image:: ./media/dev_form.png
+
+Answer the Additional Form
+==========================
+
+This section lists all questions asked by Amazon in additional forms. The answers are tailored for
+*Odoo Online* and *Odoo.sh*. If you did not receive any additional form after your :ref:`initial
+application for Amazon MWS keys <amazon/developer-form>`, disregard the following.
```
dbo-odoo: [...] you may disregard this section.

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,148 @@
...
+
+
+.. warning::
+   If you host Odoo yourself (*on-premise*), adapt the answers related to hosting to your own
```
dbo-odoo: If you are not hosted on Odoo.com or Odoo.sh (*on-premise*), you should adapt your answers for the Hosting section according to your own infrastructure and data protection policy.

=> they may not be hosting *themselves*, as they might have a partner

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,148 @@
...
+    [Description of your organization's policy regarding Amazon Information]
+
+- **List all non-Amazon MWS sources where you retrieve Amazon Information.**
+
+  Odoo only relies on MWS to retrieve Amazon Information.
+
+- **Describe how your organization restricts public access to databases, file servers, and
+  desktop/developer endpoints.**
+
+  | - Access to the postgreSQL database through HTTP is disabled and standard ports are closed. The database is only accessible through a socket on the server itself.
```
dbo-odoo: Access to the postgreSQL database through ~HTTP~ the network

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,148 @@
...
+  encrypt this information (e.g., algorithm).**
+
+  Amazon Information is stored in an unencrypted database. Direct access to the database is not
+  possible for the customer outside of UI interactions or API calls. Granular access rights control
+  ensures that access is not shared to all users of the database.
+
+- **Describe how your organization backups or archives Amazon Information and provide details on how
+  you encrypt this information (e.g., algorithm).**
+
+  Backups are stored in an unencrypted database.
```
dbo-odoo: The entire database is backed up once a day and backups are kept for a minimum of three months according to the [Odoo Online SLA](https://www.odoo.com/cloud-sla). Backups are hosted on several remote servers as unencrypted database dumps; these backups can only be retrieved by Odoo S.A. employees through support requests.

sales/sale_amazon/apply.rst
```diff
@@ -0,0 +1,148 @@
...
+
+- **Describe where your organization monitors and detects malicious activity in your application(s).**
+
+  We have automated probes on our server that report their status in Munin, an opensource monitoring
```
dbo-odoo: Odoo Online uses automated probes...

sales/sale_amazon/apply.rst
AntoineVDV: Amazon calls that a function in the subtitle but it's indeed meaningless.
Let's just remove the word to avoid any confusion.

odony:
Duh, my review was in progress and I had a bunch of comments and suggestions. For some reason I can't post it now, GH fails with "422 Unprocessable Entity". Tried to restore the branch, just in case, but no.

Well, the bulk of it was about the "Answer the Additional Form". I think it is meant to be filled from the standpoint of the company owner, not Odoo SA. In many cases the suggested answer does not seem to take that into account.

Some examples:
- Amazon would have no idea what "Odoo" is in the context of ACME Inc filling in the form, so any references to "Odoo" should either mean "our enterprise software", or "our SaaS/PaaS service provider".
- The answer "Odoo does not share any information with outside parties" makes little sense. It might instead be something like "We share Amazon Information with Odoo SA, our cloud service provider, for the exclusive purpose of hosting our enterprise management system (including the Amazon MWS integration). See also Odoo SA's Privacy Policy for details at https://www.odoo.com/privacy.". And then they need to complete that answer with info about the other parties they share with.
- Instead of copy/pasting partial descriptions of the security of the Odoo Cloud, we should link to the reference description, as that's what a customer would do. And those references should only be a part of the answer to a question, because the question is never "Does your service provider...", but "Do you...".

Other specific remarks about the answers for the hosting details:
- "unencrypted database" is misleading, considering that in most cases it's hosted on an encrypted filesystem. I'd just link to the Odoo Security page instead of paraphrasing it or including info that would be obsolete at some point (less documentation updates needed, less obsolete docs)
- same thing for backups
- Munin is an internal tool and an implementation detail that customers are not aware of - and it could change at any time. It's also not quite related to intrusion detection. Generally we don't disclose the details of our IDS/IPS tools and strategies, so customers would not know exactly. I imagine customers simply say that their rely on our intrusion detection strategies.
- Incident response plan: the OE contract guarantees that Odoo SA will notify the customer in case of data breach, so that would be a basis to cover the database level. The rest is up to them.

AntoineVDV:
Updated form answers in #523

PR #529, opened 2020-01-16
[REV] pos: revert doc for ingenico and six

Lau-Odoo:
Payments providers ingenico and six were deployed in Odoo 13.0, so we revert the commits in 12.0.


AntoineVDV:
@Lau-Odoo rebase needed

PR #534, opened 2020-01-23
[IMP] discuss: added note about config of outbound server

kbr-odoo:


discuss/email_servers.rst
```diff
@@ -35,6 +34,13 @@ Here is a typical configuration for a G Suite server.
 
 Then set your email domain name in the General Settings.
 
+.. note:: If you get a *[AUTHENTICATIONFAILED] Invalid credentials (Failure)* warning when you *Test
```
AntoineVDV: Nitpicking: You could take the opportunity to return to a new line after the `::` for better code readability.

discuss/email_servers.rst
```diff
@@ -35,6 +34,13 @@ Here is a typical configuration for a G Suite server.
...
+   Connection* on a Gmail address, activate the *Less secure app access* option.
+   A direct link can be `accessed here. <https://myaccount.google.com/lesssecureapps?pli=1>`_
```
AntoineVDV: Nitpicking: We generally avoid returning to a new line in the code if it does not results in a new line in the render. This is purely to avoid misleading writers and make them believe that a new line will be rendered.

AntoineVDV: Super nitpicking: Move the punctuation point after the url so that it's not part of the hyperlink.

discuss/email_servers.rst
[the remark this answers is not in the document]
kbr-odoo: But this is additional information, not a tip/trick. It seems like it should be a note for me.

PR #545, opened 2020-02-06
[IMP] building with python3

arj-odoo:
This PR allows one to build the user documentation with python 3 and sphinx 2.3.1 (stable).


_extensions/odoo/pycompat.py
```diff
@@ -1,13 +1,6 @@
 # -*- coding: utf-8 -*-
 import sys
 
-PY2 = sys.version_info[0] == 2
```
arj-odoo: As the doc builds with python3 and python2 now obsolete, I prefer to delete these instructions.

_extensions/odoo/translator.py
```diff
@@ -44,7 +45,16 @@ class BootstrapTranslator(nodes.NodeVisitor, object):
...
+    def __init__(self, arg1, arg2):
+        if int(sphinx.__version__.split('.')[0]) >= 2:
+            # TODO: remove this hack in V14.0
+            # This preserve compatibility between sphinx 1.6 and 2.3
+            # in python3, the arguments must be inverted.
+            builder = arg2
+            document = arg1
+        else:
+            builder = arg1
+            document = arg2
```
xmo-odoo: Just check if the 1st param is the builder and swap if it's not, there's no reason to rely on version checks.

_extensions/odoo/translator.py
```diff
@@ -58,6 +68,10 @@ def __init__(self, builder, document):
         self.context = []
         self.section_level = 0
 
+        if int(sphinx.__version__.split('.')[0]) >= 2:
+            # TODO: do that by default in V14.0
+            # This preserve compatibility between sphinx 1.6 and 2.3
+            self.config = builder.config
```
xmo-odoo: Either set it unconditionally or set IFF it's unset, no need to version-check. Or fix the places which access self.config to use self.builder.config. Or make self.config into a property.

_extensions/odoo/pycompat.py
xmo-odoo: I'd suggest removing the entire file and fixing up the few callsites then.

arj-odoo: As the function is still useful to handle None and bytes, I suggest to move it at the end of translater.py where all the callsites are.

xmo-odoo: I doubt bytes are still involved in there (might have been a concern in Python 3 but even that's doubtful), and `None` should probably be handled ad-hoc specifically where it's a concern.

I say strip it out and log what's happening where it's called, if one of the items ends up potentially being None (not sure which one would even be that) then it can be appended with `or ''` to specifically handle that case.

Feyensv:
Still works in 1.6.7 after last changes :+1:.

_extensions/odoo/pycompat.py
arj-odoo: After checking, the argument is never strictly equal to None. I modified the code to strip the strings.

xmo-odoo: Nice! Is the str() call even useful?

arj-odoo: The build passes without it...

arj-odoo:
@AntoineVDV 

 * This branch can be used to build the doc with at least sphinx-1.6.1 and sphinx-2.3.1 (stable pipy package)
 * compatibility with python2 is broken. 

To avoid breaking the workflow of marketing, we are waiting for you before merging the branch. All the environments will have to be updated. You can ping me to help you update them when you are ready.

_extensions/odoo/__init__.py
```diff
@@ -28,6 +28,8 @@ def setup(app):
     app.connect('html-page-context', update_meta)
 
 def update_meta(app, pagename, templatename, context, doctree):
+    if not context.get('meta'):
+        context['meta'] = {}
```
AntoineVDV: Can `context['meta']` take another falsy value than `{}` ?

requirements.txt
```diff
@@ -1 +1,2 @@
-Sphinx<2
\ No newline at end of file
+Sphinx==2.3.1
```
AntoineVDV:
```suggestion
Sphinx>=2.4.0
Werkzeug>=1.0.0
```

_extensions/odoo/__init__.py
arj-odoo: I did not experienced any. Either the key does not exist or it is a dictionary.

AntoineVDV: Then I may be missing something but why do you need to set its default value if it's directly followed by a setdefault ?

arj-odoo: Because setdefault returns the value of the key. 

When context['meta'] is not defined, it is set to None.
```
context = {'meta': None}
meta = context.setdefault('meta')
```
meta is equal to None and therefore, meta.setdefault('banner', app.config.odoo_cover_default) crashes. You have the same problem with the .get method:
```
context = {'meta': None}
var = context.get('meta', {})
```
var is still equal to None.

PR #549, opened 2020-02-09
13.0 user doc design update kts

stefanorigano:


stefanorigano:
@ktsamouras , I created this draft PR to facilitate the discussion.

Apparently you forgot to port all the updates from [variables.less](https://github.com/odoo/odoo/blob/13.0/doc/_extensions/odoo_ext/static/variables.less).
As result the headings are keeping the old design.

 I don't know if you're still reordering your branch's commits, but I propose to:
- Squash the commits and use a meaningful commit message; something like `[REF] port developer doc's design`, with "Task #[id]" in the description  (you'll find the ID in our backend addressbar)
- Move in a separate commit your fix for the search-form,  explain the issue in the commit's message and description (read the email I sent you last week)

Beside these all ok to for me 👍

mart-e:
Technically, looks good to me :+1: 
cc @AntoineVDV

stefanorigano:
Hello @AntoineVDV , further improvements are actually out-of-scope.

The aim here is just to update the user doc design to the latest version available (the dev one).
Indeed the dev version gathered all the design feedbacks we received in years while the user one was left behind by this point of view.

If you have more feedback we can open a task in our pipeline and discuss about it ;)

The design is approved, do you have technical remarks?
Thanks

AntoineVDV:
> Hello @AntoineVDV , further improvements are actually out-of-scope.
> 
> The aim here is just to update the user doc design to the latest version available (the dev one).
> Indeed the dev version gathered all the design feedbacks we received in years while the user one was left behind by this point of view.
> 
> If you have more feedback we can open a task in our pipeline and discuss about it ;)
> 
> The design is approved, do you have technical remarks?
> Thanks

The issue is that some features that have a poor design in the dev doc are disabled there but not in the user doc. For instance, the user doc shows toctrees with children toctrees inside the same page, while the dev doc does not.
If fixing this proves to be more than a quick edit let's merge these changes as is and, then, create a follow-up task to later fix the issues introduced by the dev doc's design.

stefanorigano:
Thanks @mart-e and @AntoineVDV ... could you take care of merge?
Designers do not push in production 🐣 

> The issue is that some features that have a poor design in the dev doc are disabled there but not in the user doc.

@AntoineVDV , yep , @ktsamouras added a new commit on top to handle links design for index pages,  but I don't think that there are issues related to the introduction of this design (maybe something I'm not aware of? )
There is room for improvements, of course, but it's still an huge improvement IMO.

Thanks!

PR #554, opened 2020-02-11
13.0 account type jcs

jcs-odoo:
ADD doc: Account Type is removed and Chart of Accounts replaces it
The accounting PO hasn't read this doc yet.


_extensions/odoo/__init__.py
```diff
@@ -14,12 +14,12 @@
...
+       app.set_translator('html', translator.BootstrapTranslator)
     else:
-        if getattr(app.config, 'html_translator_class', None):
-            app.warn("Overriding the explicitly set  html_translator_class setting",
-                     location="odoo extension")
-        app.config.html_translator_class = 'odoo.translator.BootstrapTranslator'
+       if getattr(app.config, 'html_translator_class', None):
+           app.warn("Overriding the explicitly set  html_translator_class setting",
+                    location="odoo extension")
+       app.config.html_translator_class = 'odoo.translator.BootstrapTranslator'
```
xmo-odoo: Why did these lines get reindented / unintented?

AntoineVDV:
@jcs-odoo The fix will need to be back-ported to 11.0 and 12.0

_extensions/odoo/__init__.py
jcs-odoo: I moved it to #559 to port the changes starting from 11.0 and added the spaces.

PR #572, opened 2020-02-21
[REM] *: remove all demo.action buttons from the user documentation

jcs-odoo:
*: accounting, inventory, purchase

@mart-e , would you also remove the code related to the actions?
Cheers :)

Jon

### DIFF
--- a/accounting/bank/setup/create_bank_account.rst
+++ b/accounting/bank/setup/create_bank_account.rst
@@ -42,12 +42,6 @@ support with this bank account.
     the company on your user preferences in order to add, edit or delete bank 
     accounts from another company.
 
-.. demo:fields:: base.action_res_partner_bank_account_form
-
-.. demo:action:: base.action_res_partner_bank_account_form
-
-   View *Bank Account* in our Online Demonstration
-
 .. todo:: add inherited field tooltip
 
     **Display on reports :** Display this bank account on the documents that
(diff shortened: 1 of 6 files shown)


mart-e:
How about the `demo:fields`? The same problems occures no?

jcs-odoo:
> How about the `demo:fields`? The same problems occures no?
@mart-e 
I didn't know that...
All they do is list all the fields' descriptions in html. They are useless and I didn't know they were causing the same issue. Should I remove them as well?

mart-e:
The issue is maybe a bit less obvious. The problem is that the demo... works on demo.odoo.com which uses the last version of odoo. So when you will rebuild the 13.0 documentation in 2 years, you will list the fields of demo.odoo.com that runs with 15.0. The fields tends to change less often but still it is an issue IMHO.
@AntoineVDV agree?

jcs-odoo:
That makes sense.
I delete them right now :)

AntoineVDV:
> The issue is maybe a bit less obvious. The problem is that the demo... works on demo.odoo.com which uses the last version of odoo. So when you will rebuild the 13.0 documentation in 2 years, you will list the fields of demo.odoo.com that runs with 15.0. The fields tends to change less often but still it is an issue IMHO.
> @AntoineVDV agree?

:+1:

PR #556, opened 2020-02-12
[IMP] update content of "Terms & conditions" (Sales - V13)

cdu-odoo:

PR #588, opened 2020-03-03
[FIX] update user manual for set default terms

Czq96:


Czq96:
#1 change one line in /sales/send_quotations/terms_and_conditions.rst
FROM ---`Accounting --> Configuration --> Settings`
TO --- `Settings --> Invoicing` 

#2 add a printscreen of the setting menu
terms_and_conditions00

jcs-odoo:
Hi @Czq96 
Thanks for this contribution.

Actually, this content has already been updated with the PR #556 but it is still not merged.
Sorry about that.

Also, for the menuselection path, we have usually APPNAME --> Menu etc
Or just `menu`.

However, we don't write Settings --> APPNAME.

But this raises a good point, I'll check in the accounting docs to write Accounting/Invoicing instead of just Accounting, when it is appropriate.

Thanks and have a great day!

PR #597, opened 2020-03-16
[ADD] accounting: new documentation about synchronization with Ponto

jcs-odoo:


accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+**Odoo Accounting** can synchronize directly with your bank to get all bank statements imported
+automatically into your database. This allows for easier **bank reconciliation**.
+
+When :doc:`adding a bank account on Odoo <../setup/create_bank_account>`, you can see if your bank
+requires a connection through Ponto by searching for your bank institution, and clicking on it.
+
+.. image:: media/ponto_add_bank.png
+   :align: center
+   :alt: Click on a bank institution to see which third party service is required to synchronize
+         your bank with Odoo Accounting.
```
kbr-odoo:
```suggestion
         your bank with Odoo Accounting
```

accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+
+.. image:: media/ponto_logo.png
+   :align: center
+   :alt: Logo of the Ponto brand
+
+**Odoo Accounting** can synchronize directly with your bank to get all bank statements imported
+automatically into your database. This allows for easier **bank reconciliation**.
+
+When :doc:`adding a bank account on Odoo <../setup/create_bank_account>`, you can see if your bank
+requires a connection through Ponto by searching for your bank institution, and clicking on it.
```
kbr-odoo: Break the line but not the paragraph: https://www.odoo.com/documentation/user/13.0/contributing/documentation/rst_cheat_sheet.html#break-the-line-but-not-the-paragraph

accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+   You can find more information about bank synchronization :doc:`on this page
+   <bank_synchronization>`.
+
+Configuration
+=============
+
+Link your bank accounts with Ponto
+----------------------------------
+
+#. Go on `Ponto's website (https://myponto.com) <https://myponto.com>`_.
```
kbr-odoo:
```suggestion
#. Go to `Ponto's website (https://myponto.com) <https://myponto.com>`_.
```

accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+#. Create an account if you don’t have one yet.
+#. Once you are logged in, create an *organization*.
+
+   .. image:: media/ponto_organization.png
+      :alt: Fill out the form to add an organization in Ponto
+
+#. | Go to :menuselection:`Accounts --> Live`, and click on **Add account**.
+   | You might have to add your **Billing Information** first.
```
kbr-odoo:
```suggestion
#. Go to :menuselection:`Accounts --> Live`, and click on **Add account**. You might have to add your **Billing Information** first.
```

I did not find anything in the documentation guide that says we should do this.. so, I am not sure it should be removed, or not, I leave the comment in case you want to check or something.

accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+   :alt: Logo of the Ponto brand
+
+#. To generate these IDs, go to :menuselection:`Ponto Dashboard --> Integration --> Live`, click on
+   **Add Integration**, fill out the form, and select which accounts you want to synchronize.
+#. Copy the *Client ID* and *Secret ID* generated and paste them in Odoo.
+#. | Configure the synchronization options:
+   | - *Action*: define if Odoo must create a new *Journal* for this bank account, link to an
+     existing Journal, or create no link.
+   | - *Journal*: select the appropriate *Journal*.
+   | - *Synchronization Frequency*: define how often Odoo should fetch the bank statements.
```
kbr-odoo: Same as above here. If this is something we should do, please let me know 'cause I didn't know (:
Otherwise, I believe when we have bullet points we don't need "|".

accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+
+#. | **Update** your database.
+   | - *SaaS* and *Odoo.sh* users: you can skip this step, as your database is automatically updated.
+   | - *Community* and *Enterprise* users: download the latest Odoo source code for your version
+     (you can download it from `this link <https://odoo.com/download>`_ or from GitHub), install it,
+     and restart your server.
+#. | Do a **hard refresh** of your Odoo page by pressing *CTRL + F5*.
+   | A hard refresh clears the cache and the javascript code for the current page.
+#. To **generate your access key**, follow :ref:`these steps above <ponto-link-odoo>`.
+#. To **update your credentials**, follow :ref:`these steps above <ponto-update-credentials>`.
```
kbr-odoo:
```suggestion
#. To **update your credentials**, follow :ref:`the steps above <ponto-update-credentials>`.
```

accounting/bank/feeds/ponto.rst
```diff
@@ -0,0 +1,115 @@
...
+token instead of the current synchronization system with a *Client ID* and a *Secret ID*.
+
+#. | **Update** your database.
+   | - *SaaS* and *Odoo.sh* users: you can skip this step, as your database is automatically updated.
+   | - *Community* and *Enterprise* users: download the latest Odoo source code for your version
+     (you can download it from `this link <https://odoo.com/download>`_ or from GitHub), install it,
+     and restart your server.
+#. | Do a **hard refresh** of your Odoo page by pressing *CTRL + F5*.
+   | A hard refresh clears the cache and the javascript code for the current page.
+#. To **generate your access key**, follow :ref:`these steps above <ponto-link-odoo>`.
```
kbr-odoo:
```suggestion
#. To **generate your access key**, follow :ref:`the steps above <ponto-link-odoo>`.
```

accounting/bank/feeds/ponto.rst
jcs-odoo: applied

accounting/bank/feeds/ponto.rst
jcs-odoo: right :)

accounting/bank/feeds/ponto.rst
jcs-odoo: I did the whole flows myself. That's why I added this part about the Billing Information. Thanks for having checked though :)

jcs-odoo: do you mean the line break?

That is one of my uses of the line breaks:
If I want to separate some information inside a list, then I do a line break.
Here, the info is "add account". the additional info that I want to keep apart is about the billing information.

accounting/bank/feeds/ponto.rst
jcs-odoo: Here I use the line break because of the limitations of the RST code:
I cannot use nested lists when the main list is a numbered list.
Therefore, I used line breaks to display the following points separately.
The - are not rendered as lists but as text.

jcs-odoo: @AntoineVDV any thoughts about this?
I couldn't do a nested list with the main numbered list. Is my use of line breaks with "-" as text a bad idea?

accounting/bank/feeds/ponto.rst
jcs-odoo: Same as above. I used the line break to keep the main info apart from the explanation because we are in a list.

accounting/bank/feeds/ponto.rst
jcs-odoo: indeed

accounting/bank/feeds/ponto.rst
kbr-odoo: hum..thank you for the explan!

AntoineVDV: RST doesn't like when you mix different formatting.
Insert a blank line before `- *Action* ...` and it will render fine without `|`.

jcs-odoo: OH AND THE NUMBERING KEEPS WORKING if I were to add another #. 
Thank you!
Why am I so excited about this.

PR #606, opened 2020-03-23
[ADD] accounting: add new doc about tax/vat returns

jcs-odoo:


kbr-odoo:
Nothing to suggest. Reay to me moves forward.

PR #615, opened 2020-04-02
[IMP] technical: add a static redirection feature for missing pages

AntoineVDV:


redirects.txt
```diff
@@ -0,0 +1,39 @@
...
+#
+#   1. An RST file is moved from one location to another.
+#      Example: The TOC of the Sales app is moved into a file index.rst located in sales/ .
+#      Rule: sales.rst sales/index.rst
+#
+#   2. An RST file is renamed.
+#      Example: The file create_quotations.rst in sales/ is renamed to quotations.rst.
+#      Rule: sales/create_quotations.rst sales/quotations.rst # no longer limited to creating quotes
+#
+#   3. An RST file does not exist is an earlier supported version. It can be due to cases 1 and 2.
```
jcs-odoo, line 21:
```suggestion
#   3. An RST file does not exist in an earlier supported version. It can be due to cases 1 and 2.
```

redirects.txt
```diff
@@ -0,0 +1,39 @@
...
+#      Example: The file batch_sdd.rst is added in accounting/ as of 13.0.
+#      Rule (11.0 to 12.0 only): accounting/batch_sdd.rst accounting.rst # new page added in 13.0
+#
+#   4. An RST file does not exist is an ulterior supported version. It can be due to cases 1 and 2.
```
jcs-odoo, line 25:
```suggestion
#   4. An RST file does not exist in an ulterior supported version. It can be due to cases 1 and 2.
```

_extensions/redirects.py
```diff
@@ -0,0 +1,54 @@
...
+
+            redirect = match_result.group(1).rstrip()
+            if redirect.count(' ') != 1:
+                logger.error("Ignoring malformed redirection: %s" % redirect)
+                continue
+            from_path, to_path = redirect.split()
+            logger.debug("Redirecting '%s' to '%s'" % (from_path, to_path))
+
+            from_path = from_path.replace(source_suffix, '.html')
+            to_path_prefix = '..%s' % os.path.sep * (len(from_path.split(os.path.sep)) - 1)
```
mart-e, line 40: this assume we support both `\` and `/` as separator depending of the plateform? I think we should force to use `/` in the `from_path`.
Have you tried using [pathlib](https://docs.python.org/3/library/pathlib.html)? I think it can clarify the code (and can handles all the `os` operations).

AntoineVDV: I considered pathlib but it was added in Python 3.4. Do we want to have this restriction all of sudden when we just updated our requirements to just "Python 3" ?

Ok for enforcing `/` in the from_path.

AntoineVDV: In fact I removed this line completely and replaced it with `absolute_to_path = os.path.join(app.builder.outdir, to_path)` so that both paths start from the root, which makes much more sense.

mart-e: > I considered pathlib but it was added in Python 3.4. Do we want to have this restriction all of sudden when we just updated our requirements to just "Python 3" ?

We [never supported python below 3.5](https://www.odoo.com/groups/community-59/community-32859400) afaik

AntoineVDV:
@mart-e Please also check that all use case are covered :)

_extensions/redirects.py
```diff
@@ -0,0 +1,55 @@
...
+        return
+
+    source_suffix = next(iter(app.config.source_suffix))
+
+    if not type(app.builder) == builders.StandaloneHTMLBuilder:
+        logger.info("Redirects are only supported by the 'html' builder. Skipping...")
+        return
+
+    with open(path) as redirects:
+        pattern = re.compile(r'^([\w\-./ ]+)(#.*)?$')
```
mart-e, line 27:
```suggestion
        pattern = re.compile(r'^[ ]*([\w\-./]+)[ ]*([\w\-./]+)[ ]*(#.*)?$')
```

_extensions/redirects.py
```diff
@@ -0,0 +1,55 @@
...
+        for line in redirects.readlines():
+            # Exclude comment or empty lines
+            match_result = pattern.search(line)
+            if not match_result:
+                continue
+
+            # Parse the rule
+            redirect = match_result.group(1).rstrip()
+            if redirect.count(' ') != 1:
```
mart-e, line 36: this can be avoided with above regex

AntoineVDV: I extracted the `[ ]*` as suggested above but I think that it's better to warn the user if he wrote something that looks like a rule but it malformed.

For instance, I wouldn't be surprise that someone writes `sales/my_new_page.rst  # This is the new file` instead of `sales/old_page.rst sales/my_new_page.rst`.

What do you think ?

mart-e: Indeed, if we don't match the regex, we should log an error "please use format `source.rst target.rst`. Just that we don't need to make the job that can done with a regex.
By the way, we can direct include the `.rst` extension in the regex too :+1:

AntoineVDV: Fine for me :+1:

PR #627, opened 2020-04-07
[FIX] sales/send_quotations: add redirection rule (CDU)

cdu-odoo:


redirects.txt
```diff
@@ -42,4 +42,6 @@ sales/sale_ebay.rst sales/ebay_connector.rst                    # sale_ebay -> e
 sales/ebay/setup.rst sales/ebay_connector/setup.rst             # ebay/* moved to ebay_connector/* (#524)
 sales/ebay/manage.rst sales/ebay_connector/manage.rst           # ebay/* moved to ebay_connector/* (#524)
 
-accounting/localizations/nederlands.rst accounting/localizations/netherlands.rst  # (#514)
\ No newline at end of file
+accounting/localizations/nederlands.rst accounting/localizations/netherlands.rst  # (#514)
+
+sales/send_quotations/optional_items.rst sales/send_quotations/optional_products.rst (#533)
```
jcs-odoo, line 47: @AntoineVDV 

isn't it an issue that there is no # before the comment, line 47?

AntoineVDV: Yes it is, good call

PR #633, opened 2020-04-08
[REM] accounting: delete old doc about expenses

jcs-odoo:
This documentation is five years old and the related content is in the Expenses section rather than Accounting.


accounting/payables/misc.rst
```diff
@@ -5,5 +5,4 @@ Miscellaneous
 .. toctree::
    :titlesonly:
 
-   misc/forecast
-   misc/employee_expense
```
AntoineVDV, line 9: According to your commit message, a redirect rule is needed here.

jcs-odoo: I wondered if to do it or not, as these were two separate files. 
I saw it as a removal, not as a merge, but you're right :) Thanks

PR #643, opened 2020-04-16
Country pricing Compliance

gorecrds:
Country pricing Compliance. Customer using VPN and other mechanism


odony:
Nope, I don't see the value of those changes, it's more noise in the contract.

odony:
@AntoineVDV I've added a CODEOWNERS file, all changes to legal material (basically `/legal*`) have to be approved because they are published on our website and have legally-binding consequences that can impact our business.

PR #665, opened 2020-05-05
[IMP] google_spreadsheet: warn about timeouts

dbo-odoo:


dbo-odoo:
@mao-odoo can you confirm what I say is correct regarding opw 2244764 and 2152346?

i've checked the rendering, my rst syntax is correct, i just need confirmation that i'm saying the truth :p

PR #675, opened 2020-05-17
Update partnership.rst

Karwancheecho:


odony:
Hello, you are proposing to update our master agreement template to put you company address, I don't think this is what you want.
The usual process for signing the agreement is to use an electronic signature process like Docusign, or use a PDF editor to add your signature in it, or the old school way: print the agreement, sign it, and scan/take a picture of it.
Have a nice day!

PR #683, opened 2020-05-28
[IMP] general: new version of the Authorize.Net doc

jcs-odoo:


general/payment_acquirers.rst
```diff
@@ -7,5 +7,5 @@ Payment Acquirers
 
    payment_acquirers/payment_acquirers
    payment_acquirers/wire_transfer
-   payment_acquirers/paypal
    payment_acquirers/authorize
+   payment_acquirers/paypal
```
dbo-odoo, line 11: any reason for this switch?

general/payment_acquirers/authorize.rst
```diff
@@ -1,81 +1,116 @@
...
-    :align: center
+If not done yet, choose a plan and `Sign Up for an Authorize.Net account
+<https://www.authorize.net/sign-up.html>`__.
 
-* Go through the registration steps. 
-* The account is set as a test account by default. You can use this test
-  account to process a test transaction from Odoo.
-* Once ready, switch to **Production** mode. 
+Odoo needs your **API Credentials & Keys** to connect with your Authorize.Net account, which
+comprises:
```
dbo-odoo, line 22:
```suggestion
comprise:
```

general/payment_acquirers.rst
jcs-odoo: alpahbetical, after the bank wire.
The logic would be 
- bank payment acquirers
- online payment providers (alphabetical)

general/payment_acquirers/authorize.rst
```diff
@@ -1,81 +1,116 @@
...
+provider, allowing businesses to accept **credit cards**.
 
+.. image:: media/authorize-net.png
+   :align: center
+   :alt: Authorize.Net logo
 
-Create an Authorize.Net account
-===============================
+This :doc:`Payment Acquirer <payment_acquirers>` offers additional options that are not available
+for other Payment Acquirers, such as the ability to process your customer's payment after delivery.
```
Lau-Odoo, line 13: Just a detail but I would put the link on the second "Payment Acquirers" instead of the first one. The first one is Authorize.Net only, while the second is about all the payment acquirers. So, it would be more logical (to me) if the link was on "Payment Acquirers" (instead of payment acquirer).

general/payment_acquirers/authorize.rst
```diff
@@ -1,81 +1,116 @@
...
-and activate the **Enabled** state.
+To configure Authorize.Net as Payment Acquirer in Odoo, go to :menuselection:`Accounting -->
+Configuration --> Payment Acquirers`, open **Authorize.Net**, and change the **State** to *Enabled*.
+Don't forget to click on *Save* once you've set everything up.
 
-.. image:: media/paypal_live.png
-    :align: center
+.. note::
+   Please refer to the :doc:`Payment Acquirers documentation <payment_acquirers>` to read how to
+   configure this payment acquirer.
```
Lau-Odoo, line 50: This link is already available above. Not sure it is useful to have it several times in the same doc.

general/payment_acquirers/authorize.rst
jcs-odoo: good idea :)

general/payment_acquirers/authorize.rst
jcs-odoo: It's a different section and it specifically directs readers to configure according to that doc, so I'd keep it here, at least. But I think it's okay to have it twice in the same doc. I don't expect everyone to read everything from top to bottom.

Cheers :)

PR #686, opened 2020-05-28
[IMP] odoo_sh: clarify dev branch behavior

jop-odoo:
Some users have reported that current wording sets the expectation that there will be a notification when dev branches will be eliminated after the three days. This makes it clear that after 3 days dev branches are automatically garbage collected at any moment without prior notice.


AntoineVDV:
Hey @jop-odoo, could you merge your commits and have a final commit message like
> [IMP] odoo_sh: clarify dev branch behavior
>
> Some users have reported that current wording sets the
> expectation that there will be a notification when dev
> branches will be eliminated after the three days.
> This makes it clear that after 3 days dev branches are
> automatically garbage collected at any moment without
> prior notice.

Thanks !

PR #688, opened 2020-05-29
[IMP] odoo_sh: clarify dev branch behavior

jop-odoo:
Some users have reported that current wording sets the
expectation that there will be a notification when dev
branches will be eliminated after the three days.
This makes it clear that after three days dev branches are
automatically garbage collected at any moment without
prior notice.


jop-odoo:
@AntoineVDV Created a new PR with only one commit.

PR #697, opened 2020-06-11
[FIX] expense: Reference in first position in email subject

simongoffin:
As commited here: 

https://github.com/odoo/odoo/commit/4ba8f51d87be394988e6e2889cf03f351afea1a0

The reference must be in first position in the email subject.

https://github.com/odoo/odoo/blob/13.0/addons/hr_expense/models/hr_expense.py#L568

opw:2255175

### DIFF
--- a/expense/expense.rst
+++ b/expense/expense.rst
@@ -81,8 +81,8 @@ For security purposes, only authenticated employee emails
 
 .. tip::
     The expense product is set automatically if the mail subject contains 
-    the product's internal reference between brackets (e.g. [Food]). 
-    Type the expense amount in the mail subject to set it on the expense too.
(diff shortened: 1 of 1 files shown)


mart-e:
The docstring says
```
             Some valid examples:
                 Travel by Air [TICKET] USD 1205.91
                 TICKET $1205.91 Travel by Air
                 Extra expenses 29.10EUR [EXTRA]
```
but the `_parse_product` code indeed just check the first word.

@tivisse was it intended to support the bracket format too?

tivisse:
@mart-e it was, at least at the beginning. 

Then it was cleaned during a back2basics tasks. The doc wasn't updated though.

PR #708, opened 2020-06-24
[ADD] support for meta directive

xmo-odoo:
```rst
.. meta::
    :description: blah blah blah
    :keywords lang=en: x y z
```
will generate
```html
<meta name="description" content="blah blah blah">
<meta name="keywords" lang="en" content="x y z">
```
Also cleaned up `layout.html` a tad while at it:
* we're now properly setting the default metas via the translator so setting them in the template is unnecessary
* toggle the `html5_doctype` Sphinx flag as it... uses an HTML5 doctype so we don't need to do that by hand anymore (also uses a `charset` meta instead of an `http-equiv="Content-Type"` and suppresses sphinx's own `X-UA-Compatible`)

### DIFF
--- a/_extensions/odoo/layout.html
+++ b/_extensions/odoo/layout.html
@@ -1,4 +1,5 @@
 {% extends "basic/layout.html" %}
+{% set html5_doctype = True %}
 
 {% set classes = [] %}
 {% if pagename == master_doc %}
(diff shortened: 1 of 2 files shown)


AntoineVDV:
I'll let Martin review since I never played much with the odoo extension.

mart-e:
@xmo-odoo already done, double forward !
https://github.com/odoo/documentation-user/commits/13.0

xmo-odoo:
@mart-e @AntoineVDV also says it needs to be forward-ported up to 13.3, and 13.2 closed, or something, IDK.

mart-e:
oh right, we do have other versions :neutral_face:

PR #714, opened 2020-06-29
[FIX] iot: Connection process changed

aprieels:
We had to modify the connection method due to changes that will be
introduced soon in modern web browsers. (See odoo/odoo#51088)

We also replace the numbering of the connection steps with bullets
because they were incorrectly numbered because of the images.


aprieels:
@Lau-Odoo we had to change the way the IoT Box is connected to a database. I've updated the doc, can you have a look?

iot/config/connect.rst
```diff
@@ -20,35 +20,42 @@ Follow the steps to connect your IoT Box.
...
-1. Connect to the IoT Box all the devices that have to be connected with
-   cables (ethernet, usb devices, etc.).
+* Connect to the IoT Box all the devices that have to be connected with
+  cables (ethernet, usb devices, etc.).
 
-2. Power on the IoT Box.
+* Power on the IoT Box.
 
-3. Then click on the Scan button.
+* Read the Pairing Code from a screen or a receipt printer connected to the IoT Box.
```
AntoineVDV, line 28: You can keep the numberings (in WiFi Connection too).
Just indent the `image` directive with 3 spaces so that it aligns with this line and it should work just fine. Also, you can remove the `:align: center` since the image is now part of a list item.

iot/config/connect.rst
```diff
@@ -20,35 +20,42 @@ Follow the steps to connect your IoT Box.
...
 
 .. image:: media/connect04.png
    :align: center
 
+* Input the Pairing Code and click on the Pair button.
+
+.. note::
+   Recent changes in modern web browsers forced us to modify the connection wizard.
+   If your screen is different from the screenshots, make sure that the
+   `iot_pairing` module is installed.
```
AntoineVDV, line 38:
```suggestion
   ``iot_pairing`` module is installed.
```
This will throw an *UNKNOWN NODE* error otherwise.

iot/config/connect.rst
```diff
@@ -20,35 +20,42 @@ Follow the steps to connect your IoT Box.
...
+* Connect to the IoT Box all the devices that have to be connected with
+  cables (ethernet, usb devices, etc.).
 
-2. Power on the IoT Box.
+* Power on the IoT Box.
 
-3. Then click on the Scan button.
+* Read the Pairing Code from a screen or a receipt printer connected to the IoT Box.
 
 .. image:: media/connect04.png
```
AntoineVDV, line 30: I would decrease the image size so that it doesn't take the entire page width for 2 words :p

iot/config/connect.rst
Lau-Odoo: I would, too, keep the numbering :)

aprieels: Wouldn't it be strange to remove the `:align: center` as other images on the page will be centered?

Lau-Odoo: I think it will be automatically centered because it is part of a list item, am I right? @AntoineVDV

AntoineVDV: No, it won't be automatically centered. But the general guideline for lists' images is to align them to the left because they are usually quite small and the page looks weird if they are centered.

PR #732, opened 2020-07-31
[IMP] Point of Sale: Six >< Worldline

aprieels:
Multiple customers tried to connect Worldline payment terminals, even
though they're not supported, because the documentation was not clear.


aprieels:
@Lau-Odoo @qle-odoo Two consultants reached out to me in the last two weeks because they couldn't connect payment terminals to their customer's DB. Both were trying to connect Worldline terminals. I think that the documentation wasn't clear enough about the compatibility.

Lau-Odoo:
I didn't know about the incompatibility with Worldline's terminals.
My bad! :sob:

PR #737, opened 2020-08-12
[IMP] link to buy compatible checks

fhe-odoo:


accounting/payables/pay/check.rst
```diff
@@ -48,6 +48,7 @@ For the United States, Odoo supports by default the check formats of:
 - **Peachtree**: check in the middle, stubs on top and bottom
 - **ADP**: check in the bottom, and stubs on the top.
 
+You can purchase checks that are compatible with the standard Odoo format (v12.0 and up) at `<https://www.checkdepot.net/checks/laser/Odoo.htm>`_
```
jcs-odoo, line 51:
```suggestion
You can purchase checks that are compatible with the standard Odoo format on `checkdepot.net
<https://checkdepot.net/collections/computer-checks/Odoo>`_.
```

Although, why is checkdepot.net featured?

jcs-odoo (commented):
Hello @fhe-odoo 

I'm cleaning a bit some Pull Requests and I see yours has been left for quite a long while now.

I'm not sure this addition is appropriate for the documentation since it promotes a specific shop. Is there a reason for this?

Thanks and best regards,
Jonathan

fhe-odoo:
@jcs-odoo Because it's the only online shop that we have tested provides the paper checks that we have designed this feature on. We don't have a choice but to promote some shop since we are not going to print and sell paper checks ourselves. So our customers have to buy them from somewhere.

jcs-odoo:
> @jcs-odoo Because it's the only online shop that we have tested provides the paper checks that we have designed this feature on. We don't have a choice but to promote some shop since we are not going to print and sell paper checks ourselves. So our customers have to buy them from somewhere.

Hello Fabrice!
While I agree we could add some general information to help customers acquire compatible checks, I'd like to keep the promotion out of the functional documentation.

As far as I know, Odoo requires the user to set up the proper check layout (top/middle/bottom + margins) before printing the first check. Since we are compatible with multi-purpose laser checks from various suppliers, I wouldn't want to mislead users into thinking they have to limit their purchasing choices to "Odoo compatible" laser checks.

Let me know if I'm misunderstanding, and I'll happily reopen this PR if necessary.

Best regards,

fhe-odoo:
> Since we are compatible with multi-purpose laser checks from various suppliers

Well I think this is the misunderstanding. You can buy US checks on many websites but there is no standard format, and certainly not down to the millimeter, which is precisely the issue because if for example the amount-in-words line is 2mm above what Odoo expects, you'll end up with a strikethrough the amount and won't be readable by US ATM's OCR for example. So it needs to be very precise and we made all our developments and tests with the checks from Checkdepot. We chose this site because it seemed to be the cheapest and longest-lasting site for this purpose.
I know it looks like promotion but since Odoo doesn't manufacture laser checks, we kinda have to provide a go-to solution to our users. We have created favoritism/promo with several other vendors/platforms (axivox, taxcloud, unsplash, etc...)

PR #747, opened 2020-09-17
[FIX] Remove api.multi which doesn't exist in 13.0

raf-odoo:
`api.multi` was removed in 13.0 and is the default behavior. Attempting
to use it in this tutorial will result in a build error.


raf-odoo:
@rim-odoo This was the main issue because it broke the build and got introduced in 13.0. The other issues are in a separate 12.0 PR.

rim-odoo (approved):

PR #745, opened 2020-09-09
[ADD] Localizations: Functional documentation for l10n_cl_edi

ren-odoo:
Functional guidance for the Chilean localization:
- Electronic Invoice configuration 
- Use cases and testing
- Financial Reports

PR #757, opened 2020-10-05
[FIX] accounting: fix bad indenting in chile localization

jcs-odoo:
The original PR #745 had errors that led to error messages.

Please forward port in 14.0 as well

### DIFF
--- a/accounting/fiscal_localizations/localizations/chile.rst
+++ b/accounting/fiscal_localizations/localizations/chile.rst
@@ -397,7 +397,7 @@ from a sales order, proceed to validate the invoice.  After the invoice is poste
 
   .. image:: media/Chile16.png
      :align: center
-	 :alt: DTE XML File displayed in chatter.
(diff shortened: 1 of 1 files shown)


AntoineVDV:
It looks like you rebased your branch on 14.0 rather than 13.0!

jcs-odoo:
> It looks like you rebased your branch on 14.0 rather than 13.0!

My bad, it should be fixed now

AntoineVDV (changes requested):
Please update your commit message to:
- Get rid of the version (a commit can live in several versions, mentioning them in the title makes no sense).
- Allow readers to understand quickly what it's about and which files it impacts.

For example ->
> [FIX] accounting: fix bad indenting in chile localization

jcs-odoo:
There it is done.
So where should I mention the forward port to14.0 as well?
Here, as a comment?
Thanks

AntoineVDV (approved):

PR #763, opened 2020-10-08
[IMP] odoo_sh: branch settings

sts-odoo:
- test tags have been added
- modules settings have been moved from global settings to branch
settings

### DIFF
--- a/odoo_sh/getting_started/branches.rst
+++ b/odoo_sh/getting_started/branches.rst
@@ -94,7 +94,8 @@ When you push a new commit in one of these branches,
 a new server is started, with a database created from scratch and the new revision of the branch.
 The demo data is loaded, and the unit tests are performed by default.
 This verifies your changes do not break any of the features tested by them. If you wish, you can
-disable the tests in the :ref:`branch's settings <odoosh-gettingstarted-branches-tabs-mails>`.
+disable the tests or allow specific tests to be run with custom tags in the :ref:`branch's settings
+<odoosh-gettingstarted-branches-tabs-settings>`.
 
 Similar to staging branches, the emails are not sent but are intercepted by a mailcatcher and
 scheduled actions are not triggered as often is the database is not in use.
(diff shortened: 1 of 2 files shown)


odoo_sh/getting_started/branches.rst
```diff
@@ -94,7 +94,7 @@ When you push a new commit in one of these branches,
 a new server is started, with a database created from scratch and the new revision of the branch.
 The demo data is loaded, and the unit tests are performed by default.
 This verifies your changes do not break any of the features tested by them. If you wish, you can
-disable the tests in the :ref:`branch's settings <odoosh-gettingstarted-branches-tabs-mails>`.
+disable the tests allow specific test to b erun with custom tags in the :ref:`branch's settings <odoosh-gettingstarted-branches-tabs-settings>`.
```
AntoineVDV, line 97:
```suggestion
disable the tests allow specific test to b erun with custom tags in the :ref:`branch's settings
<odoosh-gettingstarted-branches-tabs-settings>`.
```
You can break the line inside a `ref` directive anywhere before the `<`. Preferably after the label.

AntoineVDV (commented):
LGTM

I let DVE make the approving review.

odoo_sh/getting_started/branches.rst
Icallhimtest, line 97: "OR allow specific testS to BE RUN"

Did I put the mails section mails :P ?

odoo_sh/getting_started/branches.rst
```diff
@@ -275,9 +275,30 @@ to manually set it up again on every commit. If you choose new build for a stagi
...
+
+If the tests are enabled, the standard Odoo modules suite can take up to 1 hour.
+This setting applies to development builds only.
+Staging builds duplicate the production build and the production build only installs base.
+
+
 **Test suite**
 
 For development branches, you can choose to enable or disable the test suite. It's enabled by default.
+When the test suite is enabled, you can restrict the tests to bu run for specific tests by adding test tags.
```
Icallhimtest, line 300: "to BE run" :D

But I'd simplify in "you can restrict them by specifying test tags"

For the line below: I'm not sure how it gives in the interface but maybe include the link in the "test tags" from the phrase above?

Icallhimtest:
@AntoineVDV The odoo_sh "module" was added in v11.0, as it's independant of any Odoo version can we target our PRs to v11.0? Or what would you recommend we do?

AntoineVDV:
> @AntoineVDV The odoo_sh "module" was added in v11.0, as it's independant of any Odoo version can we target our PRs to v11.0? Or what would you recommend we do?

Good question. With the doc apocalypse slowly approaching we were planning on freezing the versions <= 11.0 of the doc and stop building them but have them available as a static version. This however doesn't work if you need to update the doc of Odoo.sh for unsupported versions of Odoo. Is it the case, or do we need to find a solution for that?
Imho I don't think that we should keep updating the 11.0 of the doc, but rather keep it as is and make it clear on the website that it is outdated.

Icallhimtest:
> This however doesn't work if you need to update the doc of Odoo.sh for unsupported versions of Odoo. Is it the case, or do we need to find a solution for that?
> Imho I don't think that we should keep updating the 11.0 of the doc, but rather keep it as is and make it clear on the website that it is outdated.

If we can add a warning with a link to the most recent version it's fine by me to abandon the outdated branches.

AntoineVDV:
> If we can add a warning with a link to the most recent version it's fine by me to abandon the outdated branches.

This is planned as part of the soon to come doc apocalypse :rocket:

Icallhimtest (approved):

PR #766, opened 2020-10-10
[FIX] Correction of note in budget.rst

jeroen-vet:
Correction of obvious mistake.


mart-e:
Thanks for the PR. It was targetting the wrong branch but I figured it out ;-)

PR #770, opened 2020-10-15
[REM] project: remove overview/setup.rst + overview/share.rst

kbr-odoo:
- as decided by FP, certain docs should be removed


jcs-odoo:
@kbr-odoo 
I agree for the "share" doc. But for setup, it could maybe stay, but with a different focus? For example, explaining the features in the settings (and linking to the appropriate doc)? And maybe the way the mail template works with the stages options? What do you think?

kbr-odoo:
> @kbr-odoo
> I agree about the "share" doc. But for setup, it could maybe stay, but with a different focus? For example, explaining the features in the settings (and linking to the appropriate doc)? And maybe the way the mail template works with the stages options? What do you think?

@jcs-odoo No. It would still be too simple for documentation, plus I am organized towards eLearning. Thank you for your input.

jcs-odoo (approved):
sorry, I forgot to approve with the previous comment :)

PR #772, opened 2020-10-19
[REM] project: remove get_started.rst + sub_tasks.rst

kbr-odoo:
- as decided by fp, some docs should be removed as the content can be easily figured by the user alone.
@jcs-odoo Thanks.


jcs-odoo (approved):
👍

AntoineVDV (changes requested):
Please indicate the actual reason for this change in the commit message. "fp told us" is not one that will still make sense/be understandable in 6 months.

AntoineVDV:
@kbr-odoo you requested a new review but without any change ?

kbr-odoo:
> Please indicate the actual reason for this change in the commit message. "fp told us" is not one that will still make sense/be understandable in 6 months.

Hello, Antoine!
Could you please approve it? Otherwise, let me know if you would like me to change it still.
thanks! :)

jcs-odoo (approved):

AntoineVDV (approved):

AntoineVDV:
Hey @kbr-odoo, I force-pushed the commit with an updated message to reflect the real motivation behind the removal of these pages. Please understand that those blocking reviews are not made with the intention of annoying you, but to make sure we don't do the opposite work in a few months due to a lack of motives for the original task :wink:

kbr-odoo:
> Hey @kbr-odoo, I force-pushed the commit with an updated message to reflect the real motivation behind the removal of these pages. Please understand that those blocking reviews are not made with the intention of annoying you, but to make sure we don't do the opposite work in a few months due to a lack of motives for the original task 😉

I understand. You don't even need to explain yourself 👍
I see your message, and now I know it is clearer what we expect from us. I will share it with the others!
Thank you very much!

PR #773, opened 2020-10-20
[IMP] accounting: Warning about currency exchange entries

raf-odoo:
It's not really obvious that payments and invoices need to be fully
reconciled before their currency exchange entries get created, so this
is a warning about that.

### DIFF
--- a/accounting/others/multicurrencies/exchange.rst
+++ b/accounting/others/multicurrencies/exchange.rst
@@ -105,6 +105,8 @@ journal entries. All the exchange rates differences are recorded in it.
 .. tip::
     The Exchange difference journal can be changed in your accounting settings.
 
+.. include:: full_reconcile_warning.rst
+
 .. seealso::
     * :doc:`../../bank/reconciliation/configure`
     * :doc:`../../bank/reconciliation/use_cases`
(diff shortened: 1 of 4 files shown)


raf-odoo:
@rim-odoo

raf-odoo:
I also verified that this is still the case on 14.0.

AntoineVDV:
I understand the needs for those notes but this is exactly the same text repeated in all three notes. Please look into "sphinx substitutions" to see if you can make them work across several pages (never tried myself). This is the equivalent of a `t-call` in QWeb.

raf-odoo:
@AntoineVDV Yeah, I agree the duplication isn't great. It's a result of those 3 articles duplicating a lot from each other. I'll look into that substitution mechanism you mentioned.

raf-odoo:
@AntoineVDV RST substitution didn't work since that's for single lines only. So I used the include mechanism instead. I think that to eliminate the structural duplication between those 3 documents, they need to be reworked. I can create an issue for it if you like.

AntoineVDV:
> RST substitution didn't work since that's for single lines only. So I used the include mechanism instead.

Works too. It might be even better suited in this case.

> I think that to eliminate the structural duplication between those 3 documents, they need to be reworked. I can create an issue for it if you like.

Either they have already been reworked or they will be (as part of the entire rewrite of the doc started in 13.0). In both cases, don't bother.

rim-odoo (approved):

PR #783, opened 2020-11-06
VoIP doc: move from CRM to General + new doc for Axivox as VoIP provider

jcs-odoo:


general.rst
```diff
@@ -11,3 +11,4 @@ General
    general/unsplash
    general/base_import
    general/auth
+   general/voip
```
jcs-odoo, line 14: For the forward port in ulterior version, this line can be put after **Payment Acquirers** and before **In-App Purchase**

AntoineVDV (approved):
> [MOV] general: moving VoIP docs from "CRM" to "general" 

Please only do that in >= 13.0
The structure in 12.0 is too much different from that of 13.0. That'd lead to overly complicated gitmnastic and content loss.

AntoineVDV (changes requested):
Clicked the wrong radio ><

jcs-odoo:
Duplicate of #784 

which is based on 13.0

jcs-odoo:
@AntoineVDV I guess we can delete the branch as well, then.

PR #808, opened 2020-12-14
[ADD] practical: new doc about database upgrade

jcs-odoo:


jcs-odoo:
Changed to v14 in #809

AntoineVDV:
> Changed to v14 in #809

@jcs-odoo Why?

jcs-odoo:
> > Changed to v14 in #809
> 
> @jcs-odoo Why?

There is some content that the author wrote that links to another doc published in v14 only.

AntoineVDV:
> > > Changed to v14 in #809
> > 
> > 
> > @jcs-odoo Why?
> 
> There is some content that the author wrote that links to another doc ( /db_management/hosting_changes) published in v14 only.
> I guess I could put this PR back to 13.0 if the related doc was also in 13.0.

Alternatively, a new PR could target the 12.0 so that the content is available in all versions, but without the links.
Then the currently open PR (in 14.0) merges the same content, but with the links.

In fact, the balance is between the availability of the content and the effort of adding the links (how many btw) back. Which one is best?

jcs-odoo:
> > > > Changed to v14 in #809
> > > 
> > > 
> > > @jcs-odoo Why?
> > 
> > 
> > There is some content that the author wrote that links to another doc ( /db_management/hosting_changes) published in v14 only.
> > I guess I could put this PR back to 13.0 if the related doc was also in 13.0.
> 
> Alternatively, a new PR could target the 12.0 so that the content is available in all versions, but without the links.
> Then the currently open PR (in 14.0) merges the same content, but with the links.
> 
> In fact, the balance is between the availability of the content and the effort of adding the links (how many btw) back. Which one would be best?

You're right @AntoineVDV . 
It would work in 12.0, and it is still only `/db_management/hosting_changes` that is missing, and I see no reason not to port it to 12.0 and 13.0.

It's just that I don't know if I can do that? I guess you can :) 
I was merely working with what I had.

AntoineVDV:
You just need to open a PR based on 12.0 with a cherry-pick of the commit in 14.0, then adapt the references to non-existing pages.

jcs-odoo:
> You just need to open a PR based on 12.0 with a cherry-pick of the commit in 14.0, then adapt the references to non-existing pages.

That I know, for my PR.
I meant to port the only missing doc `hosting_changes` to the previous versions so that it's there as well and I can leave my content as is.

AntoineVDV:
Ah ok, then you should be able to back-port the commit of that doc by creating a PR in 12.0 with that commit in it.
Still, make sure that that doc can fit in 12.0. It was not merged in 12.0 in the first place so there may have been a good reason not to have done so at the time.

jcs-odoo:
Alright #809  is now based on 12.0

PR #815, opened 2020-12-22
[IMP] purchase: update the reordering rules doc for v14

xpl-odoo:
With the introduction of V14, reordering rules went through some changes, such
as the replenishment view. In addition, due to the importance of this feature,
a much more in-depth presentation of all the options was deemed necessary.
Key points were also not mentioned, such as: which vendor will be
selected, what are quantities on the product's vendor, etc.


purchase/products/reordering.rst
```diff
@@ -1,51 +1,79 @@
...
-  :align: center
+.. image:: media/reordering-product-vendor.png
+   :align: center
+   :alt: Add vendor to a product in Odoo
 
 .. tip::
-  The replenishments will take place when the scheduler in the
-  *Inventory* module runs. By default in Odoo, the schedulers will run
-  every night at 12:00PM.
+   By default, a draft purchase order is created. However, if you have enabled *Purchase Agreements*,
```
AntoineVDV, line 29: That's 101 chars! :p 
You can safely break in the middle of \*a text between\*.

AntoineVDV (changes requested):
The commit title is perfect but the commit message is not. It should focus on *why* you are doing these changes instead of on *what* you are changing.
Once you lose the context, you wonder why it is stated that "This document was kept as reordering rules are quite important".
This doesn't need to be kilometric. The point of the commit message is to spare lots of time to future readers who will dig in the commit history to figure out why this section was added, or why this picture was updated this way, etc.

Appart that, RST looks good to me :+1:

AntoineVDV (approved):

PR #816, opened 2020-12-29
Update upgrade_your_database.rst

asc-odoo:
fixing a link

### DIFF
--- a/odoo_sh/advanced/upgrade_your_database.rst
+++ b/odoo_sh/advanced/upgrade_your_database.rst
@@ -10,7 +10,7 @@ Download and Upload your database
 =================================
 
 Download a dump of your database (from the :ref:`Builds view <odoosh-gettingstarted-builds-download-dump>`), choose the
-exact copy and without filestore options. Upload the .sql.gz dump on https://upgrade.odoo.com/database/upload and
+exact copy and without filestore options. Upload the .sql.gz dump on https://upgrade.odoo.com/upload and
 select the Testing Purpose. If you have custom code, you can choose to have it upgraded by us, or do it yourself. Once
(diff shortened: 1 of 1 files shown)


AntoineVDV:
Hi @asc-odoo, shouldn't it be fixed in 12.0 and 13.0 too?

asc-odoo:
Yes it should! just missed that my bad @AntoineVDV , for future reference is there a way to add to this pull or should I submit a seperate commit for each?

AntoineVDV:
> Yes it should! just missed that my bad @AntoineVDV , for future reference is there a way to add to this pull or should I submit a seperate commit for each?

With the "[canonical Git workflow](https://www.odoo.com/documentation/user/14.0/contributing/documentation/introduction_guide.html#use-the-canonical-git-workflow)" it would be quite easy to retarget your PR onto 12.0, as it is possible to do it directly from the GitHub interface. It would also be necessary to update the commit accordingly but it's not a big deal with git.

However; I see that this pull request is for a branch on your fork of the documentation-user project (`asc-odoo/documentation-user` being the fork, and `asc-odoo-patch-1` the branch) so I assume that you followed the "[GitHub interface workflow](https://www.odoo.com/documentation/user/14.0/contributing/documentation/introduction_guide.html#use-the-github-interface)". With this workflow, it's not easy (or even possible? not sure) to rebase the commit on 12.0.

If you are in the second case, just close this PR et create a new one from a new branch based on 12.0. Doc updates landing in 12.0 are regularly "forward-ported" to 13.0 and 14.0.

asc-odoo:
@AntoineVDV ,  you're correct I did follow the Github Interface workflow so I will close and create a new one in 12, thanks for the help I appreciate it! :)

PR #823, opened 2021-01-06
[IMP] purchase: merge and update blanket orders and calls for tenders

xpl-odoo:
Both blanket orders and calls for tenders shared a good chunk of information
(namely, enabling the feature and purchase agreements types configuration), and,
at the same time, aren't complex enough feature to have their own page, so this
commit merges both pages under a new folder "manage deals" where another commit
will add 3-way matching later on. At the same time, some of the content had to be
updated for v14 and it was necessary to provide slightly more in-depth information
across the board.


purchase.rst
```diff
@@ -8,5 +8,5 @@ Purchase
    :titlesonly:
 
    purchase/products
-   purchase/agreements
+   purchase/manage_deals
```
AntoineVDV, line 11: This needs a redirection.

purchase/agreements.rst
```diff
@@ -1,9 +0,0 @@
-===================
-Purchase Agreements
-===================
-
-.. toctree::
-   :titlesonly:
-
-   agreements/blanket_order
-   agreements/call_for_tender
```
AntoineVDV, line 9: Those need a redirection.

AntoineVDV (changes requested):
Same thing for the blank line between the commit title and message.

The commit message is good :)
This is a really small detail, but we generally avoid the first person in commit messages. "This commit introduces..." is preferred to "I introduce... and will do this..." but it's really not worth correcting now.

RST is good.

AntoineVDV (approved):

PR #824, opened 2021-01-06
[IMP] purchase: add alt text to UoM images

xpl-odoo:
Alt text added to all images, improved image filenames' readability by using dashes (-)


purchase/products/uom.rst
```diff
@@ -47,10 +48,11 @@ Back to the first example, if you purchase orange juice from your vendor in **ga
...
 (gallons) as the *Purchase Unit of Measure*, then click on *Save*.
 
-.. image:: media/uom-productconfiguration.png
+.. image:: media/uom-product-configuration.png
    :align: center
+   :alt: Configure a product's units of measure in Odoo
 
-Create ynew units of measure and units of measure categories
+Create new units of measure and units of measure categories
 ------------------------------------------------------------
```
AntoineVDV, line 56: Adapt the line length according to the title

AntoineVDV (changes requested):
Same thing for the blank line in the commit message.

RST is good 👍

Just a little precision: image filenames' readability is not really important as long as it is somewhat readable. In fact, as every little change to an image file actually **duplicates** the file in git, it wouldn't even be allowed to rename such files for that only purpose. What matters is that words within a file name are separated by dashes (-), as it improves the SEO score of the page. For most systems, this-is-a-five-words name while thisisasingleword name, as well as this_is_a_single_word.

AntoineVDV (approved):

PR #825, opened 2021-01-07
[REM] accounting: removing none existing module

Abridbus:
The module ' l10n_fr_certification ' has been integrate in the account module.
so this is not installable anymore. Removed this but kept the explanation
for the POS.
opw-2389435

### DIFF
--- a/accounting/fiscal_localizations/localizations/france.rst
+++ b/accounting/fiscal_localizations/localizations/france.rst
@@ -171,6 +171,18 @@ Odoo absolves itself of all and any responsibility in case of changes
 in the module’s functions caused by 3rd party applications not certified by Odoo.
 
 
+Re sequencing of the invoices
+-----------------------------
+
+It remains possible to re sequence the invoices but with some restrictions:
+
+1. The feature will not work when entries are previous to a lock date.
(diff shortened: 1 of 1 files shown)


Abridbus:
Adding second commit to explain the re sequencing for french customer.

As discussed with @tsb-odoo

Abridbus:
@alt-odoo 
If you can merge :D

alt-odoo:
I'm OK with the first commit, but for the resequence commit I'm not sure it should be put only in french localization as it concerns a generic feature...

@jcs-odoo I believe you are actively working on maintaining the doc, you might have a good opinion on this too :)

jcs-odoo:
I agree that if it is not restricted to a certain localization, it should be in the general documentation or Odoo Learn.

Now, I read that this situation is particularly problematic for French users, but I'm not sure to see why.
Are the restrictions written here limited to french companies -as the title of the commit suggests-, or is it standard for all countries?

Cheers,
Jonathan

jcs-odoo:
New Pull Request #940

jcs-odoo:
@Abridbus can I delete the branch `13.0-l10n_fr-remove-not-existing-module`?

PR #832, opened 2021-01-14
[IMP] localization: add "webinars" section to Chile

vbe-odoo:
l10n_cl (Chile):

The change is due to a request to add webinars we had per localization to improve the visibility of this as a tool both for customers and Odooers. Why at the beginning? This is because is the first thing you see when accessing the documentation.

This is a minor change.

### DIFF
--- a/accounting/fiscal_localizations/localizations/chile.rst
+++ b/accounting/fiscal_localizations/localizations/chile.rst
@@ -2,6 +2,13 @@
 Chile
 =====
 
+Webinars
+========
+
+Below you can find videos with a general description of the localization, and how to configure it.
+
+- `VIDEO WEBINAR OF A COMPLETE DEMO <https://youtu.be/BHnByZiyYcM>`_.
+
 Introduction
 ============
 


ren-odoo (approved):

ren-odoo:
Hi @jcs-odoo can you help us with you review on this small PR please?

Thank you 

cc. @vbe-odoo

accounting/fiscal_localizations/localizations/chile.rst
```diff
@@ -2,6 +2,13 @@
...
 
+Webinars
+========
+Below you can find videos with a general description of the localization, how to configure it and 
+what changes were made for version 14.
+
+- `VIDEO WEBINAR OF A COMPLETE DEMO <https://youtu.be/BHnByZiyYcM>`_.
+
 Introduction
 ============
```
jcs-odoo, line 13: May I suggest to put it as a note, at the very beginning of the intro, rather than making a whole section just for one link?

It would look like this:
image


```suggestion
Introduction
============

.. note::
   Here is a `webinar video on YouTube with a general description of the localization and how to
   configure it <https://youtu.be/BHnByZiyYcM>`_.
```

jcs-odoo (changes requested):
Hello @vbe-odoo and @ren-odoo !

Since the documentation is different according to the version, we usually do not mention the documentation version. It is implied that what we read is about the version selected in the menu of the documentation.
Another reason for this is that this content is duplicated automatically in new versions. Unless updated, this makes the content look outdated.

For this particular PR, since the video is about v14, it's okay to leave it based on branch 14.0. And for other docs, such as Argentina (#833 ), the change could be done on v13 since the video is in v13. The same change is then forward-ported to v14 as well (it takes a few days)

If you want to keep a separate section as you wrote it, please add a blank line after the `====` of the title.

Also, the commit message isn't good. Here is a suggestion for a correct commit message:
`[IMP] acc/localization: add "webinars" section to Chile`

Now, I realize that you probably created this Pull Request directly from GitHub and that all this might seem complex to do. If you agree with the suggestions I made, I can make the changes for you for all the docs. (#833 #834 #835 #836 )

Cheers,
Jonathan

vbe-odoo:
Hi @jcs-odoo !
Happy Friday!

Many thanks for all the recommendations! I totally agree with your comments, make sense to remove the version, and to add the ==== to make a new section.

If you can help me with these changes also in the other docs (#833 #834 #835 #836 ) I will be more than thankful!, because as you said I did t directly through a PR on GitHub.

Let me know if I need to need something else from my side.

Thanks! Bests!
--
cc @ren-odoo

Lau-Odoo (approved):

jcs-odoo (approved):

PR #834, opened 2021-01-14
[ADD] Minor change - Webinars on top l10n_co

vbe-odoo:
l10n_co (Colombia):

The change is due to a request to add webinars we had per localization to improve the visibility of this as a tool both for customers and Odooers. Why at the beginning? This is because is the first thing you see when accessing the documentation.

This is a minor change.


jcs-odoo (changes requested):
see #832

AntoineVDV:
See https://github.com/odoo/documentation-user/pull/832#pullrequestreview-569363553

PR #839, opened 2021-01-22
[FIX] upgrade: remove the contract parameter

remybar:
documentation update as we have removed the `-c` option from
the upgrade script to provide the subscription code. Now
it is directly read from the database.


kbr-odoo (approved):

jcs-odoo (approved):
all good
and as did @kbr-odoo  (thanks Keli) we add `content improvement` as a label for all modification, addition, deletion of content.
and the milestone is the version on which the branch is based.

cheers

PR #802, opened 2020-11-27
[REF] bank_synchronization: Rewrite the online synchronization docume…

csnauwaert:
…ntation.

PR #843, opened 2021-01-25
[IMP] accounting: improve doc about bank statements

FlorianGilbert:


jcs-odoo (approved):
Good for #802 too

Cheers :)

PR #844, opened 2021-01-27
Update my_odoo_portal.rst

ray-odoo:
@jubodoo

### DIFF
--- a/portal/my_odoo_portal.rst
+++ b/portal/my_odoo_portal.rst
@@ -111,8 +111,8 @@ Success Packs
 -------------
 With a Success Pack/Partner Success Pack, you are assigned an expert to provide
 unique personalized assistance to help you customize your solution and optimize
-your workflows as part of your initial implementation. These hours never expire,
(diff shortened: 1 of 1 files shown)


AntoineVDV:
Wouldn't this need to be updated in 12.0 and 13.0, in addition to 14.0?

jcs-odoo:
@ray-odoo @AntoineVDV 
Hello!
I didn't see this PR.
I did a similar change for all versions (11.0 -> 14.0 ) with #862 and #863  but rather than modifying the sentence, I deleted it and added a link to the pricing page, so that it reduces the risk of having contradictions.
What do you think?

AntoineVDV:
I don't have an opinion on the content itself. But the PR should either be closed or re-targeted on 12.0.

ray-odoo:
I have no problem with removing the information about success packs and linking to the pricing pages.

jcs-odoo:
Alright, Thanks then @ray-odoo

PR #860, opened 2021-02-19
[FIX] small fixes

jcs-odoo:


jcs-odoo:
I also added ``[REM] *: remove all demo.action buttons from the user documentation`` which is a back-port from #572  although I also deleted more action buttons than the original commit in 13.0

This makes the build of 12.0 much faster


Also, the commit ``[FIX] two rst code errors: unknown node title_reference`` creates unknown nodes in all current branches (13.0 and 14.0)

Cheers,

Jonathan

AntoineVDV (approved):
"[FIX] two rst code errors: unknown node title_reference" -> "[FIX] odoo_sh,portal: fix broken references

PR #872, opened 2021-03-04
[REM] quality: straightforward docs about quality removed

Lau-Odoo:
These pages document very straightforward screens and flows users can easily figure out without dedicated documentation. Besides, those screens are frequently reworked, requiring to maintain the related pages closely. This commit then removes low-value pages to reduce the technical debt and allow focusing on actually useful documentation pages.


jcs-odoo (changes requested):
Ola Ludvig!

Looks good to me.

Although, before merging could you update the commit message?
The main message could rather indicate that it's the whole "quality" section that is removed

And include in the second part of the message:
- the message you wrote at the beginning of this PR
- explain that the content of control points is going to be dispatched in other docs.

What about warning the PO/devs that some links in Odoo may be broken? I don't recall what was said about this.

Cheers,
Jonathan

jcs-odoo (approved):

PR #884, opened 2021-03-15
[IMP] website: rework badly worded instructions

ksc-odoo:
Made some grammatical and structure fixes throughout this documentation.

PR #885, opened 2021-03-15
[IMP] website: improved various grammatical errors

ksc-odoo:


AntoineVDV (changes requested):

ksc-odoo:
I deleted this documentation, as it falls into the category of being "irrelevant/self-explanatory/already explained in a tutorial video" - this will be the first of many to be deleted for this same reason.

AntoineVDV (changes requested):
Hi, please take a look at the discussion in https://github.com/odoo/documentation-user/pull/772 as it gives guidance on how to proceed with page deletion, regarding commits.

Also, you delete `translate.rst` but not the reference to the page in (I suppose) `publish.rst`. I believe that building your branch will yield errors/warnings which must be fixed before merging.

AntoineVDV (changes requested):
You removed the page reference but did not make the commit any clearer.
It should have the [REM] tag and explain why the page is removed since this is quite rare and not obvious at all ;)

PR #886, opened 2021-03-15
[IMP] website: cleaned up grammatical errors.

ksc-odoo:


website/publish/multi_website.rst
```diff
@@ -37,25 +37,25 @@ Then, select a theme. This new website might have an entirely different
 purpose or audience than the first one. So feel free to go for a
 different theme!
 
-Once the theme is selected, you can start to build the homepage of your
+Once the theme is selected, you can start to build the homepage for your
 website. Follow the purple drops; they will help you in the first steps.
 
-.. image:: media/multi_website08.png
```
AntoineVDV, line 43: Is this image file still referenced somewhere else? If not, it should be removed.

website/publish/multi_website.rst
```diff
@@ -37,25 +37,25 @@ Then, select a theme. This new website might have an entirely different
...
+.. image:: media/updated_website_builder.png
```
AntoineVDV, line 43: Please use `-` instead of `_` in image file names to improve the SEO.

AntoineVDV, line 43: Please use [pngquant](https://www.odoo.com/documentation/user/14.0/contributing/documentation/introduction_guide.html#contributing-pngquant) to compress this file! 1.4 MB is quite large for a simple image. Merging this would add weight to the repository itself and slow down the rendering of the documentation page.

AntoineVDV (changes requested):

website/publish/multi_website.rst
ksc-odoo: I no longer see this on line 43. I am assuming it has been removed. it was replaced by the image that I need to retitle and compress.

website/publish/multi_website.rst
ksc-odoo: will do

website/publish/multi_website.rst
AntoineVDV: Indeed the *reference* to the image was removed, but not the image file itself :p

PS: Usually we either answer a conversation or resolve it, but rarely both. This allows reading answers without expanding all closed conversations ;)

AntoineVDV:
Tip: When all remarks have been addressed, please re-request a review (🔄 button) so that I get a specific notification for that.

AntoineVDV (commented):
You don't need to change anything this time, but we usually write commit messages like this: "[IMP] website: (If merged, this commit will...) **clean** up errors & **remove** old image"

AntoineVDV (approved):

PR #888, opened 2021-03-16
[IMP] general settings: update google auth

tas-odoo:


jcs-odoo (changes requested):
Hello Tal!

Good job working on your first doc.

### Content review
I understand you're updating an existing doc, but ideally, you make it your own and rewrite it all if necessary. Have a look at recent docs. They follow our current [Content Guidelines](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html) while older docs don't.

1.  Is your content really just for v14.0? If the flow (and screenshots) are mostly the same as in 13.0, then the doc should be based on 13.0 as the changes would also be *forward-ported* to the following versions. If it is completely different, then it is okay to just work on 14.0 as you did.
2. The title of this doc is "old-fashioned". See [Content Guidelines - Titles](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#titles)
3. Document stucture: please read carefully [Content Guidelines - Document Structure](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#document-s-structure)
   - There is no intro describing what the feature is
   - Then you could have a *h2* title named "Configuration" and two *h3* titles, one about the actions to do on Google, and one for the Odoo-side
4. Images (content) see [Content Guidelines - Images](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#images)
   - There are plenty of images. Here, it isn't odoo related content so it's more okay to do so If it were Odoo-related content, then I'd say to limit the use of images when they actually help understand the situation described in text. It shouldn't replace the use of text (for example "do as the image below") nor illustrate all actions described.
   - Avoid realistic examples such as *odoo.com* as it may mislead readers.
   - Add "alt tags" (see guidelines)
   - Blur any key and prefer blurring rather than "drawing" on top.
5. if you tell people to copy-paste something they'll do it exactly and won't replace [mydomain] with their domain
6. The whole doc is described as a list, here. We don't do that, usually. We write concise paragraphs instead.

### Technical review
Please refer to the [RST Guidelines](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_guidelines.html) for most points.

1.  Break the line at 100th character. You can setup your text editor (notepad++ I believe) to display a line at the 100th character. [RST Guidelines - 100th character](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_guidelines.html#start-a-new-line-before-the-100th-character)
2. About the images
   - In general, if you replace something with an updated content, you simply do that: replace the content. So if an image needs to be updated, you keep the same filename and simply replace the image. But you may not have to do that here (see below)
   - as said in the content review: add alt tags
   - your images were not compressed with pngquant See [Introduction Guide - Perform your changes](https://www.odoo.com/documentation/user/14.0/contributing/documentation/introduction_guide.html#perform-your-changes)
   - image filenames: this wasn't described anywhere in our docs but we decided to avoid meaningless filenames for images, such as a name+number. Also, for image filenames, we use `-` instead of `_` as we do for rst filenames. So, for example, instead of `google4.png`, it could be named `google-oauth-consent.png`
   - your images should be in the same *media* folder, not in a new *media1* folder. To move files with git, use the `git mv` command (I'll explain)
   - you forgot to delete all unused images from the repository. To do so, you could use the `git rm` command or also delete the files in you Windows explorer and make sure to stage their changes (with `git add`)
   - When you write a menu path, make sure to use the `:menuselection:` directive. See [RST guidelines - menuselection directive](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_guidelines.html#use-the-menuselection-directive)
   - the identation of the images isn't aligned with the list. But I suggested in the content review section to write paragraphs rather than lists, anyway. I just thought it'd be interesting to point it out.
   - Similarly, the identation of the `:align:` options isn't aligned with the image directives. See [RST guidelines - identation](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_guidelines.html#be-consistent-with-indentation)

### Additional tips:

1.  Consistency is the key (as said in the content guidelines ;) ). I think shared with you the way I use **bold** or *italic* during our first training sessions, so a while ago. It's personnal, and you don't have to do the same but here goes anyway: I tend to use **bold** to highlight feature names or important official documents, or a first name or something like that. I use *italic* for the name of the buttons, or the name of a state.     So for example "In your **Customer Invoice** click on *Confirm*. The state is *In Payment*"
2. Use admonitions (see [RST cheat sheet - admonitions](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_cheat_sheet.html#admonitions-alert-blocks)) such as a seealso section at the end of the document. There must be an official Google documentation that you can link here, for example.

We'll discuss all this together, no worries.
Good luck ;)

AntoineVDV:
Hi, in addition to Jon's comments, you may find useful information regarding the commit guidelines in my review comments on PR #884.

AntoineVDV:
>    * image filenames: this wasn't described anywhere in our docs but we decided to avoid meaningless filenames for images, such as a name+number. Also, for image filenames, we use `-` instead of `_` as we do for rst filenames. So, for example, instead of `google4.png`, it could be named `google-oauth-consent.png`

@jcs-odoo As this guideline is slowly making its way in GitHub reviews, we should probably add a note about this in our "Contributing" documentation. Simply keep in mind that renaming unchanged image files is **_NOT_** a good idea as it would double the weight of renamed image files on the repository.

tas-odoo:
Thank you Antoine.

jcs-odoo:
@tas-odoo  For your title, may I suggest something such as 
### Integrate Google Sign-In authentication
This suggestion is based on Google's guide for devs: https://developers.google.com/identity/sign-in/web/sign-in

tas-odoo:
Hi Jonathan @jcs-odoo ,
I've made the changes.
Could I move this PR into the Final Review stage in the Functional Documentation Projects?
Thanks!

general/auth/google.rst
```diff
@@ -1,49 +1,95 @@
...
+#. Click on *Create Credentials* and select **OAuth client ID**.
+
+   .. image:: media/google-oauth-client-id.png
+      :align: center
+      :alt: Oauth client id selection
+
+#. Select **Web Application** as the Application type. Now configure the allowed pages on which you 
+   will be redirected. To achieve this, complete the field **Authorized redirect URIs**. Copy paste 
+   the following link into the box: http://mydomain.odoo.com/auth_oauth/signin (replace "mydomain"
+   with your own domain). Then click on *Create*.
```
jcs-odoo, line 77: This point is bulky and full of really important points.
If your doc is separated in paragraphs, it could be easier to present in a clearer way with a list, for example.

The URL is not well presented. Not all databases are formatted this way. Even for those that are, the link you provided won't work as there is a "s" missing in `https`.
I would write it this way 

> In the **Authorized redirect URIs** field, enter your database's domain immediately followed by `/auth_oauth/signin`. For example: `https://mydomain.odoo.com/auth_oauth/signin`

Also, consider using backticks around the URI , to make it look nicer 
\`https://mydomain.odoo.com/auth_oauth/signin` --> `https://mydomain.odoo.com/auth_oauth/signin`

general/auth/google.rst
```diff
@@ -1,49 +1,95 @@
...
+
+   .. image:: media/google-create-oauth-client-id.png
+      :align: center
+      :alt: Creating oauth client id
+
+#. Once done, you receive two infos (your Client ID and Client Secret). *Insert* your Client ID in 
+   the **General Settings** on Odoo.
```
jcs-odoo, line 84: You don't describe at all how to do it in Odoo.
Also, the image below shows the "OAuth providers" menu and not the specific Google Authentication setting.

If divided in sections/paragraphs, you could have a part about how to retrieve it in Google and another on how to add it in Odoo.

jcs-odoo (changes requested):
Hello Tal!

My main advice, especially for old docs, is: try to rewrite your doc and make it yours.
It looks like you mostly tried to "update" it. The list-form looks more like an Odoo Tutorials script rather than a doc. It looks less clear, and points that could have a list of their own look weird. Also, the "odoo" part isn't documented at all. The old documentation was like that but here, there is nothing written about the "Google Authentication" setting in odoo (even the image shows something else). So... don't update if the doc's basis isn't good. Make it yours!

I left some comments directly in the doc, and in general, I think there are too many images.
Screenshots can help understand the context, but if they're too many, it quickly overloads the page. Also, the text should describe the situation and not the image.  And if it's clear enough in the text, then maybe that removes the need for a screenshot. I'm not saying you should remove all the images of course, but take the time to evaluate if they are really helping the user.

As we discussed, you'll have to restore some images that you deleted with your commit, but that were still needed by other docs.

Additionally, there was a point in my first review that is still unclear:
is this appliable to older versions as well? If so, consider basing this commit on 12.0 or 13.0.

Finally, about the presentation as a list, you told me it couldn't really be in another form than a list. Well, I'm afraid I have to disagree. It _could_ be divided into titles such as
- Configuration
  - Google API Dashboard (including the part about selecting/creating a project)
    - OAuth consent screen
    - Credentials
  - Google Authentication on Odoo
    - Retrieve the client ID
    - ...

Apologies, as it's another long review that involves a lot of re-work. Reviewers won't be able to review all docs this deeply. That's impossible as it's way too much time-consuming. That is why it is crucial that you own your content, that you do the researches, ask experts or POs, understand it and double-check what you present.
But it's your first doc, so it's normal. I'm confident it's going to be easier for the next ones, now that you have all these comments and suggestions in mind :)

Good luck :)

jcs-odoo:
### Restore the deleted files:

First thing to do:
On your branch and run `make clean html`
You'll have a list of missing files that other docs require.
image
Copy this list somewhere so you can have the list of files to bring back to life 🧟🧟🧟

Now, to restore a file, we have to run `git checkout [commit in which the file still exists] <file>`
In your current commit, the files are deleted. In your case, the files were still okay in the commit just before.

To write "commit before the current HEAD" we can either write `HEAD^` (although it is actually `HEAD^^` on windows) or `HEAD~1`. That's the same thing. You could also copy the commit's number after having done `git log`. That works as well.

And the file path comes from the list of files you copied before.

So here are three examples that achieve the same result (on Windows), on your branch in its current state:
- `git checkout HEAD^^ general/auth/media/authori.png`
- `git checkout HEAD~1 general\auth\media/authori.png`
- `git checkout 6a779e8990b52c891bc823ccf813f65aa87d948f general\auth\media/authori.png`

You do the same for all files that need to be restored.
Run `git status` and you'll see that your restored files are staged. A simple `commit --amend` and your commit is fixed!

### change the base to another branch

You have to determine first if it's appliable or not.
Do this once you've restored the files first.

Cheers :)

jcs-odoo:
Hello @tas-odoo !
I hope you're well. Do you need something from me to finish this PR?

Cheers,
Jonathan

tas-odoo:
Not yet, but I'll probably need some help with rebasing it next week...
Many thanks Jonathan!

tas-odoo:
Hello @jcs-odoo 
I've made the changes. Please have a look once again.
Thanks a lot!

Cheers,
Tal

tas-odoo:
Hello Jonathan @jcs-odoo,
I hope everything is going well on your side.
I'd appreciate if you could have a look once again at this PR, after all the changes/fixes I've made... if it could be merged..
Thanks!
Wishing you a great week.
Tal

general/auth/google.rst
```diff
@@ -1,49 +1,116 @@
...
+-----------------------------
+
+.. _google/client-id:
+
+Retrieve the Client ID
+~~~~~~~~~~~~~~~~~~~~~~
+
+Once you have done the previous steps, two infos are received: Client ID and Client Secret. 
+
+.. image:: media/google-auth-secret-ids.png
```
jcs-odoo, line 100: This image seems enlarged and a bit blurry (this is nitpicking ;) )

general/auth/google.rst
```diff
@@ -1,49 +1,116 @@
...
+
+Odoo activation
+~~~~~~~~~~~~~~~
+
+*Fill out* your Client ID into the **General Settings** on Odoo (under **Integrations**).
+	
+.. image:: media/odoo-settings-client-id.png
+   :align: center
+   :alt: Filling out the client id in odoo settings
+
```
jcs-odoo, line 114: and then, what about the settings for each user?
image

jcs-odoo (changes requested):
Hello @tas-odoo 

I hope you are well :-)
Good job changing it all, it's starting to look good :)

Buuut... I'm afraid there is a whole part missing: how to link a google account to a user. Without this info, users may take quite a while to figure out how it works.  This was missing in the old doc as well.

When you write a doc, make sure that you get the full topic, entirely, before writing it. That's another illustration that when you take a topic, you should make it completely yours and not update it.

I took some time to investigate how to do so and here is a big hint: for old users, it implies re-sending the invitation email. And for new users can simply use their google account immediately when they first log in. But that's what I found, there is maybe another way.

------

Images are still not compressed with pngquant. I believe this was already in a previous review.

Also, regarding image files: we have [new guidelines about the media files organization](https://www.odoo.com/documentation/user/12.0/contributing/documentation/content_guidelines.html#media-files). But it's still quite new and it's not a problem if it is not applied in this Pull Request, just keep it in mind for future PRs. 

------

and other suggestions directly in the text.

Last advice: try it again on a new saas database: plenty of options/modules are enabled by default on runbot while they aren't on a fresh database.

Good luck with all this. 
I am aware it's harder and more painful nowadays as we want to raise the level of our docs.

Cheers,
Jonathan

jcs-odoo (commented):
Hello @tas-odoo 

I wrote down some feedback but I'll upload the changes on this branch.

Content-wise:
The part about linking the accounts is important but has to be done _after_ the configuration of Google Sign In.
And the last point you wrote in "Odoo Activation" is actually misleading the users.

jcs-odoo:
@tas-odoo  you can check my changes here
https://github.com/odoo/documentation/commit/b6ac6f28112b4535a0ee19d777b11f394820660e

I also rewrote the "link to google" section and put it at the end of the doc.

jcs-odoo (approved):

content/applications/general/auth.rst
```diff
@@ -7,7 +7,7 @@ Authentication
 .. toctree::
    :titlesonly:
 
-   auth/google
+   auth/google_sign_in
```
AntoineVDV, line 10: Is this really necessary? We should avoid breaking URLs in versions before `master` unless there is a real need to support it.

Here, it seems that the file is renamed only to align with the branding chosen by Google. If we keep the file name as it is now ('google' in the 'auth' folder), would it confuse the reader?

jcs-odoo: You're right.
I wanted to distinguish "sign in" to avoid confusion with the google spreadsheet doc in the same category... But it turns out it's google spreadsheet that has nothing to do with this category. I'll remove this change right now.

jcs-odoo: @AntoineVDV done

AntoineVDV (approved):
LGTM except for a few comments below.

PR #889, opened 2021-03-16
[IMP] website: minor improvements and compressed image

ksc-odoo:


website/publish/multi_website.rst
```diff
@@ -37,25 +37,25 @@ Then, select a theme. This new website might have an entirely different
 purpose or audience than the first one. So feel free to go for a
 different theme!
 
-Once the theme is selected, you can start to build the homepage of your
+Once the theme is selected, you can start to build the homepage for your
 website. Follow the purple drops; they will help you in the first steps.
 
-.. image:: media/multi_website08.png
+.. image:: media/updated_website_builder.png
```
AntoineVDV, line 43: This change was made in PR #886. That means that if the current PR is merged before, the build of the documentation will be broken as the new file name wouldn't exist yet. You can see git branches as timelines in a sci-fi movie: if you go back in time to kill your ~~grandpa~~ file name, you'll be in trouble when going back to your original ~~timeline~~ branch.

You have several options here: make the same file name change, flag this PR as a draft until the other one is merged, merge both PRs into one, rebase this PR onto the other one, etc.
But I think that the easiest option is simply to pretend the name never changed and use "multi_website08.png". Whichever PR gets merged first wins and the second one has to be adapted.

AntoineVDV (changes requested):

website/publish/multi_website.rst
ksc-odoo: Ok. Thank you so much for the detailed feedback. I truly do appreciate it. I will go with your second option here, as it seems like the easiest way to fix this issue. Hopefully it works!

PR #892, opened 2021-03-16
[REM] survey: removes unnecessary documentation

ksc-odoo:


AntoineVDV (changes requested):

AntoineVDV (changes requested):
> removes unnecessary documentation

Why is it unnecessary?

This is important to explain in the commit message body because commits are the history of the repository. Very often, we go through the commit history and filter by [TAG] or app to find out *why* a given change was made.
As dropping an entire documentation page is not anecdotic, the reason that motivated the change must be clear.

Note that the commit body is not limited in size nor in styling. You can describe your reasons as thoroughly as you want.

AntoineVDV (changes requested):
Thanks, it's almost good now.

I'd just remove the "We were told to", "we were tasked", "instructed too" and so on because:
- This is *your* task. Even if the specifications come from someone else, you ultimately agreed with those or you wouldn't have created this commit.
- When browsing the commits history, nobody cares if the change was requested by FP or the community. We just want to know why it was requested.

AntoineVDV (approved):
LGTM (for the tech side) :)

AntoineVDV (changes requested):
My bad, the images are not removed while their only references are removed.

AntoineVDV (approved):

PR #894, opened 2021-03-16
[REM] survey: remove unnecessary documentations

ksc-odoo:


AntoineVDV (changes requested):

AntoineVDV (changes requested):
> removes unnecessary files

Suggestion: `remove unnecessary documentations`

Also, why is it unnecessary?

This is important to explain in the commit message body because commits are the history of the repository. Very often, we go through the commit history and filter by [TAG] or app to find out why a given change was made.
As dropping an entire documentation page is not anecdotic, the reason that motivated the change must be clear.

Note that the commit body is not limited in size nor in styling. You can describe your reasons as thoroughly as you want.

ksc-odoo:
Merged with #892

PR #900, opened 2021-03-17
[REM] ecommerce: managing variants products 

ksc-odoo:


AntoineVDV (changes requested):
LGTM except that the commit says that you remove the managing_products page while it's the variants page that you're actually removing.

ksc-odoo:
> LGTM except that the commit says that you remove the managing_products page while it's the variants page that you're actually removing.

You are absolutely right. That is 100% my mistake. Apologies. I'll fix that issue right away. Thanks!

AntoineVDV:
Are the images still references somewhere? If not, they should be removed as well.

PR #901, opened 2021-03-17
[REM] ecommerce: removes this unnecessary doc

ksc-odoo:


Lau-Odoo:
Don't you think that this doc is straightforward and could be deleted instead? @ksc-odoo

ksc-odoo:
> Don't you think that this doc is straightforward and could be deleted instead? @ksc-odoo

Yes. I totally agree. I'll *attempt to* delete it now!

ksc-odoo:
I've gone ahead and deleted this doc. If you could please review this at your earliest convenience, that'd be great. Thanks!

Lau-Odoo (approved):

AntoineVDV (changes requested):
There is something wrong with this PR. The PR title says that it fixes something while the commit says that it removes something else. And the diff is even more confusing because you remove pages while adding new images.

jcs-odoo:
Hello @ksc-odoo and @Lau-Odoo 

I hope you are well!

**Technical review:**
- don't forget to delete the images references in the doc, as they aren't needed anymore in the repository (--> `make clean html` after deletion of these images to double-check they aren't needed somewhere else
- building the doc (= make) also shows you that this deleted file is referenced in another doc. This means that this reference must be deleted there as well as it would otherwise result in a broken link. Sometimes, it also implies that the text must be rewritten in that other doc as well. It's not the case here

**Content review:**
I agree it would be better to remove this doc as a single doc. However, its content - and particularly the tip at the end - could still be part of another bigger doc about the features that can be shown on the eCommerce website (front-end) for example. It would take two or three lines instead of a full single doc.
So if you plan to include this small point in another doc soon, I agree this doc should be deleted.

Cheers,

Jonathan

jcs-odoo (changes requested):
Hello @ksc-odoo and @Lau-Odoo 

I hope you are well!

**Technical review:**
- don't forget to delete the images references in the doc, as they aren't needed anymore in the repository (--> `make clean html` after deletion of these images to double-check they aren't needed somewhere else
- building the doc (= make) also shows you that this deleted file is referenced in another doc. This means that this reference must be deleted there as well as it would otherwise result in a broken link. Sometimes, it also implies that the text must be rewritten in that other doc as well. It's not the case here

**Content review:**
I agree it would be better to remove this doc as a single doc. However, its content - and particularly the tip at the end - could still be part of another bigger doc about the features that can be shown on the eCommerce website (front-end) for example. It would take two or three lines instead of a full single doc.
So if you plan to include this small point in another doc soon, I agree this doc should be deleted.

Cheers,

Jonathan

jcs-odoo (approved):
Hello @ksc-odoo 

Looks good to me!

I "pre-approve" this PR, but could you change the commit message before merging it, please?
You can do so by typing `git commit --amend`, change the commit message, then force push your branch.

the commit message and the pull request's title should be more understandable. It could be, for example:
`[REM] ecommerce: doc about product availability`
Also, in your commit message, you can add an explanation as to *why* you delete this doc, by adding extra lines to your commit message.

Once it's done, you can merge by selecting "rebase and merge", and, in Odoo, you can move your task to "deployed".

Cheers!

Jonathan

jcs-odoo:
Ah, I see that it is still blocked by @AntoineVDV who has additional rights! So once you've updated the commit message and PR's title, you can re-request a review from Antoine. The merge button won't be available until he approves the PR.

AntoineVDV (approved):

PR #903, opened 2021-03-19
[IMP] guidelines: add media files section in content guidelines

jcs-odoo:
As you suggested it, @AntoineVDV , here is an addition about media filenames in the content guidelines.

### DIFF
--- a/contributing/documentation/content_guidelines.rst
+++ b/contributing/documentation/content_guidelines.rst
@@ -160,6 +160,38 @@ A few tips to improve your screenshots:
    Resizing the window's width is the most important step to do as Odoo's responsive design
    automatically resizes all fields to match the window's width.
 
+.. _contributing/media-files:
+
+Media files
+-----------
+
+A **media filename**:
+
+- is written in **lower-case letters**
+- is **relevant** to the media's content. (E.g., :file:`screenshot-tips.gif`.)
+- separates its words with a **hyphen** ``-`` (E.g., :file:`awesome-filename.png`.)
+
+Each document has its own folder in which the media files are located. The folder's name must be the
+same as the document's filename.
+
+For example, the document :file:`doc_filename.rst` refers to two images that are placed in the
+folder ``doc_filename``.
+
+::
+
+  ├── section
+  │   └── doc_filename
+  │   │   └── screenshot-tips.gif
+  │   │   └── awesome-filename.png
+  │   └── doc_filename.rst
+
+.. note::
+   Previously, image filenames would mostly be named with numbers (e.g., :file:`feature01.png`) and
+   placed in a single ``media`` folder. While it is advised not to name your *new* images in that
+   fashion, it is also essential **not to rename unchanged files**, as doing this would double the
+   weight of renamed image files on the repository. They will eventually all be replaced as the
+   content referencing those images is updated.
+
 .. _contributing/alt-tags:
 
 ALT tags


jcs-odoo:
@AntoineVDV  @jubodoo  
does this seem good like this to you?

I wondered if to indicate that they had to be in the "media" folder but I assumed most users would do it correctly without reading this anyway.

Cheers

contributing/documentation/content_guidelines.rst
```diff
@@ -160,6 +160,30 @@ A few tips to improve your screenshots:
...
 
+.. _contributing/media-filenames:
+
+Media filenames
+---------------
+
+A **media filename** is written in **lower-case letters** and is composed of three parts:
+
+1. A **prefix** which is the same name as the *RST filename* in which the media is referenced. E.g.,
+   ``content-guidelines``.
```
AntoineVDV, line 171: I'm not sure that requiring to prefix the image filenames with the page name is a good practice. The context is already provided by the page itself, and it is generally recommended to describe the image in its filename, not so much the page theme.
Furthermore, Google recommends using the same keywords (as a rule of thumb) in `alt` texts as in image filenames.
Also, image filenames should comprise too many words. Start from a long *page* name and you're already screwed.

contributing/documentation/content_guidelines.rst
```diff
@@ -160,6 +160,30 @@ A few tips to improve your screenshots:
...
+2. An **identifying name** relevant to the media's content. E.g., ``filename``.
+3. The **file extension**, which is ``.png`` for most images.
+
+Use a **hyphen** ``-`` to separate the words in the media filename. E.g.,
+``content-guidelines-filename.png``.
+
+.. note::
+   **RST** filenames use **underscores** ``_`` to separate words. Replace them with hyphens for the
+   media filenames.
```
AntoineVDV, line 180: If the page name is removed from the image filename, this is no longer relevant.

contributing/documentation/content_guidelines.rst
```diff
@@ -160,6 +160,30 @@ A few tips to improve your screenshots:
...
+
+.. important::
+   Previously, image filenames would mostly be named with numbers (e.g., ``feature01.png``). While
+   it is advised not to name your *new* images in that fashion, it is also important not to rename
+   unchanged files. They will eventually all be replaced with new content.
```
AntoineVDV, line 185: IMO the *why* is as important as the *what*.

contributing/documentation/content_guidelines.rst
jcs-odoo: it's to keep a form of organization in the folder.
I understand your point of view, but from a writer's perspective, it nearly never happens that we use an image in two different documents. Also, it can become a problem for later updates as the media's usage in various docs isn't always clear to the person modifying/deleting it. 

alternatively, having a media folder per page would be another way to keep the media files organized, I guess.

about the alt thing, do you want me to add a comment about this both in the filename and alt tags sections?

AntoineVDV: > alternatively, having a media folder per page would be another way to keep the media files organized, I guess.

Yeah, or even a folder per page inside the media folder.
That means longer references in RST but at least it doesn't hurt the SEO nor the clarity of the filename.

jcs-odoo: up to you! But I think it'd be better organized indeed this way, indeed.
Also, soooo much easier to compress with `pngquant *` for all files at once :p

AntoineVDV: If really image files of two pages have nothing in common, you won't probably create them all in the same PR anyway :p

jcs-odoo: So your it would be something like this, with one media folder and sub folders:

```
├── section
│   └── media
│   │   └── doc1
│   │   │   └── something.png
│   │   │   └── something-else.png
│   └── doc1.rst

```
or multiple media folders:

```
├── section
│   └── doc1-media
│   │   └── something.png
│   │   └── something-else.png
│   └── doc1.rst
```

AntoineVDV: Either are fine, a single media folder is too IMO. Simply pick the best fit for your use-case.

jcs-odoo: A single media folder is fine, but then we come full circle ^^ :
 I would add a prefix in the filename to organize the files. That's how I do it currently.

xpl-odoo: I prefer the "one media folder and sub folders" option.
Either way, I think we should impose a single way to organize images. It makes navigating folders easier than if several ways are used - and I don't see any benefit in using multiple ways.

jcs-odoo: I'm rewriting this with the option 2 in mind (but without the `-media` after). So that would mean one .rst file + it's folder (same name)

contributing/documentation/content_guidelines.rst
jcs-odoo: right, I add the why.

jcs-odoo:
@AntoineVDV @jubodoo
Hello :)
I updated this one about media files following Antoine's comments + a new "rule" : one RST file = one media folder for this document only, as described in the doc.
If you agree with this suggestion and we merge this PR, it will change the repository's files organization.

Cheers,
Jonathan

contributing/documentation/content_guidelines.rst
AntoineVDV: That's a bad idea IMO.
This structure is already used for RST indexes. The `sales.rst` file references files inside `sales/` directory.
If you use it for media too, we'll look into the folder and find images instead of RST.

AntoineVDV: I gave it a second thought and it doesn't sound that bad actually. It's surely confusing from a technical perspective but I guess that we'll get used to it.

Another good argument in favor of the individual media folder named after the referencing RST file is the (future) new structure of the doc. After the doc apocalypse, we'll have to move content up in the hierarchy: introduction pages' content should be written in toctree pages rather than buried deep in the tree of contents.
It means that RST files that already have their homonym directory will start referencing images. Those images could then be added in those homonym directories alongside the referenced RST files (sub-pages). The folders won't be media-only folders, nor doc-only, but will start acting as source folders which is just fine.

contributing/documentation/content_guidelines.rst
```diff
@@ -160,6 +160,42 @@ A few tips to improve your screenshots:
...
+-----------
+
+A **media filename** is written in **lower-case letters** and is composed of two parts:
+
+#. An **identifying name** relevant to the media's content. E.g., ``awesome-filename``.
+
+   .. important::
+      Use a **hyphen** ``-`` to separate the words in the media filename.
+
+#. The **file extension**, which is ``.png`` for most images.
```
AntoineVDV, line 175: This is not wrong but don't you think that it may add confusion for nothing?
If someone is asked to insert "duck" in every filename I have faith that I'll never see a `is-this-correct.duckpng` file.

contributing/documentation/content_guidelines.rst
```diff
@@ -160,6 +160,42 @@ A few tips to improve your screenshots:
...
+  │   │   └── awesome-filename.png
+  │   └── doc_filename.rst
+
+
+.. note::
+   Previously, image filenames would mostly be named with numbers (e.g., :file:`feature01.png`) and
+   placed in a single ``media`` folder. While it is advised not to name your *new* images in that
+   fashion, it is also essential **not to rename unchanged files**, as doing this would double the
+   weight of renamed image files on the repository. They will eventually all be replaced with new
+   content.
```
AntoineVDV, line 197:
```suggestion
   weight of renamed image files on the repository. They will eventually all be replaced as the
   content referencing those images is updated.
```

AntoineVDV (changes requested):

contributing/documentation/content_guidelines.rst
jcs-odoo: 😅

AntoineVDV (approved):

PR #905, opened 2021-03-19
Inventory: Add updated image

dca-odoo:


jcs-odoo:
This PR is part of the PR #906 . I close it as the change should be in a single commit.

PR #907, opened 2021-03-20
[FIX] contributing: fix a little typo

tas-odoo:


jcs-odoo (changes requested):
Hello @tas-odoo 

- Please rebase the commit to 12.0 and target 12.0 with your Pull Request. Here, you are fixing the typo only for branch 14.0 while this content is the same in other versions as well.
- About the commit name: no need to write "section" before the `:`. That's always the place where to write the name of the chapter modified :)

I'll explain all this to the team during our next meeting, with this PR as an example if you don't mind.

Cheers,
Jonathan

PR #908, opened 2021-03-22
[REM] ecommerce: removed 'default taxes' from menu

ksc-odoo:
Removed the link to Accounting's "Default Taxes" from the doc menu because it falls into the "self-explanatory" category that we are trying to eliminate.


Lau-Odoo (approved):

AntoineVDV (changes requested):
> Removed the link to Accounting's "Default Taxes" from the doc menu because it falls into the "self-explanatory" category that we are trying to eliminate.

Explaining the *why* in the PR description is good as it helps with the review, but explaining it in the commit message body is even better because the PR is no longer read after that its merged.

About the change itself now: if the "Default Taxes" documentation is self-explanatory, why isn't it deleted altogether? Here you simply unlist it from the eCommerce doc but that's all.

PR #909, opened 2021-03-22
[REM] ecommerce: customer portal

ksc-odoo:


Lau-Odoo (approved):

AntoineVDV (changes requested):
[REM] **ecommerce:** removed redundant customer portal portion

While you're at it, could you also take the opportunity to clean the empty lines at the end of `ecommerce/shopper_experience.rst`? There should only be one empty line left at line 9.

ecommerce/shopper_experience.rst
```diff
@@ -1,13 +1,8 @@
 ========
-Get paid
+Get Paid
```
jcs-odoo, line 2: If your commit is about removing [REM] a doc, then this change
- shouldn't be part of this commit
- could be put in another commit

Also, I don't think it's that important. (To clarify things about capitalization, I'll talk with Julien and see if we put it in the guidelines or not)

jcs-odoo (changes requested):
@ksc-odoo @Lau-Odoo 

That is indeed quite a straightforward doc!
Could you please improve the commit message (and pull request's title along).
For example:
```
[REM] ecommerce: customer portal

+explanation
```
And see the point about the title.
I'll approve this pr after these two small changes :)

Cheers!

jcs-odoo:
@Lau-Odoo @ksc-odoo 
Sorry, I forgot again a point:

The images referenced in this doc are not deleted. They should be, if they arent' used in other docs.

PR #910, opened 2021-03-22
[REM] ecommerce: removed translation link

ksc-odoo:


Lau-Odoo (approved):

AntoineVDV (changes requested):
Your previous commit title phrasing for removing cross-doc references was good, you could go with "[REM] ecommerce: unlist translate doc from the menu" or even "[REM] ecommerce: remove cross-doc reference to website translate page".

Now, what is important to add is the *why*.
Was the referenced page out of context, obsolete, already listed on the same page?

PR #912, opened 2021-03-23
[MOV][FIX] *: place static files in their place of use

Feyensv:
Prepare the incoming documentation refactoring, ensuring the files used by a doc page
are correcly moved together with the page, avoiding broken links during the reorganisation.

Partial backport of f7bedfc
to ease forward-ports

NB: the `:download:` directive automatically manages the copy of static files in the sphinx build directory (https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-download).


AntoineVDV (approved):
I'm not sure that it'll actually ease the apocalypse's forward-ports but it must be done at some point anyway.

Feyensv:
Well, it's reducing the effective content changes we'll have to do in the apocalypse branches/PR, which helps keeping the commits & diff as focused as possible.

PR #914, opened 2021-03-23
[IMP] ecommerce: fix grammatical issues in doc about importing products

ksc-odoo:


Lau-Odoo:
Wouldn't it be interesting to add some screenshots? @ksc-odoo

AntoineVDV (approved):

AntoineVDV:
> Wouldn't it be interesting to add some screenshots?

@Lau-Odoo It could be part of a second task to enhance the clarity of this doc, don't you think? This commit is about fixing grammatical issues and it does the job.

jcs-odoo:
Hello @ksc-odoo !

I hope you're doing well :)

First of all, my deepest apologies as I'm afraid we are, once more, changing the capitalization rule thing. (It happens so often, these last days I wouldn't be surprised we changed again... Maybe with something such as "ALL LETTERS ARE CAPITALIZED!" 😜😜😜

After some debate, we agreed to go for the sentence case rule, [as described in this PR](https://github.com/odoo/documentation-user/pull/927/files). 

That's why I took the liberty to re-adapt your PR instead of asking you to re-re-re-re-rework it.
I also changed the target to 12.0 since this doc is the same in older versions + slightly changed the commit message.

PR #915, opened 2021-03-23
[IMP] ecommerce: doc about orders paid with Payment Acquirers

ksc-odoo:


Lau-Odoo (approved):

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,77 +1,82 @@
...
-What are the payment status
-===========================
-At anytime, the salesman can check the transaction status from the order.
+What Are The Different Payment Statuses
+=======================================
+At anytime, the user can check the transaction status from the order, by selecting
+*Payment Transactions* under the *Configuration* menu.
 
-.. image:: media/payment_transaction.png
+.. image:: media/payment-transactions.png
```
AntoineVDV, line 16: What happened to the previous image file? If it's no longer used it should be removed.

AntoineVDV: And yet, in this particular case, it shouldn't be removed. Instead, it should be renamed to `payment-transactions.png`.
This is because git doesn't behave like a simple system folder. Instead of just holding the most recent record of the repository content, it remembers *everything*. So, renaming a file actually adds load (equal to the weight of the image file) on the repository. If we rename image files to often, we'll end-up with a multi-GB repository that will take ages to download.

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,77 +1,82 @@
...
 you are also requested to invoice the order.
 
 
-What are the payment status
-===========================
-At anytime, the salesman can check the transaction status from the order.
+What Are The Different Payment Statuses
+=======================================
+At anytime, the user can check the transaction status from the order, by selecting
+*Payment Transactions* under the *Configuration* menu.
```
AntoineVDV, line 14: This is not true. Users need to be in debug mode to see this menu item.

AntoineVDV (changes requested):

AntoineVDV:
Side note: the ";new images" part of the commit title isn't necessary. It's part of the doc improvement itself.

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,77 +1,82 @@
...
 If you invoice based on ordered quantities,
 you are also requested to invoice the order.
 
 
-What are the payment status
-===========================
-At anytime, the salesman can check the transaction status from the order.
+What Are The Different Payment Statuses
+=======================================
+In Debug mode, the user can check the transaction status from the order, by selecting
```
AntoineVDV, line 13: This could use a link to `general/developer_mode/activate.rst`

AntoineVDV (changes requested):
There are still images that are no longer referenced while their file is not deleted. What was true for the first image is obviously true for *all* images of *all* similar PRs.

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,80 @@
...
 by a payment acquirer. This triggers the delivery.
-If you invoice based on ordered quantities,
-you are also requested to invoice the order.
+If you invoice based on ordered quantities, you are also requested to invoice the order.
 
+Checking the status of a payment
+================================
+In `Debug <https://odoo.com/documentation/user/14.0/general/developer_mode/activate.html>`_
+mode, the user can check the transaction status from the order, by selecting
+*Payment Transactions* under the *Configuration* menu.
```
jcs-odoo, line 13: Developer/debug mode is not the normal flow. If it's hidden from users, it's probably for a reason. There may be additional views hidden from most users but that doesn't mean it's the normal flow. In my understanding of SO, the payment is confirmed with an automatic note in the chatter, and if you create an invoice, the payment is directly reconciled.

I guess - not sure about it - that in older versions of odoo, the transaction was directly shown on the SO as the current screenshot suggests. These states were probably visible then.

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,80 @@
...
 
-What are the payment status
-===========================
-At anytime, the salesman can check the transaction status from the order.
-
-.. image:: media/payment_transaction.png
+.. image:: media/payment-transactions.png
     :align: center
```
jcs-odoo, line 16: fix identation + add alt tags to all images

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,80 @@
...
+.. image:: media/capture-manually.png
     :align: center
 
-To capture the payment, open the transaction from the order.
-Then click *Capture Transaction*.
+To capture the payment, open the transaction from the order in *Payment Transactions*.
+Then, click *Capture Transaction*.
 
-.. image:: media/payment_capture.png
+.. image:: media/payment-capture.png
```
jcs-odoo, line 76: This image looks quite small once the doc is built

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,80 @@
...
-Once the delivery processed, you can capture the payment from Odoo.
-This mode is only available with Authorize.net.
-
-.. image:: media/payment_capture_mode.png
+.. image:: media/capture-manually.png
     :align: center
 
-To capture the payment, open the transaction from the order.
-Then click *Capture Transaction*.
+To capture the payment, open the transaction from the order in *Payment Transactions*.
```
jcs-odoo, line 73: This is not in the hidden payment transactions menu anymore.
You can read more about it here https://www.odoo.com/documentation/user/14.0/general/payment_acquirers/authorize.html#capture-the-payment-after-the-delivery

jcs-odoo (changes requested):
Hello @ksc-odoo and @Lau-Odoo 

### Content review

You're modifying a "legacy" doc. Its current state comes from v10 from what I can see online. We have to be careful with these as they often reflect a different approach to a topic than the one we currently have. The apps and flows have changed a lot since then.

- I believe for example that the 1st point isn't relevant anymore, but I may be wrong
- The second point is eCommerce specific (although exactly similar to the Sales settings), but not specific to Payment Acquirers.
- The third point is specific to Authorize.Net, but not to eCommerce.

This doesn't mean that this page is not necessary, but I think that we should rethink its content.

People coming here may want to know how to automize the Sales process with payment acquirers, or how to check the payment status and see how to fix this.
You can explain how to do this, and add links to the relevant doc, rather than reexplain it.

- [Authorize.Net: capture payment](https://www.odoo.com/documentation/user/14.0/general/payment_acquirers/authorize.html#capture-the-payment-after-the-delivery)
- [Payment Acquirers: payment messages](https://www.odoo.com/documentation/user/14.0/general/payment_acquirers/payment_acquirers.html#messages-tab)

*Now, this PR isn't wrong per se, and is already an improvement, although still misleading for odoo's current version. So we can merge this PR anyway after a few small tweaks that I write directly as suggestions in the file, and right here below.*

- Images:
  - Missing [ALT tags](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#alt-tags)
  - Some images look really small once the doc is built. [Here are some recommendations on how to take better screenshot.](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#screenshots)
  - Also, for future docs that you rewrite, try to write paths with the [menuselection](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#document-s-structure) directive rather than describing the menu as a text.

### Technical review
A commit message indicating which doc it is about is better. No need for the v14. Also, this isn't [FIX] but [IMP] in this case For example:
`[IMP] ecommerce: doc about orders paid with Payment Acquirers`

The following are just comments, but they are not blocking.
- Leave a blank line between each command, in this case, between a title and the content. It still works like that so not a biggie.
- [Indentation](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_guidelines.html#be-consistent-with-indentation) for the `:align:` options under the image directive. But here again, it doesn't prevent the builder from working.

Good luck :)

ecommerce/shopper_experience/payment_acquirer.rst
ksc-odoo: Ah, I see what you mean. Makes sense. So, seeing as how this is only accessible through Debug mode, and you mentioned that this first point was no longer relevant anyway, I will go ahead and remove this portion of the documentation. But, I will mention how the payment is confirmed in the Chatter of an SO (via an automatic note). And, I will also mention that if they create an invoice, a payment will be directly reconciled immediately.

ecommerce/shopper_experience/payment_acquirer.rst
ksc-odoo: Great point! For the sake of efficiency, I'll simply add a quick link to this documentation, so as to avoid any unnecessary re-explaining. If this is the incorrect move, however, feel free to let me know how I should go about fixing this issue. Thanks again for all your continued patience and feedback during this all-too-important learning process. I sincerely appreciate it.

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,70 @@
...
-   If you choose this mode you are requested to select a payment journal in order to record payments
-   in your books. This payment is automatically reconcilied with the invoice, marking it as paid.
-   Select your **bank account** if you get paid immediately on your bank account. If you don't you
-   can create a specific journal for the payment acquirer (type = Bank). That way, you can track
+   If you choose this mode, you are requested to select a payment journal, in order to record 
+   payments in your books. This payment is automatically reconciled with the invoice, marking it as
+   paid. Select your **bank account** if you get paid immediately on your bank account. If you don't 
+   you can create a specific journal for the payment acquirer (type = Bank). That way, you can track
    online payments in an intermediary account of your books until you get paid into your bank
    account (see :doc:`../../general/payment_acquirers/payment_acquirers`).
```
jcs-odoo, line 60: This note is not about about generating invoices automatically.
I believe it is linked to the next point "capture payment". Also, I think that the part about creating new journals for payment acquirers isn't completely correct anymore, since you would configure which journal to use when you first activate a new payment acquirer. (also, the new journal is now created automatically by odoo, when you do that).

I would simply remove this note.
```suggestion
```

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,70 @@
...
 
+Capture payment after the delivery
+==================================
+With this mode (which is only available with Authorize.net), the order is confirmed, 
+but the amount is kept on hold. Once the delivery is processed, you can capture the payment from
+Odoo.
```
jcs-odoo, line 66: - as at line 12: add a blank line between two commands (title and paragraph)
- break the line at 100th character
- "from Odoo" --> "from the related SO"
```suggestion
Capture payment after the delivery
==================================

With this mode (which is only available with Authorize.net), the order is confirmed, but the amount
is kept on hold. Once the delivery is processed, you can capture the payment from the related Sales
Order.
```

ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,70 @@
...
-
-.. image:: media/payment_capture.png
-    :align: center
-
-With other payment acquirers, you can manage the capture in
-their own interfaces, not from Odoo.
\ No newline at end of file
+To learn more about that, simply follow the procedure explained `here <https://www.odoo
+.com/documentation/user/14.0/general/payment_acquirers/authorize
+.html#capture-the-payment-after-the-delivery>`_.
```
jcs-odoo, line 70: Never use external links to target an internal doc or a specific part of a doc. Also, try to add the text on a meaningful description of the link rather than on "here" or "click here".

Instead, you can use either a `:doc:` directive or an internal hyperlink target. (See [RST cheat sheet: hyperlinks](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_cheat_sheet.html#hyperlinks))

In this case, since the purpose of the paragraph is to refer to other docs, we can replace it with a [seealso](https://www.odoo.com/documentation/user/14.0/contributing/documentation/rst_cheat_sheet.html#seealso) section.

```suggestion
.. seealso::
   - :doc:`../../general/payment_acquirers/payment_acquirers`
   - :ref:`Payment Acquirers: Place a hold on a card <payment_acquirers/capture_amount>`
   - :doc:`../../general/payment_acquirers/authorize`
```

image

jcs-odoo (changes requested):
Hello @ksc-odoo and @Lau-Odoo ,

Apologies, I forgot about this PR. But it's the only one left in the "final review" stage in the project, so I'm re-reviewing it now.

I still feel like this doc is an update, while we want to rewrite it with a different approach. Make the full topic yours, think how you want to present it, and avoid using old docs as a basis.

- In "Checking the status of a payment," you don't describe where to go to open an order.
- [Content guidelines: grammatical tenses](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#grammatical-tenses): avoid the use of the future to describe a normal behavior.
- [Content guidelines: screenshots](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#screenshots): as explained in a previous review, most screenshots look too small for this documentation. You can see this when you build your doc.

and more comments directly in the text.

Thanks, and have a good Friday!

Jonathan

jcs-odoo (approved):

jcs-odoo:
@AntoineVDV 
I believe this PR is good to go, now (and correctly rebased to the new v14.0 )

Could you have a look at it and approve it if it's okay?

Thanks!

jcs-odoo:
@ksc-odoo 
I quickly updated the images so they fit the [new media file organization](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#media-files)

content/applications/websites/ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,63 @@
...
-To capture the payment, open the transaction from the order.
-Then click *Capture Transaction*.
+Capture payment after the delivery
+==================================
 
-.. image:: media/payment_capture.png
-    :align: center
+With this mode (which is only available with Authorize.net), the order is confirmed, but the amount
+is kept on hold. Once the delivery is processed, you can capture the payment from the related Sales
+Order.
```
AntoineVDV, line 58: Although this part is correct, I'd rephrase it a bit. The "capture mode" is being added to Adyen and Stripe as we speak and I'm pretty sure that this doc won't be updated accordingly when it's done.

Instead, why not formulate it like this:
```suggestion
If the acquirer handling the payment is configured to capture amounts manually, the order is confirmed but the amount is kept on hold. Once the delivery is processed, you can capture the payment from the related Sales
Order.
```

content/applications/websites/ecommerce/shopper_experience/payment_acquirer.rst
```diff
@@ -1,78 +1,63 @@
...
 
-With other payment acquirers, you can manage the capture in
-their own interfaces, not from Odoo.
\ No newline at end of file
+.. seealso::
+   - :doc:`../../../general/payment_acquirers/payment_acquirers`
+   - :ref:`Payment Acquirers: Place a hold on a card <payment_acquirers/capture_amount>`
+   - :doc:`../../../general/payment_acquirers/authorize`
```
AntoineVDV, line 63:
```suggestion
```
Following the previous comment, the exhaustive list of acquirers supporting manual capture should be added in the appropriate doc section.

content/applications/websites/ecommerce/shopper_experience/payment_acquirer.rst
jcs-odoo: I had no idea the capture mode was being developed. Thanks Antoine.

content/applications/websites/ecommerce/shopper_experience/payment_acquirer.rst
jcs-odoo: I'll do that on a different PR, but sure, let's remove it from here.

AntoineVDV (approved):

PR #917, opened 2021-03-24
[IMP] updates steps and adds a couple new screenshots

ksc-odoo:


Lau-Odoo:
Don't you think that this doc could be deleted? I think it is pretty straightforward 😄 
@ksc-odoo

Lau-Odoo:
I just saw that you want to push this in 12.0
So I approve it for 12.0 and 13.0, but I still think that it should be removed from 14.0 @ksc-odoo

Lau-Odoo (approved):

ecommerce/maximizing_revenue/reviews.rst
```diff
@@ -1,59 +1,49 @@
...
 .. image:: ./media/comment_post.png
    :align: center
 
 Setup
 =====
 
-Activate comments & rating from the *Customize* menu of the product web page.
+Activate comments & ratings from the *Customize* menu of the product web page.
 
-.. image:: ./media/comment_setup.png
```
AntoineVDV, line 17: Is this image still used somewhere?

ecommerce/maximizing_revenue/reviews.rst
```diff
@@ -1,59 +1,49 @@
...
-.. image:: ./media/comment_unpublish.png
-   :align: center
-
-..tip::
-    You can access the web page from the detail form by clicking the *Published*
-    smart button (and vice versa).
+To change the visibility of the post, open the product web page and click the *Visible* button
+to turn it red (*Employees Only*).
 
-.. image:: ./media/comment_webpage.png
```
AntoineVDV, line 58: Is this image still used somewhere?

AntoineVDV:
> I just saw that you want to push this in 12.0
So I approve it for 12.0 and 13.0, but I still think that it should be removed from 14.0 @ksc-odoo

If I'm not mistaken, this doc is indeed removed in 14.0. Considering this, I believe that this commit shouldn't be merged. Fixing a doc that you later delete to avoid having to fix it makes no sense, it goes directly against the idea of removing large parts of the doc in order to get rid of their inherent technical dept.

jcs-odoo (changes requested):
Hello @ksc-odoo !

While I agrre with @Lau-Odoo that most of this doc is straightforward (especially if explained in the Tutorials), I would still make sure that adding options under the "customize" section is documented somewhere before deleting this doc.

If you decide to delete this doc in v14.0, then the current commit isn't relevant anymore since our team's focus is to update 14.0 and apply these changes to previous versions, if they are appliable. Therefore, this commit would need to be completely different: the deletion of this page, based on 14.0, with the PR targetting 14.0.

Technically we _could_ merge this in 12.0 and 13.0, but our team wouldn't have done this work for 12.0 and 13.0.

In any case, the commit message isn't good as it is. Please follow the recommendations in the commit template. (that's the note that appears when you commit if you correctly set up the commit template as explained in[ the point 3 of this section of the introduction guide to contributing](https://www.odoo.com/documentation/user/14.0/contributing/documentation/introduction_guide.html#fetch-the-sources).)

I understand that fixing grammar issues was a good way to start working with Git, but updating 14.0 is still our main goal :)

Cheers,
Jonathan

PR #918, opened 2021-03-25
ID Locale Translation

Altela:
Add inventory.po for Indonesian locale translation.


mart-e:
Hi!

Thanks for the translation but we manage our translations [on Transifex](https://www.transifex.com/odoo/public), not Github.
You may see on Transifex, on the [odoo-14-doc project](https://www.transifex.com/odoo/odoo-14-doc/) that Indonesian is not present. This is because we try to favour first the translation on the main software (so [odoo-14 project](https://www.transifex.com/odoo/odoo-14) as of today). Indonesian on Odoo 14.0 is only translated in Indonesian at 46% at the moment.
If you could focus the translation there first, and then we can enable the Indonesian on the documentation once this is higher (like 75%).
Best,

Altela:
Hi @mart-e , thank you for the feedback.
I would like to contribute to the main software via transiflex but It seems like it's a paid platform that offer trial version. Is translation only made by officials (employee of Odoo) or it's also accept the community translation?

mart-e:
@Altela it's paying for us, Odoo, to have our translations there but it's free for the translators.
You can join the team and start translating there.
You can read a bit more about the process here: https://github.com/odoo/odoo/wiki/Translations
In case you have an issue/question on Transifex, don't hesite to contact me over there https://www.transifex.com/user/profile/mart_e/
Thanks for your help

Altela:
Just joined transifex, i'll read some instruction and documentation first before get going on translation. 
thanks @mart-e !

PR #919, opened 2021-03-25
[IMP] ecommerce: fixes subtle grammatical errors

ksc-odoo:


Lau-Odoo (approved):

ecommerce/maximizing_revenue/pricing.rst
```diff
@@ -1,31 +1,32 @@
...
+Currency Selector
 =================
 
-In case you sell in several currencies, you can let your customers choose
-their own currency. Check *Selectable* to add the pricelist to the website drop-down menu.
+In the case that you sell in several currencies, you can let your customers choose
+their own. Check the *Selectable* box to add the pricelist to the website drop-down menu, which
+can be found in *Pricelists* under the *Products* menu, located in the Website application.
 
-.. image:: ./media/website_pricelist_select.png
```
AntoineVDV, line 28: Is this image still used somewhere?

AntoineVDV (changes requested):

ecommerce/maximizing_revenue/pricing.rst
ksc-odoo: I don't believe so. I'll delete it right away.

jcs-odoo (changes requested):
Hello @ksc-odoo and @Lau-Odoo !

I see, KC, that you decided to modify many titles in many of your pull requests.
So let's take a step back and discuss this choice: I'm not really fond of "all first letter" capitalized titles.
This isn't in our content guidelines (yet), so it's more of a personal choice. Although I think it would be great to decide together on a capitalization style.

- I think that capitalizing all first letters makes the title heavy.
- It's not consistent with the other titles, in general. [And consistency is key](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#consistency)
  - Of course, if everything else is wrong in the doc, then we can create a pull request to fix all these docs at once. And the explanation of the commit's message would explain why we do that for all these docs. I guess this is a poor reason :p 
- There are some rules when you choose to capitalize titles, depending on the writing style that we choose. I wonder if to adopt APA, for example. If so, the title would become `How to Adapt the Prices to My Website Visitors`

Depending on the issue of our future little debate about this, you may have to update many commits! 😈

ADDITIONALLY,
[There are rules about titles in our content guidelines.](https://www.odoo.com/documentation/user/14.0/contributing/documentation/content_guidelines.html#titles).
Of course, your commit is about fixing grammar issues, so it's okay if you don't reword the title. But if you were to update this doc, the "how-to" title wouldn't be good.

-----

Now, as Antoine pointed out if you remove a reference to an image, don't forget to delete the file, then run `make clean html` to double-check that the file isn't missing for other docs, and commit.

-------

about the image
You are modifying documentation for v12.0 and following, and decided to replace an image.
The new screenshot is NOT from Odoo 12, but from Odoo 14. So that's not good at all.
So you can either
- take a screenshot from v12 to fix this commit and pull request (in this case I'd go for this option)
- or update just v14 as per our team's focus

----

Please improve the commit's title and the pull request's title as well.
Since you make plenty of grammar fixes in various branches and commits, it is hard to tell which one is which.
Since this commit also changes an image a better commit message could be

`[FIX] ecommerce: grammar and screenshot in doc about pricing`


Cheers,

jcs-odoo:
@ksc-odoo 
I added two points in my review above. I'm pinging you here just in case you read the notification email rather than the message on GitHub directly.

Cheers :)

jcs-odoo (approved):

AntoineVDV (approved):

PR #921, opened 2021-03-29
[REM] ecommerce: deletes unnecessary documentation

ksc-odoo:


AntoineVDV (changes requested):
You are already doing the same changes in #911 (except for the removal of `reviews`). Shouldn't you merge your changes in that PR and close this one instead?

PR #929, opened 2021-04-09
[REM] user-documentation: all unused images

jcs-odoo:
This commit deletes all images that aren't referenced anywhere in 12.0 content.


jcs-odoo:
We'll do this after the docapocalypse.

PR #931, opened 2021-04-12
Clarify documentation of routes

mdb-odoo:
Starting at line 259 - updated the text. This text was a mix of incorrect and confusing statements. I changed some words and sentences to correct the text and make it make more sense. Also updated the rule name to Pull & Push as that is what it is called in Odoo.

### DIFF
--- a/inventory/routes/concepts/use-routes.rst
+++ b/inventory/routes/concepts/use-routes.rst
@@ -257,19 +257,19 @@ The available rules trigger various actions. If Odoo offers *Push* and
 all of them:
 
 -  **Pull From**: this rule is triggered by a need for the product
-   in a specific stock location. The need can come from a sale order
-   which has been validated or for a manufacturing order which
+   in a specific location. The need can come from a sale order
+   which has been validated or from a manufacturing order which
    requires a specific component. When the need appears in the
-   source location, Odoo generates a picking to fulfill this need.
+   destination location, Odoo generates a picking to fulfill this need.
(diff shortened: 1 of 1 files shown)


jcs-odoo:
@Lau-Odoo 
I believe this is for you. Also, it could be based on 13.0 but I don't think mdb can do this change directly since it looks like a patch made directly on GitHub.

Cheers :)

AntoineVDV:
> it could be based on 13.0 but I don't think mdb can do this change directly since it looks like a patch made directly on GitHub.

Then it's best to close this PR and open a new one targeting the appropriate branch. cc @mdb-odoo

mdb-odoo:
@AntoineVDV The documentation for the mentioned page is the same for V13 and V14, both incorrect.

AntoineVDV:
> @AntoineVDV The documentation for the mentioned page is the same for V13 and V14, both incorrect.

Alright, do you create the PR for 13.0 then?

mdb-odoo:
@AntoineVDV I will. This was my first time doing this so I wasn't sure on the full protocol

AntoineVDV:
> @AntoineVDV I will. This was my first time doing this so I wasn't sure on the full protocol

If you know Git, it might be easier for you to create a branch on this repository. See https://www.odoo.com/documentation/master/contributing/documentation/introduction_guide.html#use-the-canonical-git-workflow.

I close this PR as you will open another one.

PR #933, opened 2021-04-13
[IMP] l10n_pe: Improve configuration instructions.rst

ren-odoo:
The detail to register Digiflow as PSE was missing, with this change the instructons will completed.


jcs-odoo (approved):
All good, thanks @ren-odoo !

PR #934, opened 2021-04-14
[IMP] user-doc: rename "odoo learn" links as "odoo tutorials"

jcs-odoo:
3 out of 4 files could also be changed in 13.0 but it's not important as it is a minor change.


AntoineVDV (approved):
--> [IMP] **\***: rename "odoo learn" links **to** "odoo tutorials"

PR #937, opened 2021-04-22
[REM] odoo_sh: remove incorrect information about cups

jev-odoo:
CUPS the server will never be installed on an Odoo.sh machine.
That being said you can use a remote printer with a cups client.
The problem lays in the sorry state of libcups dependencies so we will
try to get a reliable section in the documentation for 20.04 as soon as
we can get a testcase.

Meanwhile, it has been reported that the command line `lp` cups cli
client tool could be used as a workaround in the OCA concerned module:

https://gist.github.com/rm-jamotion/0453e91ff3e13350a1e46a11ebb02e8d

Reapply of https://github.com/odoo/documentation-user/commit/3b4c795


jev-odoo:
@Feyensv As discussed ;)

Feyensv (approved):

PR #943, opened 2021-04-29
[FIX] legal: use direct links to pdf files

mart-e:
This avoids to have URL pointing to
`https://www.odoo.com/documentation/user/<VERSION>/<LANG>/odoo_enterprise_agreement.pdf`
when switching a language

The pdf are built only once and are global per version

### DIFF
--- a/content/legal.rst
+++ b/content/legal.rst
@@ -31,7 +31,7 @@ Terms and Conditions
 
 .. |download_enterprise_en| image:: legal/img/pdf.svg
     :alt: Download Odoo Enterprise Agreement
-    :target: odoo_enterprise_agreement.pdf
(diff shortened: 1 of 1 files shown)


legal.rst
```diff
@@ -32,15 +32,15 @@ Terms and Conditions
 
 .. |download_enterprise_en| image:: _static/banners/pdf.svg
     :alt: Download Odoo Enterprise Agreement
-    :target: odoo_enterprise_agreement.pdf
+    :target: https://www.odoo.com/documentation/user/12.0/odoo_enterprise_agreement.pdf
```
Feyensv, line 35:
```suggestion
    :target: https://www.odoo.com/documentation/12.0/odoo_enterprise_agreement.pdf
```

PR #956, opened 2021-05-06
[IMP] Accounting: Update Peruvian localization

ren-odoo:
Update important explanation on Peruvian localization configuration process.

### DIFF
--- a/content/applications/finance/accounting/fiscal_localizations/localizations/peru.rst
+++ b/content/applications/finance/accounting/fiscal_localizations/localizations/peru.rst
@@ -91,8 +91,8 @@ Here are some terms that are essential on the Peruvian localization:
 - **OSE**: Electronic Service Operator, `OSE SUNAT's definition 
   <https://cpe.sunat.gob.pe/aliados/ose#:~:text=El%20Operador%20de%20Servicios%20Electr%C3%B3nicos%20(OSE)%20es%20qui%C3%A9n%20se%20encarga,otro%20documento%20que%20se%20emita>`_.
 - **CDR**: Receipt certificate (Constancia de Recepción). 
-- **SOL Credentials**: User and password provided by the SUNAT and grants access to Online 
-  Operations systems. 
+- **SOL Credentials**: Sunat Operaciones en Línea. User and password are provided by the SUNAT and
+  grant access to Online Operations systems.
 
 
 Signature Provider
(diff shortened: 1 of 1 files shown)


jcs-odoo (approved):

content/applications/finance/accounting/fiscal_localizations/localizations/peru.rst
```diff
@@ -195,6 +195,9 @@ in your configuration. In this case you need to consider:
   
 - Provide you SOL credentials.
 
+.. important::
+   When using direct connection with the SUNAT, the Sol User must be set with the Company RUT + User
```
Feyensv, line 199: shouldn't it be `SOL` user ? (Also, I don't see any explanation on what the SOL means on the page :thinking:, if it is an acronym for something, it could be interesting to specify its representation on the first use of the acronym).

jcs-odoo: You're right.

From what I understand, SOL is a bit like the Belgian Intervat. I guess that this is why it isn't explained here. So the description on line 94 explained its use well. I added the acronym detail in the same fashion as the other lines before. (I think that the `:abbr:` directive wouldn't work between `**` anyway...)

Thanks @Feyensv :)

content/applications/finance/accounting/fiscal_localizations/localizations/peru.rst
```diff
@@ -91,8 +91,8 @@ Here are some terms that are essential on the Peruvian localization:
 - **OSE**: Electronic Service Operator, `OSE SUNAT's definition 
   <https://cpe.sunat.gob.pe/aliados/ose#:~:text=El%20Operador%20de%20Servicios%20Electr%C3%B3nicos%20(OSE)%20es%20qui%C3%A9n%20se%20encarga,otro%20documento%20que%20se%20emita>`_.
 - **CDR**: Receipt certificate (Constancia de Recepción). 
-- **SOL Credentials**: User and password provided by the SUNAT and grants access to Online 
-  Operations systems. 
+- **SOL Credentials**: Sunat Operaciones en Línea. User and password provided by the SUNAT and
+  grants access to Online Operations systems.
```
Feyensv, line 95:
```suggestion
- **SOL Credentials**: Sunat Operaciones en Línea. User and password are provided by the SUNAT and
  grant access to Online Operations systems.
```

PR #960, opened 2021-05-10
[FIX] install: adapt debian and rpm package installation

d-fence:
* Change 13.0 references to 14.0
* Change Ubuntu version to 20.04 as it was the stable version at Odoo
  release time
* Remove unnecessary postgresql pre-installation as Debian and Ubuntu
  install the recommended packages by default
* Replace `root` commands by `sudo` commands as it's the preferred way
  to run admin commands
* Remove the manual installation of xlwt and num2words as the Debian
  packages now exist.
* Bump the supported Fedora version to Fedora 32.
* Add a warning about unsupported Fedora 33 (because of Python 3.9 that
  would need a separate package)

### DIFF
--- a/content/administration/install.rst
+++ b/content/administration/install.rst
@@ -107,18 +107,7 @@ Linux
 Debian/Ubuntu
 '''''''''''''
 
-Odoo 13.0 'deb' package currently supports `Debian Buster`_, `Ubuntu 18.04`_ or above.
-
-Prepare
-^^^^^^^
-
-Odoo needs a `PostgreSQL`_ server to run properly. The default configuration for
-the Odoo 'deb' package is to use the PostgreSQL server on the same host as your
-Odoo instance. Execute the following command in order to install the PostgreSQL server:
-
-.. code-block:: console
-
-  $ sudo apt install postgresql -y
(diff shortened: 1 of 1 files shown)


mart-e (approved):

content/administration/install.rst
```diff
@@ -107,18 +107,7 @@ Linux
...
-^^^^^^^
-
-Odoo needs a `PostgreSQL`_ server to run properly. The default configuration for
-the Odoo 'deb' package is to use the PostgreSQL server on the same host as your
-Odoo instance. Execute the following command in order to install the PostgreSQL server:
-
-.. code-block:: console
-
-  $ sudo apt install postgresql -y
+Odoo 14.0 'deb' package currently supports `Debian Buster`_, `Ubuntu 20.04`_ or above.
```
Feyensv, line 110: `.../content/administration/install.rst:110: WARNING: Unknown target name: "ubuntu 20.04".`

FYI `Ubuntu 20.04`_ is a link and should also have been specified/updated.
I'll fix it in another PR ;).

d-fence: Oops, you are right.

d-fence: #963 should fix that

PR #967, opened 2021-05-12
[FIX] extensions: support parallel read

Feyensv:
Now that patchqueue was removed, we can consider supporting the parallel 
read with our local extensions.

Since we only have basic extensions, not storing any data on the build 
environment, the change mainly consists of specifying the extensions as 
parallelizable.

The only specific case is the html domain, since the base html domain 
requires the support of the method merge_domaindata in parallel read 
mode.  Since we do not need to share anything between the envs for this 
extension, we can simply ignore the method.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 # Pass WORKERS=auto for parallel build
 ifndef WORKERS
-  WORKERS = 1
+  WORKERS = auto
 endif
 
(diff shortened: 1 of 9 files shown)


AntoineVDV:
Is this still a WIP?

Feyensv:
Still looking to find if html_domain can be simplified/removed :thinking:

AntoineVDV (approved):

PR #969, opened 2021-05-14
[FIX] theme: hide "Try Odoo" button on iOS

adr-odoo:
According to the app-store guidelines we are not allowed to sell from
the iOS app.

Our updates are currently blocked because of this.
In the rush, we have no choice but to hide "Try Odoo for FREE" buttons
that redirects the customer to the subscription page.
This is unfortunately not allowed by Apple who consider this
as "In-App Purchase".

Note that in the old documentation (https://www.odoo.com/documentation/user),
this button was located in the header which was hidden in the app
(see related PR).

Source: https://developer.apple.com/app-store/review/guidelines/#business

Related PR:
odoo/mobile-ios#102
odoo/mobile-ios#103

Task-id: 2483253

### DIFF
--- a/extensions/odoo_theme/layout_templates/homepage.html
+++ b/extensions/odoo_theme/layout_templates/homepage.html
@@ -4,7 +4,7 @@
             <h2 class="text-white">Odoo Documentation</h2>
             <p class="text-white">The comprehensive guide for Odoo users. Easily find the tutorial and helpful tips that suit your needs.</p>
             <div class="mt-3">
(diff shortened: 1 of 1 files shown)


johandem (approved):

AntoineVDV:
Merged in 14.0 because mobile fixes for iOS are only done as of 14.0.

Feyensv (commented):
Your change only hides the "Try odoo for free" on the homepage, don't you want to hide it in the header of all pages ?

AntoineVDV:
To quote Adrien, "the header is fine since it's hidden on iOS".

PR #970, opened 2021-05-14
[I18N] translate new doc

mart-e:


conf.py
```diff
@@ -128,7 +129,11 @@
 github_user = 'odoo'
 github_project = 'documentation'
 
-locale_dirs = ['locale/']
+locale_dirs = ['../locale/']
+sphinx.transforms.i18n.docname_to_domain = (
+    sphinx.util.i18n.docname_to_domain
+) = lambda docname, compact: docname.split('/')[1 if docname.startswith('applications/') else 0]
```
xmo-odoo, line 135: Maybe add a comment to explain the unholy mess we're wrecking here (e.g. what's the issue, how `docname_to_domain` is involved and what this monkeypatch achieve).

xmo-odoo: Or at least expound on it a bit more in the commit message since you have a commit dedicated to the issue.

Makefile
```diff
@@ -55,13 +55,10 @@ latexpdf:
...
 
-l10n:
+gettext:
 	@echo "Generating translatable files..."
-	$(SPHINX_BUILD) -c $(CONFIG_DIR) -b gettext $(SOURCE_DIR) $(BUILD_DIR)/gettext
+	$(SPHINX_BUILD) -c $(CONFIG_DIR) -b gettext $(SOURCE_DIR) locale/sources
 	@echo "Generation finished."
-	@echo "Localizing translation strings..."
-	sphinx-intl update -p $(BUILD_DIR)/gettext -l $(L10N_LANGUAGES)
-	@echo "Localization finished."
```
xmo-odoo, line 64: Is that because the files are getting resolved / fetched / updated from transifex straight from the pots, so we don't need to generate the PO files by hand?

Makefile
```diff
@@ -55,13 +55,10 @@ latexpdf:
 	cp $(BUILD_DIR)/latex/*.pdf $(BUILD_DIR)/html/
 	@echo "Build finished."
 
-l10n:
+gettext:
 	@echo "Generating translatable files..."
-	$(SPHINX_BUILD) -c $(CONFIG_DIR) -b gettext $(SOURCE_DIR) $(BUILD_DIR)/gettext
+	$(SPHINX_BUILD) -c $(CONFIG_DIR) -b gettext $(SOURCE_DIR) locale/sources
```
xmo-odoo, line 60: Does `Makefile` always set pwd to its directory?

Makefile
mart-e: exactly, we do a `tx pull` every week to get the latest translations, based on the `.tx/config` file

.tx/config
```diff
@@ -2,127 +2,52 @@
...
-file_filter = locale/<lang>/LC_MESSAGES/ecommerce.po
-source_file = locale/sources/ecommerce.pot
-source_lang = en
-
-[odoo-12-doc.expenses]
-file_filter = locale/<lang>/LC_MESSAGES/expenses.po
-source_file = locale/sources/expenses.pot
+[odoo-12-doc.finance]
+file_filter = locale/<lang>/LC_MESSAGES/finance.po
+source_file = locale/sources/finance.pot
```
xmo-odoo, line 12: No way to give transifex some sort of pattern for this? Seems like a pain in the ass to maintain and keep sync'd.

Makefile
mart-e: I think so, can you trigger `make` from another directory?

xmo-odoo (commented):
The not-renamed POs are a bit brutal, is there no way to make github find similarities between them and see them as updated renames somehow?

.tx/config
mart-e: Yep, generated by hand, super fun. Sphinx has an [option `update-txconfig-resources`](https://sphinx-sr-rs.readthedocs.io/sr/latest/usage/advanced/intl.html#using-transifex-service-for-team-translation) but wasn't working well. For odoo/odoo, I have [a script](https://github.com/mart-e/odoo-scripts/blob/master/transifex/export-translations2.py#L72) to generate it.
Hopefully, this file content is not supposed to change inside a version.

mart-e:
> The not-renamed POs are a bit brutal, is there no way to make github find similarities between them and see them as updated renames somehow?

Ok, I have squashed the removal and regenation in the same commit, it seems _a bit_ better but a lot of content has been reorganised so all the path inside the .po files changes anyway

PR #981, opened 2021-05-20
[FIX] Corrected cr.execute example

raf-odoo:
For SQL parameters `execute` takes a single argument with a list, tuple
or dict, not multiple arguments.

`execute` isn't explicitly documented here, and that might have been on purpose, so I did not expand further on its arguments. That being said, if we were to document it properly, this might be a good spot. Thoughts?


raf-odoo:
This a port of https://github.com/odoo/odoo/pull/61718 which had gotten lost.

AntoineVDV (approved):

PR #983, opened 2021-05-21
[FIX] orm: readonly is True for related fields.

yelizariev:
Since Odoo 12: https://github.com/odoo/odoo/commit/52a8ed3c0c72b21baf589bc9b91a030d9da661a8

Also this adds a note about `copy` attribute.

### DIFF
--- a/content/developer/reference/orm.rst
+++ b/content/developer/reference/orm.rst
@@ -351,6 +351,17 @@ Add the attribute ``store=True`` to make it stored, just like computed
 fields. Related fields are automatically recomputed when their
 dependencies are modified.
 
+.. tip::
+
+    You can specify precise field dependencies if you don't want
+    the related field to be recomputed on any dependency change::
+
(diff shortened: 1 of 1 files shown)


content/developer/reference/orm.rst
```diff
@@ -340,16 +340,17 @@ if all fields in the sequence are required), ``groups``, ``digits``, ``size``,
 ``context``. All semantic-free attributes are copied from the source
 field.
 
-Until specified explicitly, related field has ``readonly=True`` and
-``copy=False``.
+By default, related fields are:
```
Feyensv, line 343: I thought it was better to specify all the defaults for related fields together (and fix the misleading note).
If it's ok for you, I'll squash the commit and merge it ;) (while keeping your authorship ofc).

yelizariev: Looks good :+1:

Feyensv (approved):

PR #988, opened 2021-05-25
[IMP] core: Revamp odoo.fields.Command doc

Julien00859:
Thanks to the update of sphinx, we can use `:members:` to document an
`IntEnum` class.


Feyensv (approved):

AntoineVDV:
@Julien00859 Noice! :+1: I believe that we should also mention the values behind the commands. e.g., `(6, 0, ...)`. That's because we still find such values in the code and new developers have a hard time figuring them out.

xmo-odoo:
@AntoineVDV assuming the documentation is already working using members, it seems to be showing the discriminant values right now:
Screenshot 2021-06-01 at 11-48-46 ORM API — Odoo master documentation
Might be nice to link the "command" part of command triples to the relevant attribute for clarity, though it seems fairly low-stakes, and would require stripping the "literal" (inline code) bits as Sphinx inline markup can't be nested.

PR #989, opened 2021-05-25
[IMP] manifest: add note about 'base' module in depends

yelizariev:


DavidFesquet (approved):

content/developer/reference/module.rst
```diff
@@ -21,7 +21,7 @@ dictionary, where each key specifies module metadatum.
     {
         'name': "A Module",
         'version': '1.0',
-        'depends': ['base'],
+        'depends': ['mail'],
```
Feyensv, line 24: Specifying `base` as depends is not wrong, it's even useful (cf. other comment).
Furthermore, I'd prefer specifying a 'dummy' depends in an example than a true depends that would have an effective impact if they don't modify the line.

content/developer/reference/module.rst
```diff
@@ -79,6 +79,8 @@ Available manifest fields are:
 
     When a module is installed, all of its dependencies are installed before
     it. Likewise dependencies are loaded before a module is loaded.
+
+    All modules depends on module ``base`` even if it's not specified explicitly.
```
Feyensv, line 83: 'Depending' on a module and specifying it as a `depends` are not really the same.
```suggestion
.. note::

    Modules `base` and `web` are always installed in any Odoo instance.
    But you still need to specify them as dependencies to make sure your module is updated when `base` or `web` is updated.
```

Feyensv (changes requested):

content/developer/reference/module.rst
yelizariev: Not sure that it's correct for `base` module

>     But you still need to specify them as dependencies to make sure your module is updated when `base` or `web` is updated.

It's correct for `web`, but would we need to update module on updating `web` module?

Feyensv: > It's correct for `web`, but why would we need to update module on updating `web` module?

There is one `web` which should be a `base` :D

If I understand correctly and you're speaking about 'base', unless your module doesn't rely on/modify any `res/ir` model, it makes sense to update it alongside `base` updates :thinking:.

yelizariev: Sorry, I meant

It's correct for **base**, but why would we need to update module on updating **web** module?

Feyensv: `web` still contains/modifies some python models, and some core views.
If the xml id of a core web view (html_container, external_layout, internal_layout, ...) is modified, I'd expect a template based on it to raise on update (and for that, `web` has to be specified as `depends`).

But the message could be more specific, to say that `web` depends is only useful iff the module references/modifies `web` content.

The main point I wanted to convey is that even if both base and web are auto-installed, their updates do not trigger an update of all modules =).

yelizariev: Updating `base` module will trigger updating all modules no matter do they have `base` explicitly in ``depends``.

If we reduce the statement to `web` only, we get:

> But you still need to specify `web` as dependencies to make sure your module is updated when `web` is updated.

it's kind of correct, but not unclear. You can say it about any module

> But you still need to specify `hr_fleet` as dependencies to make sure your module is updated when `hr_fleet` is updated.

In short, it's not cleary why (or when) we need to be sure that module is updated when web is updated. In most cases it's not needed.

Feyensv: > Updating base module will trigger updating all modules no matter do they have base explicitly in depends.

In master, it's not true anymore, that's why I was debating from the beginning :sweat_smile:.
But indeed, I just tested for 14.0 and it's correct.

Since it's not true anymore in master, I believe specifying this in the doc isn't really useful (it doesn't bring much and it's not an error to specify base as a depends).

yelizariev: > In master, it's not true anymore, that's why I was debating from the beginning 

I didn't know that. What do you think about keeping your version, but without `web` module?

image


I've updated the PR

content/developer/reference/module.rst
```diff
@@ -79,6 +79,11 @@ Available manifest fields are:
 
     When a module is installed, all of its dependencies are installed before
     it. Likewise dependencies are loaded before a module is loaded.
+
+.. note::
+    Module `base` is always installed in any Odoo instance.
+    But you still need to specify it as dependencies to make sure your module is updated when `base` is updated.
```
Feyensv, line 85:
```suggestion
    But you still need to specify it as dependency to make sure your module is updated when `base` is updated.
```

Feyensv (approved):

PR #995, opened 2021-05-26
hide searchbar on scroll

edi-odoo:
On mobile, the searchbar is fixed and overlays the content because its parent (header) is fixed.
In order to not have a duplicate of the searchbar for mobile, I decided the best option was to hide the searchbar on scroll.
Added JS to listen to scroll.

### DIFF
--- a/extensions/odoo_theme/static/js/menu.js
+++ b/extensions/odoo_theme/static/js/menu.js
@@ -47,14 +47,21 @@
         }
     };
 
+     document.addEventListener('scroll', () => {
+         // Allow to hide the searchbar when the page is scrolled in mobile.
+         _flagHeaderWithScrollPosition();
+     });
+
     /**
-     * Mobile: Toggle open/close sidebar on click of nav button (&& on swipe left to right?).
-     *
-     *
-     * `o_active_toc_entry` class, and their related (parent) TOC entry list (<ul> elements) receive
-     * the `show` (Bootstrap) class.
-     * Also, the deepest TOC entry receives the `o_deepest_active_toc_entry` class, and its child
-     * Sidebar receives the `o-mobile-show` class.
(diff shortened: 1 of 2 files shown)


extensions/odoo_theme/static/js/menu.js
```diff
@@ -57,4 +57,18 @@
...
+
+     document.addEventListener('scroll', () => {
+        let scroll = this.scrollY;
+        const header = document.querySelector('.o_main_header');
+        if (scroll >= 1) {
+            header.classList.add('o_header_scroll');
+        } else {
+            header.classList.remove('o_header_scroll');
+        }
+     });
```
AntoineVDV, line 72:
```suggestion
     document.addEventListener('scroll', () => {
         // Allow to hide the searchbar when the page is scrolled in mobile.
         _flagHeaderWithScrollPosition();
     });
     
    /**
     * Add/Remove the class `o_header_scrolled` on the header according to the scroll position.
     */
     const _flagHeaderWithScrollPosition = () => {
         const header = document.querySelector('.o_main_header');
         if (this.scrollY > 0) {
             header.classList.add('o_header_scrolled');
         } else {
             header.classList.remove('o_header_scrolled');
         }
     });
```
- Extract this behavior in a separate function so that it's easier to add other scroll handlers.
- Format the docstring according to the guidelines.
- Use `const` when the variable is never assigned another value. If the variable is only used once, it doesn't need to be assigned at all (except for clarity reasons).
- `> 0` is more obvious than `>= 1` IMO.
- Use 4 spaces tabulations.
- Suggestion: replace `o_header_scroll` with `o_header_scrolled` to denotated a current state.

AntoineVDV (changes requested):
Suggested commit title: "[FIX] odoo_theme: hide searchbar on scroll"

AntoineVDV (approved):
Works much better than expected! Great work :+1:

PR #1012, opened 2021-06-04
[FIX] odoo_theme: highlight all occurrences of a same entry in the menu

AntoineVDV:
When a page is referenced multiple times in the global toctree, only its
latest TOC entry was highlighted (and expanded if it's a TOC page).
This commit allows a page to be referenced multiple times at any level
of the global TOC while correctly highlighting all of its TOC entries.

task-2551473

### DIFF
--- a/extensions/odoo_theme/static/js/menu.js
+++ b/extensions/odoo_theme/static/js/menu.js
@@ -21,22 +21,30 @@
      * TOC entries (<li> elements) that are on the path of the displayed page receive the
      * `o_active_toc_entry` class, and their related (parent) TOC entry list (<ul> elements) receive
      * the `show` (Bootstrap) class.
-     * Also, the deepest TOC entry receives the `o_deepest_active_toc_entry` class, and its child
-     * TOC entry list receives the `show` class.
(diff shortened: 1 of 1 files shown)


AntoineVDV:
@jcs-odoo Although multiple page references in TOC is now supported, it should still be discouraged. Many other (future) features will either not look good or behave weirdly for such pages. This fix was carried on nonetheless because it'll help contributors see that they screwed up.

I believe that we should now explicitly discourage TOC duplicates in the contributing guidelines. What do you think?

jcs-odoo:
> @jcs-odoo Although multiple page references in TOC is now supported, it should still be discouraged. Many other (future) features will either not look good or behave weirdly for such pages. This fix was carried on nonetheless because it'll help contributors see that they screwed up.
> 
> I believe that we should now explicitly discourage TOC duplicates in the contributing guidelines. What do you think?

I completely agree, and this has already been discussed with the content team.
We can, of course, add it in the guidelines as well, but I believe that once the structure is cleaned, it will be unlikely that writers add TOC duplicates afterward.

PR #1013, opened 2021-06-04
[FIX] Indentation was inconsistent in nginx config

os-ia:
All indentation is now consistent to two spaces.

This needs to be forward-ported to 13 and 14.

<table>
<th>Before
<th>After
<tr>
	<td>before
	<td>after

</table>

### DIFF
--- a/content/administration/deployment/deploy.rst
+++ b/content/administration/deployment/deploy.rst
@@ -296,59 +296,59 @@ in ``/etc/nginx/sites-enabled/odoo.conf`` set:
 
   #odoo server
   upstream odoo {
-   server 127.0.0.1:8069;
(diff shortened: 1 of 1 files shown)


AntoineVDV:
This could probably target 14.0 directly.

os-ia:
> This could probably target 14.0 directly.

So, should I do anything about this or is this on you guys?

AntoineVDV:
> > This could probably target 14.0 directly.
> 
> So, should I do anything about this or is this on you guys?

Given that you created this PR from a fork, the easiest would be for you to close this PR and simply another one with the same diff in 14.0 ;)

os-ia:
Sure, no problem :)

PR #1016, opened 2021-06-05
Wizard example fix

GabrielAsqui:


Feyensv (approved):
Indeed, the example was wrong :+1: 
Thanks for the contribution !

PR #1021, opened 2021-06-08
[REM] content/applications/*: delete unused images

jcs-odoo:
For the reviewer and forward-port responsible: some unused images in 12.0 are modified and "re-used" in 13.0 (doc about ingenico and six)

So I'd say it's maybe better not to port these deletions to 13.0 but to use the following PR for the other versions. What do you think? (note that there is no issue porting 13.0 to 14.0)

PR 13.0 = #1022 
PR 14.0 = #1023 

@AntoineVDV


Feyensv:
IMHO, we should keep this kind of cleanup tasks/PR's for 14.0 & master, no need to clean those images in 12/13, it doesn't cost anything (and as you said, could break some things with forward-ports, which only gives us more job, for "nearly" nothing).

(but let's hear what @AntoineVDV thinks on the subject, if he's ok for the cleanup in older branches, then fine)

AntoineVDV:
> IMHO, we should keep this kind of cleanup tasks/PR's for 14.0 & master, no need to clean those images in 12/13, it doesn't cost anything (and as you said, could break some things with forward-ports, which only gives us more job, for "nearly" nothing).

Agreed.

jcs-odoo:
> > IMHO, we should keep this kind of cleanup tasks/PR's for 14.0 & master, no need to clean those images in 12/13, it doesn't cost anything (and as you said, could break some things with forward-ports, which only gives us more job, for "nearly" nothing).
> 
> Agreed.

image

PR #1026, opened 2021-06-09
[IMP] *: exclude empty TOC pages from search results

AntoineVDV:
When searching for a string that can be found in a TOC page's content,
that page is listed in the search results. This is not a wanted behavior
because TOC pages without the 'show-content' metadata directive are not
supposed to be accessible.

This commit adds a 'nosearch' metadata directive to all TOC pages
without a 'show-content' metadata directive, hence removing them from
search results.

task-2527419


AntoineVDV:
@jcs-odoo FYI

content/contributing/documentation/rst_guidelines.rst
```diff
@@ -1,3 +1,5 @@
+:nosearch:
+
```
Feyensv, line 2: This one contains a `.. toctree::` but isn't a toc file :D

Feyensv (approved):
Didn't verify all of them manually, but seems correct :ok_hand:.
(I happened to see the guidelines specific case by chance while comparing search results :D)

jcs-odoo:
Thanks for this :)
I'll make sure `nosearch` is included when necessary

AntoineVDV:
This has been (painfully) forward-ported up to master. Now, we need to be careful to remove the :nosearch: metadata when adding the :show-content: metadata to a TOC page.

PR #1028, opened 2021-06-09
[FIX] Indentation in nginx config is now consistent (two spaces)

os-ia:
As discussed in #1013, this is the same diff but targeting v14.0 (taking into account the changes in the config, like the `ssl_protocols`.


d-fence (approved):

Feyensv:
Thanks for the contribution :thumbsup:

os-ia:
Just a heads up: This is still only applied on V14.0 docs

PR #1031, opened 2021-06-14
[I18N] translate theme

mart-e:


AntoineVDV (approved):
Noice :+1: 

It was about to say that we should explicitly translate the default of `html_title` in `conf.py` but it seems to be already wrapped in the translation function by Sphinx. :raised_eyebrow:

PR #1032, opened 2021-06-15
[IMP] l10n_mx: Add extra paragraph in Warning (MXN currency)

vbe-odoo:
This is a result of a feedback we've received, in previous MX DOC versions we were specifying that the company/main currency should be MXN.

This is good to have it as a warning, we have as a warning in previous DOV versions but now we add some extra info as well (ref: https://www.odoo.com/documentation/13.0/applications/finance/accounting/fiscal_localizations/localizations/mexico.html#set-your-legal-information-in-the-company)

### DIFF
--- a/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
+++ b/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
@@ -119,9 +119,10 @@ In the resulting form, put your full address (including zip code), RFC (VAT numb
 the data.
 
 .. important::
-   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore,
-   Odoo does not provide features to manage an alternative configuration. If you want
-   to manage another currency, let MXN be the default currency and use price list instead.
+   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore, Odoo
+   does not provide features to manage an alternative configuration. If you want to manage another
+   currency, let MXN be the default currency and use a :doc:`pricelist
+   </applications/sales/sales/products_prices/prices/pricing>` instead.
 
 .. warning::
    Make sure that in the address, for the Country field, "Mexico" is chosen from the list of
--- a/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
+++ b/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
@@ -119,6 +119,10 @@ In the resulting form, put your full address (including zip code), RFC (VAT numb
 the data.
 
 .. warning::
+   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore, 
+   Odoo does not provide features to manage an alternative configuration. If you want to manage
+   another currency, let MXN be the default currency and use price list instead.
+   
    Make sure that in the address, for the Country field, "Mexico" is chosen from the list of
    countries that Odoo shows, because if it is entered manually there is a risk of creating a "new
    country" in the system, which it will result in errors later when the CFDIs are generated.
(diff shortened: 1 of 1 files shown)


vbe-odoo:
Hi @jcs-odoo  can you help me with this small change?
Thanks!
cc' @ren-odoo

jcs-odoo (commented):
Hello @vbe-odoo 

Wouldn't it make more sense to put this point as a separate point? Maybe in an `important` block, before the "tip"?

Also:
- break the line after the 100th character
- "... and use **a pricelist** instead"

`
.. important::
   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore, Odoo
   does not provide features to manage an alternative configuration. If you want to manage another
   currency, let MXN be the default currency and use a pricelist instead.
`

vbe-odoo:
Hi @jcs-odoo !
Thanks for the feedback!
Agree, it's a good idea to use the `.. important::` section in a different block.
But I would prefer to have it before the `.. warning::` like follows:

.. important::
   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore, 
   Odoo does not provide features to manage an alternative configuration. If you want
   to manage another currency, let MXN be the default currency and use price list instead.

.. warning::
   Make sure that in the address, for the Country field, "Mexico" is chosen from the list of
   countries that Odoo shows, because if it is entered manually there is a risk of creating a "new
   country" in the system, which it will result in errors later when the CFDIs are generated.

What do you think about that? (this is because I want the user to see this as the 1st warning, important info)

Thanks!
Bests!
cc' @ren-odoo

jcs-odoo:
Sure! @vbe-odoo 
Can you edit it, then, please? :)

If I remember correctly, in another PR, you edited it by adding another commit, right?

If not, I can do it tomorrow.

Thanks

vbe-odoo:
Hi @jcs-odoo 
Thanks for the answer!
I already did a commit with the changes.
Let me know your feedback!

Bests!
cc' @ren-odoo

content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
```diff
@@ -118,6 +118,11 @@ click on *Update information* under your company name.
 In the resulting form, put your full address (including zip code), RFC (VAT number), and the rest of
 the data.
 
+.. important::
+   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore,
+   Odoo does not provide features to manage an alternative configuration. If you want
+   to manage another currency, let MXN be the default currency and use price list instead.
+
```
jcs-odoo, line 125: Thanks for the change @vbe-odoo 

It seems you forgot to apply the two other points in my previous review:
- 100th character line break (see [RST guidelines](https://www.odoo.com/documentation/14.0/contributing/documentation/rst_guidelines.html#start-a-new-line-before-the-100th-character))
- write "and use **a pricelist** instead" instead of "and use price list instead"

Additional suggestion:
- "pricelist" could reference the appropriate doc (see in the suggestion below)

```suggestion
.. important::
   From a legal point of view, a Mexican company must use the local currency (MXN). Therefore, Odoo
   does not provide features to manage an alternative configuration. If you want to manage another
   currency, let MXN be the default currency and use a :doc:`pricelist
   </applications/sales/sales/products_prices/prices/pricing>` instead.

```

Thanks :)

jcs-odoo (changes requested):

content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
vbe-odoo: @jcs-odoo Oh sorry! my bad! I will commit your suggestion!
And good idea to add the link to the functional DOC of payment terms 💯 
cc' @ren-odoo

jcs-odoo (approved):
lgtm

@odoo/doc-review in this case, do I have to squash the commits for the author, or do you simply select squash and merge?

Thanks!

AntoineVDV (approved):

AntoineVDV:
> do I have to squash the commits for the author, or do you simply select squash and merge?

No worries, squash & merge is always an option as long as a PR doesn't have 2+ independent (meaningful) commits.

PR #1035, opened 2021-06-16
[ADD] l10n_mx_reports: XML Polizas Export

lordkrandel:
This is a legal requirement when one of the 2 following scenarios are 

- The SAT (Mexican government) asks to check the JEs to the company to audit (compulsory). It can happen to any company at some point the bigger the higher the chances of getting asked. Our MX office could be a case that will need this in the future.
- When the company asks for compensation or return from the SAT (Government)

An export of all Journal Entries from a specific period of time is needed, exported in an XML so later this can be uploaded to the SAT webpage.

Target Versions: v14 & Master

Task link: https://www.odoo.com/web#model=project.task&id=2479563

### DIFF
--- a/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
+++ b/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
@@ -1308,6 +1308,179 @@ reporting a balance 0.
    :align: center
    :alt: Check BS closing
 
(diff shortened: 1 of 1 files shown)


lordkrandel:
Images are not there - still WIP.

jcs-odoo:
Hello! What's the status of this draft PR?

PR #1034, opened 2021-06-16
[IMP] legal: remove empty/useless pages

Feyensv:
Task ID - 2361085

PR #1046, opened 2021-06-25
fix licence including

yelizariev:


odony:
thanks, fix the commit message?

yelizariev:
@odony done

odony (approved):
Thanks, that was missed in #1034 apparently.

We need unit tests for the doc @Feyensv ;-)

odony:
Rebuilding the doc right now to restore the license :+1:

PR #1048, opened 2021-06-25
[IMP] accounting/l10n: add tip about fisaly registration in "germany"

jcs-odoo:
small addition request by @trj-odoo


jcs-odoo:
@AntoineVDV  very small one before your forward port if it isn't too late.
Thanks :)

AntoineVDV (approved):

PR #1051, opened 2021-06-30
Master template guideline seb

seb-odoo:


content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+- `o-` is a must have Odoo prefix.
+- `ComponentName` is the PascalCase name of the current component/template.
+- `-` is a mandatory hyphen to separate the component name and the part name.
+- `partName` is the camelCase name of the child element referred by this `name` attribute.
+
+Attributes
+~~~~~~~~~~
+
+All the attributes of an element should be on the same line as the element itself. No character limit per line.
```
ged-odoo, line 1033: not a fan. it's much easier to review the content of a diff if I can see what changed precisely, and it's easier to see that if attributes/props are on several lines. But since it is left to the appreciation of the reviewer, I guess I don't care either

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+.. note::
+    For extremely long lines, it is left to the appreciation of the reviewer to accept multi-line attributes within an element (but it is not mandatory).
+
+    In that case each attribute should be on its own line, with an indentation of 4 spaces relative to the element itself. The closing of the element (either `>` or `/>`) should be on its own line as well, with the same indentation as the corresponding opening character.
```
ged-odoo, line 1038: ok for this. makes sense

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+The following order is applied for attributes:
```
ged-odoo, line 1040: really??? I don't want to have to reread this page each time I review a template. this is extremely way too much.

ged-odoo: maybe ok as a suggested standard order, but not as an obligation.

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+.. note:: `id` should not be used unless absolutely necessary (such as for making an anchor, or as the `for` target of a `label`), and then it should be uniquely generated.
+
+Style
+~~~~~
+
+Prefer using Bootstrap components and utilities over anything else, by applying their classes on the template. Custom style should only be used if absolutely necessary.
+
+Inside the `class` attribute itself, the following order should be respected:
+
+ #. Custom template class (only if necessary for adding custom style).
```
ged-odoo, line 1063: same as above. this is way too much. This does not make a template easier to review. It will make life for everyone harder without a good reason. This should be just a suggestion, but nothing more

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+ #. Bootstrap component classes (eg. `btn`, `card`).
+ #. Bootstrap layout classes (anything that changes the size or position of the element or its children, eg. `d-flex`, `container`, `mt-2`).
+ #. Bootstrap style classes (eg. `text-muted`, `bg-primary`).
+
+.. warning:: Before writing custom style you are expected to know all existing Bootstrap components, utilities, mixin and variables, as well as all of Odoo's extra components, utilities, mixin and variables.
+
+If necessary a `css` or `scss` file should be used, one per component/template. Inside that file, the style rules should be as flat as possible and target specific parts of the component/template.
```
ged-odoo, line 1070: I disagree here actually. Doing so means that we have a flat structure on the css side and a tree structure on the html side. This is a mismatch, which is annoying.  The scss way to represent embedded structure makes it closer to the html structure, which is very useful.  Using a flat structure means that you need to edit two possibly very different place when you do stuff like position absolute/relative or displayflex (which are usually done on a parent/children and are cases where the parented relationship has some relevance).

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+
+The naming guideline for targeting parts of the template with CSS classes is: `o-ComponentName-styleName` where
```
ged-odoo, line 1073: as said above, this is horrible. The overhwelming convention in css world is to only use hypens and no capital letter.

ged-odoo (commented):
I am sorry to be negative, and I hope this review will not be taken in a personal way, but I think this is a bad idea, it does not solve much, it will cause a lot of issue if enforced, so it will not be enforced, in which case it is useless.

content/developer/reference/guidelines.rst
AntoineVDV: This kind of guideline usually goes the other way around: make code clearer and more easily reviewable at the expense of the number of lines.

Funnily enough, you didn't follow this repository's guidelines regarding the character limit per line in this PR.

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+Structure
+~~~~~~~~~
+
+.. note:: These rules are in place because the general structure of the template is considered more important than the details of each line.
+
+- Each block element should be on its own line.
+- Small inline/style/text elements can be placed either on the same line as their parent or on their own line (depending on their importance and size).
+- Indentation of 4 spaces is applied between a parent and its children elements. The closing tag of a parent should always be at the same indentation level as its opening tag.
+- It is preferable to use instructions (such as `t-if`) directly on the element on which they apply (as opposed to having them on a separate `<t>` element).
```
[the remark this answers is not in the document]
aab-odoo: There's a rationale for this. Extending a template behaves differently whether the original template has a `<t t-name>` or (for instance) `<div t-name>` root element. In the later case, the extension may replace the root node tag.

seb-odoo:
Hello, no offense taken, I'm surprised to see so many negative comments now, because according to AL everybody was verbally in agreement for the most part, but I guess that was a miscommunication.

At least I'm glad that you looked into it, hopefully we can work towards something that works for all of us.

I'll address comments separately.

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
seb-odoo: For the `t-name` I agree, it is not part of the guideline but it should be added. The template definition must be on its own `<t>` tag.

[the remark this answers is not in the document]
seb-odoo: For other `t-` instructions, our current code actually does like you suggest and like you did too, which is to have them separately. But this approach of having the conditions standalone make the template look like code/instructions instead of like a declarative render.

The goal of the guideline is to ensure a flatter HTML, and have sibling elements in the result actually be sibling in the template. The fact that some elements are shown or not depending on some conditions is actually a detail of themselves, hence the proposition to have them as "attribute" instead of as "structure".

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+Naming
+~~~~~~
+
+The root element and each other *important* element of the template should have a `name` attribute to ease the readability of the template.
```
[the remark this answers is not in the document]
seb-odoo: The problem with using comments is that they are not targetable in tests or xpath (not that I like doing them, but it helps the community too), and also they are not present in the resulting DOM for easier debugging.

The problem with using class for naming, if we use bootstrap or similar approach to style, is that class attribute becomes extremely noisy, and the name is for us immensely more important to understand the template than many style values.
Therefore a new separate attribute is proposed.

But as you noted it is optional so I don't think it matters too much.

content/developer/reference/guidelines.rst
seb-odoo: @AntoineVDV thank you for the link, I didn't know this page existed. I'll make a pass to apply those once we are ok with the content.

@ged-odoo I agree with you here, it's a tradeoff between ease of diff and better overview of the template.
But both can depend on tooling so it's not that important either way if/once we have proper tools.

Same as the other point, as this one is left to appreciation of reviewer, it doesn't really matter.

content/developer/reference/guidelines.rst
seb-odoo: If we can agree on the guideline, then we can work on tools to validate it and apply it automatically.

The goal is not to have to remember the order by hand and waste time to enforce it manually.

If we can agree with the order on principle, that is all that matters for now I suppose.

content/developer/reference/guidelines.rst
seb-odoo: Same as above too, ideally this should be automated and we would never have to worry about it.

content/developer/reference/guidelines.rst
seb-odoo: I wouldn't mind to leave this to the appreciation of the reviewer too, as long as I'm not forced to do it. But it's probably one of the only points where I feel there is no benefit to allow it and only drawbacks.
The mismatch of not having to follow the same structure is not an annoyance, it is a liberation. Much easier to know which rules apply on what, much easier to override them, much easier to move elements around.
If you are doing it correctly, you pretty much never have to update CSS when you change the template, unless you actually want to change the style.

I see your argument for position and flex, but for them we have to use bootstrap classes and not custom style (maybe not just for absolute, but then parent and child are not linked anyway).

Some interesting read in regard to this (and a lot of other) points: https://adamwathan.me/css-utility-classes-and-separation-of-concerns/

content/developer/reference/guidelines.rst
seb-odoo: What is your suggestion for name then, all lower case and only hyphens?
With double hyphens to separate semantically different parts of the name or not?

content/developer/reference/guidelines.rst
```diff
@@ -999,17 +999,88 @@ More precise JS guidelines are detailed in the `github wiki  <https://github.com
...
+
+.. note::
+    The importance of an element in regard to this rule is left to the appreciation of the reviewer. If the structure of the template is hard to follow when reading just the first words of each line, it is probably a sign that some `name` attributes are missing on those lines.
+
+The name on the root element should follow the format `o-ComponentName`. And for children elements `o-ComponentName-partName` where:
```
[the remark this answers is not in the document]
seb-odoo: Here for `name` attribute it is not a CSS class and the goal is not to write CSS on it. So the naming doesn't have to follow any convention, it is what we want it to be, just a name in our code.

I don't care too much what syntax it is as long as each part is easily identifiable at glance, so feel free to suggest something that looks less bad for you.

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
qsm-odoo: Are you actually suggesting to use a random "name" attribute on random elements I am not sure? :thinking: If that is the case, that is not a valid attribute for any HTML element so this should definitely not be done... if I understood well :grimacing:

content/developer/reference/guidelines.rst
qsm-odoo: Are we talking about doing that?
```xml
<div class="hello"
     id="world"
>
    ...
</div>
```
I am not sure to interpret the sentence correctly

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
seb-odoo: I wanted another syntax completely but AL wanted name, so this was a compromise. There are already various `name=` in templates all around Odoo code and I don't think it makes any issue?
Technically it's a valid attribute, but only on form elements.

We can use `data-name` or anything else, but it will be more verbose.
Another suggestion would be `t-name`, or another built-in instruction, in that case we can print in the DOM something different than what we have in the template, so we can make the syntax less verbose in the template (no need for the component part) while still having the expected result in DOM (but AL/FP don't like template not strictly equal DOM).

content/developer/reference/guidelines.rst
seb-odoo: Suggestion in that case would be
```xml
<div
    class="hello"
    id="world"
>
    ...
</div>
```
This is how it looks in a real example, according to the link @ged-odoo sent above: the https://github.com/odoo/odoo/blob/master/addons/web/static/src/webclient/navbar/navbar.xml#L26

I personally think both in this example and in all of that file it's way too verbose in terms of number of lines to apply this syntax.

I would be completely in favor of this syntax but only if the structure of the template was defined separately from its content. But when both are mixed like this and the template is more than a few lines, it becomes impossible to see the big picture of what the interface will contain, which I believe is more important very specific attributes of each element.

Note that as I said in another comment, this problem could actually become a non-problem with proper tooling. Tooling could go both ways too, either helping to expand for those that like it that way, or helping to collapse when the structure matters. Tooling could also be much more advanced, to define a template completely differently (for example with a drag&drop GUI like the editor, just to name an example) and then it would produce the expected result in whatever format we like to review/to have in the codebase, without even having to worry about it.

fpodoo:
I believe a good guidelines should not be too strict; often, common sense and the context are more important than strong guidelines. And we can trust developers to do well.

For example, qweb on the tag makes it obvious that you loop on the <tr>:
```
<tr t-foreach="..." t-as="...">
    <td></td>
</tr>
```

or here (it makes it clear that the "if" is related to an alert dialog):
```
<div t-if="..." class="alert alert-primary">
     Please check this alert.
</div>
```

But in other context, I agree that a separate <t> is more clean:
```
<t t-if="...">
    <div>a mix of different things</div>
    <div>bla</div>
</t>
```

The reason I don't like strict guidelines is that it increases back-and-forth during code review for no real benefit (it does not make the product better) and it will make the review+merge process even more complex. (could be frustrating for the ones who don't like that)

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
ged-odoo: I don't see how it is more declarative to put the t-if on the tag itself. Also, I could argue that there are obvious downsides: it also crowds the list of attributes on that tag, and make it potentially less clear. Multiple directives on the same node have different priorities, which is made obvious if the t-if is separated on a parent t node.  

In any case, in my opinion, I don't think either solution is best for all cases, so I think that it is better to leave that one out of the guideline, otherwise we will lose so much time arguing about it later (see FP comment on that).

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
ged-odoo: a comment is also valid html, and is formatted in a different color, which makes it easier to see. It is also totally standard and anyone outside of odoo will know what it is.  And as @qsm-odoo said, `name` is not a valid html attribute. Also, if your template is difficult to understand, I am not sure how an additional attribute will help that much.

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
ged-odoo: sorry here I thought you were talking about css. In that case, the simplest solution is not to add a `name` attribute at all, so we don't have to answer the question how to name it :)

content/developer/reference/guidelines.rst
ged-odoo: the question is how much value we get out of both representation as we write it. Obviously, they both have the same xml structure, but they convey the information in a different way. I actually think that multiple lines are easier to review and understand. I spend much more time tracking a particular attribute/prop than trying to understand the structure of a template. So in this case, I think this should be removed from the guidelines

content/developer/reference/guidelines.rst
ged-odoo: the tooling exists: you can already do fold all/unfold all or fold at level N in most IDE. And the good thing is that it is an opt-in tooling, noone is forced to install it

content/developer/reference/guidelines.rst
ged-odoo: I am sorry, but I do not agree on the guideline. I do not see the additional value. Also, I do not want to add an extra compilation step for negligible value. I want to reduce friction, rather than increase it

content/developer/reference/guidelines.rst
ged-odoo: same as above obviously :)

content/developer/reference/guidelines.rst
ged-odoo: I am unconvinced. I read the linked post, it is very interesting and I mostly agree with it. But the main point is that we should use utility classes, and I agree with that.

ged-odoo: also, agree on the file per component.

content/developer/reference/guidelines.rst
ged-odoo: just lowercase and only hyphens. no need for double hyphens. Just something like `o-webclient` or `o-form-view` looks nice. If you really want to have a sub part, I guess it could be `o-form-view-header`. On that topic, I think we made a mistake of using the `o_` prefix in the past (it was a renaming of the `oe_` prefix, so a logical evolution). I am in favor of using `o-` from now on instead of `o_`

ged-odoo:
a good guideline is difficult @seb-odoo It is clearly difficult to find common ground. Sorry to make your life harder, but this is the kind of situation where we can easily lose so much time refusing/reviewing PR, frustrating devs, asking for insignificant template changes, or writing/installing/running custom tooling.

In this case, here are the only parts of the spec that I think would bring value to the odoo ecosystem:
- a standard way to name custom css classes: prefix `o-`, and a sequence of lowercase separated by `-` such as `o-form-view`
- a recommendation to put the css specific to a component in a file next to it
- standard indentation fixed to 4 spaces, make sure closing tag is indented exactly like opening tag
- use bootstrap standard utility class as much as possible (unless it is getting ridiculous, but then, common sense should be applied)

Also, this one may be acceptable:
- all sub parts of a component template that need to be targetted by css could have a standard way to be named, with the main component as prefix. So, as an hypothetical example, the search view in the control panel could have `o-control-panel-search-bar` instead of `o-search-bar`

content/developer/reference/guidelines.rst
qsm-odoo: Agreed... most editors would prevent any need for these guidelines. I even have a XML parser in mine.
Screenshot from 2021-07-07 10-35-12


Sorry but I physically cannot put a `>` on a single line, that just look like a buggy misplaced character *for me*. And maybe I don't read enough but I have never seen anyone doing that :shrug: 

> it becomes impossible to see the big picture of what the interface will contain, which I believe is more important than very specific attributes of each element.

I really wonder why you think that. If you want to see the "big picture", just fold level 1 of your XML. Once you have seen the big picture and start unfolding... many convention here will *prevent* you to see the specific attributes correctly. Like I don't mind the one-line attributes (I do it myself most of the time now)... but that just should not be a convention to enforce. Sometimes one-line is clearer, sometimes it's not, that should be left to case-by-case judging :shrug: Typically if I have a long and important conditional attribute, I might want to put it on its dedicated line for the sake of readability and git history.

content/developer/reference/guidelines.rst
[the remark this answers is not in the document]
qsm-odoo: > Technically it's a valid attribute, but only on form elements. [...] and I don't think it makes any issue

So it's not a valid attribute on any HTML element. Although we have some errors to fix since a while, passing W3C validation is still something we want on the website side. Probably less important in the backend but still louche to go against W3C rules for no reason :wink:

content/developer/reference/guidelines.rst
JKE-be: > This is how it looks in a real example, according to the link @ged-odoo sent above: https://github.com/odoo/odoo/blob/master/addons/web/static/src/webclient/navbar/navbar.xml#L26

If you check line 18, you will see yourself that it is not indented of 4 space, and it is probably a good thing ! 
As say by @qsm-odoo it should be left to case-by-case judging. 

```xml
<a id='ok' href='ok'>
```
```xml
<a t-if='a very long condition or another very long condition'
   id='ok'
   href='ok'>
```

```xml
<a t-if='a very long condition or another very long condition'
   id='ok'
   href='ok'
>
```
 
```xml
<input t-if='a very long condition or another very long condition'
    id='ok'
    href='ok'
>
```

```xml
<input t-if='a very long condition or another very long condition'
       id='ok'
       href='ok'
>
```

```xml
<input
    t-if='a very long condition or another very long condition'
    id='ok'
    href='ok'
>
```

```xml
<input
    t-if='a very long condition or another very long condition'
    id='ok' href='ok'
>
```

Imo none are wrong ! And sometimes mix both could make sense:
image


 
For extremely long lines => Imo, when you want and when it make sense. And easy for review, history, and coding...
image

seb-odoo:
I have updated the PR according to feedback:

- removed rules that clearly didn't get consensus
- changed some rules to be suggestions instead of requirements
- adapted some phrases/wording
- added examples
- followed 100 characters limit per line guideline for rst writing

Hopefully this should be better for everyone, let me know.

AntoineVDV:
@seb-odoo Please request a review from `odoo/doc-review` when everybody is ok with the content.

AntoineVDV:
That's not what I meant @C3POdoo.

mart-e:
> That's not what I meant @ C3POdoo.

Being subscribed once is not enough, you have to ping this poor bot so I receive every email in double, why so much hate ?

PR #1067, opened 2021-07-20
[ADD] conf: add sitemap generation

mart-e:
Requires jdillard/sphinx-sitemap@6d5e7b4aeb0a9

Replaces #499

Xavier is working to add parallelism at jdillard/sphinx-sitemap#29

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -111,6 +111,9 @@ extensions = [
 
     # Strange html domain logic used in memento pages
     'html_domain',
+
+    # Sitemap generation
+    'sphinx_sitemap',
 ]
 if odoo_dir_in_path:
     # GitHub links generation
@@ -246,6 +249,11 @@ def setup(app):
     app.add_lexer('xml', XmlLexer)
 
     app.connect('html-page-context', _generate_alternate_urls)
+
+    if app.config.language == "en":
+        app.config.sitemap_url_scheme = "{version}{link}"
+    else:
+        app.config.sitemap_url_scheme = "{version}{lang}{link}"
 
 
 def _generate_alternate_urls(app, pagename, templatename, context, doctree):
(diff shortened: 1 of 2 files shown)


mart-e:
Example of sitemap.xml file generated (manually indented)
https://gist.github.com/mart-e/27a51ff3de495af3a7fd6d2a5edabf8c

xmo-odoo:
So this is going to build one sitemap file per version per language right?

mart-e:
> So this is going to build one sitemap file per version per language right?

Yes, same as for the actual content, we call `sphinx-build` for each version/language combo.

> Is it really useful to use the "localized" URLs then? Why not just format the root URL ourselves and pass that directly to sphinx-sitemap?

Would that be better? Because the code is pretty simple now :)

xmo-odoo:
>> So this is going to build one sitemap file per version per language right?
> 
> Yes, same as for the actual content, we call sphinx-build for each version/language combo.

Right so the "overlord" process managing the generation pipeline will then be generating the "index" sitemap?

Also how do we hook *that* into `odoo.com`'s sitemap? Is it possible to link an index to an index? Or to have multiple "root" sitemaps not linked through indexes?

> > Is it really useful to use the "localized" URLs then? Why not just format the root URL ourselves and pass that directly to sphinx-sitemap?
> 
> Would that be better? Because the code is pretty simple now :)

Dunno, asking. We're already having to conditionally dispatch based on the reference language...

Xavier-Do:
I summon @robodoo

Xavier-Do:
I summon @robodoo

xmo-odoo:
@robodoo ping

Xavier-Do:
@robodoo ping

mart-e:
Closing as not useful. All pages are accessible through the homepage of the doc and are already properly indexed by Google.

PR #1074, opened 2021-08-03
[FIX] core: wrong redirections

Feyensv:
Cf #1054


AntoineVDV (approved):

AntoineVDV:
@robodoo delegate+

Feyensv:
Well, it was even less correct I thought, the right side also had to be changed ^^.

Feyensv:
@robodoo r+

PR #1075, opened 2021-08-03
[FIX] Makefile: fix CSS compilation for multi-version/lang builds

AntoineVDV:
Commit 547d90a allowed `BUILD_DIR` to be defined from the CLI but didn't
update usages of `HTML_BUILD_DIR`. When building the documentation for
multiple versions and/or languages, we should rely on `HTML_BUILD_DIR`
to point to the appropriate directory.


AntoineVDV:
@robodoo r+

PR #1084, opened 2021-08-05
[FIX] makefile: fix overridden css

Xavier-Do:
The previous makefile was compiling css beside the style.scss and
the style.css was copied at the end.

The new makefile compile css directly in the build dir.

Since this file was still present in the source because of the .gitignore,
sphinx was using this old version of style.css, overriding the new compiled
css file at the end.

Removing it from gitignore should helpFix this problem.

This commit also fix some path in makefile.
(not critical but this should fix a potential error in static command)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 
 # Sphinx build files
 _build/
-extensions/odoo_theme/static/style.css
 
 # Dependencies
 odoo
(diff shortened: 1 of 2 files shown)


mart-e (approved):

AntoineVDV (approved):
@robodoo r+

PR #1093, opened 2021-08-06
[FIX] Makefile: avoid confusing // in build dir path

Feyensv:
When locally building the doc (aka make), CURRENT_LANG is not defined,
which results in HTML_BUILD_DIR = _build/html//, resulting in commands
& logs like the following

mkdir -p _build/html//_static
pysassc extensions/odoo_theme/static/style.scss _build/html//_static/style.css

This has no impact on linux builds, but could be unclear/confusing for some non tech
users (can it have any impact on other OS's/distros ?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ ifndef BUILD_DIR
   BUILD_DIR    = _build
 endif
 
+ifndef CURRENT_LANG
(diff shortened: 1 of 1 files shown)


AntoineVDV (approved):
@robodoo r+

PR #1097, opened 2021-08-06
Update difference_warehouse_location.rst

jedgalbraith:
correct much to many


jcs-odoo (approved):
needs to be backported to 13.0

jcs-odoo:
#1380

PR #1098, opened 2021-08-09
Update import.rst

jedgalbraith:
correct "use one of your" to "use one from your"


jcs-odoo:
backported to 13.0 in #1381

PR #1102, opened 2021-08-17
[FIX] app/website: Gengo doesn't work and can't have worked

raf-odoo:
Backport of https://github.com/odoo/documentation/pull/1002 .

In V12 and V13, Gengo can't be used on SaaS and Odoo.sh by default
because they don't install the gengo Python library. If the library
is installed on Odoo.sh by the user, it still won't work, because there's
a typo in `base.gengo.translations` here (the -s is missing):
https://github.com/odoo/odoo/blob/651caf0f5913b1f8c0758664b55644d7d35f00a6/addons/base_gengo/controller/main.py#L15

This code is responsible for handling the results returned by Gengo when
the translation is ready. The typo has been present since at least 2016.
Meaning this can't have worked during that period. Fixing that typo
results in the json.loads following it failing due to an unexpected
number of arguments. That code looks it hasn't ran in a long while.

So I propose to remove mention of Gengo from the documentation in V12
and V13 to avoid any future confusion about it possibly working. Since
the feature can't have worked and support for it in V14 has already been
dropped, it doesn't seem worthwhile to attempt to bugfix it at this stage.


Feyensv:
@fw-bot up to 13.0

Feyensv:
Ok for the diff (and ty for the detailed commit explanation) but do you have any confirmation from a guru/responsible/knowledgeable person on the fact that gengo doesn't and shouldn't work ?

I approved (and merged) the 14.0 PR since I knew the discussion and the removal of gengo in 14+ but for 12-13 I'd like another confirmation before removing the documentation.

I find it really strange to have a module in standard not working, and that it won't be fixed on purpose...
Also, since it doesn't work on SH and Saas, shouldn't we instead specify in the doc that it only works on premise then?

raf-odoo:
@fw-bot up to 13.0

raf-odoo:
@Feyensv I checked with APR, which was the closest I found to a PO willing to take responsibility for Gengo. CPE and PDE didn't consider it part of their domain. 

Regarding it working on-premise: that's what support is currently telling the customers, but given bugs like that typo I mentioned, I don't think it works there either. I tested both locally with gengo installed and on Odoo.sh with the gengo library added to requirements.txt. I confirmed the library loads correctly in that case and that I could authenticate with Gengo with our test credentials. On Odoo.sh I got tracebacks because of the typo when Gengo calls back to us. On local that doesn't reproduce, because Gengo can't reach the server on my laptop, so the problem doesn't trigger. I'll check if one of the devs involved with the earlier dropping is willing to comment.

raf-odoo:
@nle-odoo Would you mind having a look at this PR? I saw that in #2234713 you had encountered a V13 issue where the Gengo implementation couldn't have worked until you fixed the bug you found. I think I've found at least two more bugs in V12 and V13 where it couldn't have ever worked. Could you review my reasoning from the commit message and my previous comment?

raf-odoo:
It seems there are 2 places where we sync with Gengo: the first is the callback when Gengo actively calls us back, mentioned in the PR description and specified to Gengo here when we send a job: https://github.com/odoo/odoo/blob/23e0d2cc17f94c71496f271d0be83e7df402dc83/addons/base_gengo/wizard/base_gengo_translations.py#L230. The second is over here, triggered by a cronjob: https://github.com/odoo/odoo/blob/23e0d2cc17f94c71496f271d0be83e7df402dc83/addons/base_gengo/wizard/base_gengo_translations.py#L123 . I'll test if that alternate code path works.

raf-odoo:
Can confirm the second code path works. So the feature isn't entirely broken. It's just that the translation won't update when the job is done on Gengo's side (because that part is currently broken due to the typo and json loading issue). It will update when the cronjob triggers.

So I'll change the scope of this PR to be about updating the documentation to say that the feature is not available by default on SaaS and Odoo.sh, but that it can be used on-premise or on Odoo.sh with the necessary Python library added to the requirements. I also think that bugfixing the callback is more justified now, so I'll see if I can do a PR for that.

raf-odoo:
I think I'll close this PR for now and will create a new one when it's ready, and then link it here.

raf-odoo:
Bugfix PR: https://github.com/odoo/odoo/pull/75230

PR #1109, opened 2021-08-24
Clarify active vs auto_install docs

Yajo:
1. `active` is just a deprecated field that means exactly the same as `auto_install`
2. `auto_install` can be a list.

See https://github.com/odoo/odoo/blob/6cb3140ea8f62a265b76ae494295f466ca6a00c4/odoo/modules/module.py#L348-L352

### DIFF
--- a/content/developer/reference/addons/module.rst
+++ b/content/developer/reference/addons/module.rst
@@ -90,7 +90,7 @@ Available manifest fields are:
 ``demo`` (``list(str)``)
     List of data files which are only installed or updated in *demonstration
     mode*
-``auto_install`` (``bool``, default: ``False``)
+``auto_install`` (``bool`` or ``list(str)``, default: ``False``)
     If ``True``, this module will automatically be installed if all of its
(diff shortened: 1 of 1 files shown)


joao-p-marques (approved):

joao-p-marques:
Maybe [this commit](https://github.com/odoo/odoo/commit/2cc7f0dd49a6be1920cef232e6c010fbafa68270#diff-367842087e68d6811cf53f2ab3eb21103cd8e793b1648ef3e57f220ee0ccde4eL345) can also help clarify it.

Yajo:
Yes, although that line was actually being removed in the commit and it might seem like it's no longer deprecated, in reality that behavior was preserved.

AntoineVDV (approved):
Hi, @Yajo and thank you for your contribution. Since this PR could also target 13.0 and because the fork belongs to a GitHub organization (I'm unable to push on it), I created a new PR to make some wording change here: https://github.com/odoo/documentation/pull/1427

PR #1117, opened 2021-09-01
[FIX] sale_amazon: remove the installation request of `delivery`

Demesmaeker:
Since delivery is now a dependency of `sale_amazon`, the user don't have to
install it anymore. But they are still asked to get a tracking reference, so
this part stay untouched.

task-2635654


AntoineVDV (approved):
@robodoo r+

AntoineVDV (approved):
@robodoo r+

PR #1118, opened 2021-09-04
Update security.rst

dhavalb-emipro:
[FIX]  access rules to record rules


AntoineVDV (commented):
Hi, @dhavalb-emipro and thank you for your contributions. As you deleted your fork and because there were many other occurrences of "access rule" left, I continued your work in https://github.com/odoo/documentation/pull/1421.

PR #1122, opened 2021-09-07
Views doc

Polymorphe57:


content/developer/reference/addons/views.rst
```diff
@@ -1131,48 +1131,51 @@ The graph view is used to visualize aggregations over a number of records or
...
   selectable measures.
 
-``type``
-  indicates whether the field should be used as a grouping criteria or as an
-  aggregated value within a group. Possible values are:
+``type`` (optional)
+  if set to ``measure``, the field will be used as an aggregated value within a
+  group instead of a grouping criteria. Remark: only works for the last field
+  with that attribute but it is useful for other fields with string attribute
+  (see below).
```
Feyensv, line 1167:
```suggestion
  group instead of a grouping criteria. It only works for the last field
  with that attribute but it is useful for other fields with string attribute
  (see below).
```

Feyensv (commented):
Not much to say, only one little preference of complete sentence instead of "Remark: ...".
Also, please update the commit to target more precisely the scope modified and explain the changes in more details:

```[IMP] developer/views: update doc according to new views guidelines

Brings some modification to the documentation on cohort, graph, and
pivot views according to recent refactoring of views in owl.

Related PRs:
Community: odoo/odoo#73311
Enterprise: odoo/enterprise#19469
```
or something like that :).

PR #1123, opened 2021-09-07
[IMP] developer/views: adapt doc after views refactoring

Polymorphe57:
Brings some modification to the documentation on cohort, graph, and
pivot views according to the recent refactoring of the reporting views
in owl.

Related commits:
Community: odoo/odoo#73311
Enterprise: odoo/enterprise#19469


Feyensv:
@robodoo r+

PR #1127, opened 2021-09-10
Fix "Pull From" documentation -> Inconsistency of terms & operation

rafaelbn:
Source -> Destination

Documentation says:"When the need appears in the source location
Odoo says: "When products are needed in [Destination Location]

imagen


jcs-odoo:
Fixed with #1390 
Thank you for your contribution!

PR #1147, opened 2021-09-16
Update bank_accounts.rst

ray-odoo:
I also would like to suggest we update the very last screenshot with this:

image


ray-odoo:
@jcs-odoo small update - mainly to synch with the new design of the UI and the way we have default Outstanding Accounts now in Settings, so Users would override these now on the Journal.

AntoineVDV:
Please open a new PR with the same changes. You have a [REL] commit in this PR and it's the simplest way to fix it. Also, JCS should soon be able to review the content so I suggest we wait for his input since he was pinged ;)

PR #1155, opened 2021-10-04
Update google_spreadsheets.rst

Auneor:


jcs-odoo:
Translations are handled on [Transifex](https://www.transifex.com/odoo/)

PR #1166, opened 2021-10-06
Update partnership.rst

PaulButau:


odony:
Hello Paul,

This is the master template for the Odoo Partnership contract, we don't modify it for every partner who signs up. You can ask your Odoo account manager to send you the PDF version via Odoo Sign so you can sign it online.
The PDF version is also available here: www.odoo.com/legal

Have a nice day :)

PR #1178, opened 2021-10-14
[IMP] developer: add page on owl component system

ged-odoo:


ged-odoo:
@odoo/doc-review thanks for the review. I think I applied all your suggested changes

AntoineVDV:
@robodoo delegate+

@ged-odoo LGTM but the toctree is conflicting because of the other PR that was just merged. Take the chance to update the `seealso` in both pages ;)

PR #1203, opened 2021-10-21
Update partnership.rst

spidyvijay24:


jcs-odoo:
This modification looks like an unintentional mistake.

PR #1210, opened 2021-10-22
doc js effect

SimonGenin:


content/developer/reference/javascript/services.rst
```diff
@@ -475,3 +481,174 @@ A notification that closes after a second
...
+      message: "Boom! Team record for the past 30 days.",
+    });
+
+.. warning :: 
+    The hook `useEffect` is not related to the effect service.
+
+API
+~~~
+
+.. js:function:: add(options)
```
Feyensv, line 522: Red runbot is caused by this function because function  `add` is specified twice on the page (and in the doc).

`WARNING: duplicate function description of add, other function in developer/reference/javascript/services`

You have to link it to a specific class to differentiate the two methods.

```rst
.. js:class:: EffectService()

  .. js:method: add(params)

```

(Like it's done by Gery in #1245)

Feyensv (commented):
@robodoo delegate+

SimonGenin:
@robodoo r+

PR #1218, opened 2021-10-24
Update minimum python version

frln:
see https://github.com/odoo/odoo/blob/cafc6600749ac75630501912244d79f6be592cb4/odoo/__init__.py#L19


AntoineVDV:
Hi, @frln I took the opportunity to apply the requested changes and update a few more references to Python 3.6. Thanks for you contribution :)

@robodoo r+

PR #1234, opened 2021-10-28
[IMP] payment_acquirers/authorize: explain how to add redirect urls

Demesmaeker:
These explanations were wrongly erased from the documentation between 14.0 and 15.0

task-2679882


Demesmaeker:
Even if the redirects urls are needed in 14.0, they aren't in 15.0. My bad.

PR #1241, opened 2021-10-29
[IMP] developer: add section on cookie service

ged-odoo:
and slightly improve notification information

### DIFF
--- a/content/developer/reference/frontend/services.rst
+++ b/content/developer/reference/frontend/services.rst
@@ -116,12 +116,52 @@ Reference List
 
    * - Technical Name
      - Short Description
+   * - :ref:`cookie <services/cookie>`
+     - read or modify cookies
    * - :ref:`rpc <services/rpc>`
      - send requests to the server
    * - :ref:`title <services/title>`
      - read or modify the window title
    * - :ref:`user <services/user>`
      - provides some information related to the current user
+
+.. _services/cookie:
+
+Cookie service
+--------------
+
+Overview
+~~~~~~~~
+
+- Technical name: `cookie`
+- Dependencies: none
+
+Provides a way to manipulate cookies. For example:
+
+.. code-block:: javascript
+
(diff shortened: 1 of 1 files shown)


aab-odoo (approved):

ged-odoo:
one more for our beloved @odoo/doc-review team...

Feyensv (approved):
@robodoo delegate+

(I launched a rebuild for the red runbot)

content/developer/reference/javascript/services.rst
```diff
@@ -113,13 +113,53 @@ Reference List
...
+.. code-block:: javascript
+
+   cookieService.setCookie("hello", "odoo");
+
+API
+~~~
+
+.. js:data:: current
+
+   Object representing each cookie and its value if any (or empty string)
```
Feyensv, line 147: `WARNING: duplicate data description of current, other data in developer/reference/javascript/services`

duplicate definition of same `js:data` is the cause of the red runbot.

ged-odoo: hmm, that's annoying. Not sure what to do here. I tried the `js:module::` directive (see 
https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-js-module), the documentation says that it does not display a heading, which is what I want, but it does not work like that. It displays a ugly heading, just like classes. any idea?

Feyensv: 1) Either encapsulate it as a "false" class
image

2) Begin specifying methods and attributes with respect to their service:

image

```suggestion
.. js:data:: cookieService.current

   Object representing each cookie and its value if any (or empty string)
```

I don't have any other solution keeping the js domain logic.

ged-odoo: i finallly chose to use the `:noindex:` flag, so it works as expected.

ged-odoo:
@robodoo r+

PR #1243, opened 2021-10-31
[IMP] developer: rename addons/js => backend/frontend

ged-odoo:
The terminology did not really make sense for a lot of these sub pages.
This change should have been done long ago, but was probably delayed
because some people in Odoo call the web client the "backend", which is
really misleading.

This commit also renames the title "Reference Guides" to "Reference",
since this section does not contain any guides anyway.


ged-odoo:
one PR for @odoo/doc-review team. This one is not exactly about javascript only: it changes the structure/organization of the reference documentation. I think that the previous split into `addons` and `javascript` did not make much sense, and caused some weird issues, such as having the assets documentation into the `addons` section.  This PR tries to address theses issues. Hope you like it.

redirects.txt
```diff
@@ -284,4 +284,32 @@ purchase/replenishment/multicompany/setup.rst applications/general/multi_compani
...
+developer/reference/javascript/registries.rst                    developer/reference/frontend/registries.rst
+developer/reference/javascript/services.rst                      developer/reference/frontend/services.rst
+developer/reference/javascript/generic_components.rst            developer/reference/frontend/generic_components.rst
+developer/reference/javascript/hooks.rst                         developer/reference/frontend/hooks.rst
+developer/reference/javascript/javascript_cheatsheet.rst         developer/reference/frontend/javascript_cheatsheet.rst
+developer/reference/javascript/javascript_reference.rst          developer/reference/frontend/javascript_reference.rst
+developer/reference/javascript/mobile.rst                        developer/reference/frontend/mobile.rst
+developer/reference/javascript/qweb.rst                          developer/reference/frontend/qweb.rst
+
+developer/reference/backend/assets.rst                           developer/reference/frontend/assets.rst
```
AntoineVDV, line 315: I guess I'll have to improve the redirects extension sometime soon to allow globs. (but not today 😅)

AntoineVDV (commented):
As discussed, we should make the `reference.rst` page clickable at some point and add content explaining briefly the terminology of Odoo regarding "backend/frontend dev", "backend == web client", etc. But even without that clarification, it already makes more sense to rename.

@robodoo delegate+

ged-odoo:
@robodoo rebase-ff r+

PR #1245, opened 2021-10-31
[IMP] developer: use js domain for registry page

ged-odoo:
This commit update the registry documentation to use the js domain
directives instead of trying to format functions with code blocks.

I am not impressed with the way it looks, but in any cases, it is
certainly better to properly structure the documentation.


ged-odoo:
one pr for @odoo/doc-review team. Certainly better, but I think that in general, the resulting look for method and function documentation could be improved. No real idea how. Maybe some css?

Feyensv (approved):
@robodoo delegate+

ged-odoo:
@robodoo r+

PR #1251, opened 2021-11-01
[FIX] developer: document assets key in manifest, remove old info

ged-odoo:
This commit adds a pretty important missing piece of information: the
assets key in the manifest. At the same time, we remove two keys that
are no longer in use.

### DIFF
--- a/content/developer/reference/backend/module.rst
+++ b/content/developer/reference/backend/module.rst
@@ -118,11 +118,10 @@ Available manifest fields are:
     Whether the module should be considered as a fully-fledged application
     (``True``) or is just a technical module (``False``) that provides some
     extra functionality to an existing application module.
-``css`` (``list(str)``)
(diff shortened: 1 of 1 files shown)


ged-odoo:
I never heard of these two keys: css and images. but it is used in the HR module manifest. Are they really useless?

aab-odoo (approved):

ged-odoo:
a pr ready for @odoo/doc-review

Feyensv:
> I never heard of these two keys: css and images. but it is used in the HR module manifest. Are they really useless?

We can still discover some (old) things in the doc sometimes ^^.
Anyway, dunno if they work, but since they are unknown and there are better ways to specify the assets, let's avoid promoting their use.

@robodoo r+

AntoineVDV (approved):
@robodoo delegate+

PR #1257, opened 2021-11-02
[FIX] conf: allow to build on python 3.6

mart-e:
Since odoo/odoo@794677fb6a3391379 Odoo requires python 3.7 min.
However, it still runs fine in 3.6 and this is the python version of
the build server.
Fake the version to be able to build the doc on the nightly server.

To be removed once the nightly server is migrated.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -74,6 +74,10 @@ if not odoo_dir.is_dir():
     )
 else:
     sys.path.insert(0, str(odoo_dir.absolute()))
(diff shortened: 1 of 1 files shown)


AntoineVDV (commented):
> To be removed once the nightly server is migrated.

I'll be back with this soon...

@robodoo delegate+

mart-e:
@robodoo r+
(if you are here)

PR #1258, opened 2021-11-02
[FIX] testing: Explain "new" tagging syntax

raf-odoo:
Syntax was added to explicitly specify a module, class or function with
`--test-tags`: https://github.com/odoo/odoo/commit/b729a11a121e53857063be00308904ed4d8dd2a0

The documentation was not updated to reflect this, leading to confusion
as people tried to reconcile the documentation with the actual behavior,
without necessarily having seen `odoo-bin --help` which is explicit
about it.


raf-odoo:
As an example of this confusion, I thought `-i` and `-u` were still required to trigger tests properly and wasn't aware that a `/` needed to be in front of the module name or it wouldn't execute properly.

raf-odoo:
@Xavier-Do Since you wrote the original commit, could you review if the changes I'm making in this PR are correct?

Xavier-Do (approved):

Feyensv:
@robodoo r+

PR #1269, opened 2021-11-03
[FIX] Makefile: call pysassc from python module

d-fence:
pysassc utility is called in the Makefile. On some system (e.g.: the latest
Dockerfile of the runbot) this utility cannot be found in the system's
paths.

As the pysassc utility is a wrapper for the python module, we can simply
call the python module by it's entrypoint.

Also, this commit fixes the requirements as the pysass package was
installed to incidentally install libsass.


AntoineVDV (approved):
@d-fence Doesn't it need to be done from 13.0 rather than 15.0?

AntoineVDV:
@robodoo delegate+

d-fence:
robodoo r+

PR #1273, opened 2021-11-03
[FIX] requirements.txt: Let sphinx manage docutils version

AntoineVDV:
As of sphinx 3.5.4, docutils >0.17 is no longer supported. To avoid
forcing the install of an unsupported version of docutils, we let
sphinx install the appropriate version through its dependencies.


AntoineVDV:
@robodoo r+

PR #1291, opened 2021-11-03
15.0 frontend refs ged

ged-odoo:


ged-odoo:
as requested, here is a pr to normalize refs in the js doc. @odoo/doc-review

AntoineVDV (commented):
@robodoo delegate+

Runbot is red. And for some reason I can't browse the doc when it is :man_shrugging: 
Doesn't matter for now since there is no content change here.

ged-odoo:
@robodoo r+

PR #1296, opened 2021-11-05
[IMP] developer: add missing links in js doc

ged-odoo:
and slightly improve effect registry page

### DIFF
--- a/content/developer/reference/frontend/framework_overview.rst
+++ b/content/developer/reference/frontend/framework_overview.rst
@@ -110,7 +110,7 @@ the shared ``env`` object:
    * - `bus`
      - :ref:`main bus <frontend/framework/bus>`, used to coordinate some generic events
    * - `services`
-     - all deployed :ref:`services <javascript/services>` (should usually be accessed
+     - all deployed :ref:`services <frontend/services>` (should usually be accessed
        with the `useService` hook)
    * - `debug`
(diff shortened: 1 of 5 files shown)


ged-odoo:
a small PR for the @odoo/doc-review team. Mostly about adding a few missing links.

content/developer/reference/frontend/framework_overview.rst
```diff
@@ -195,7 +195,7 @@ every 5 second:
 Components and Hooks
 --------------------
 
-Components and hooks are ideas coming from the
+:ref:`Components <frontend/components>` and :ref:`hooks <frontend/hooks>` are ideas coming from the
```
Feyensv, line 198: FYI, there exists a `:doc:` command to reference a document instead of a hyperlink reference.

```:doc:`path_to_file` ```  -> ```:doc:`hooks` ```

I'm not saying you should use this instead of hyperlink references, because hyperlink refs are not broken by file moves, and are easier to use when referencing documents far from current one but it's good to know ;).

Feyensv (approved):
@robodoo delegate+

content/developer/reference/frontend/framework_overview.rst
ged-odoo: good to know indeed. I actually kind of knew it, but didn't even think of it.  In this case, I think i'll keep the refs.

ged-odoo:
@robodoo r+

PR #1327, opened 2021-11-19
[IMP] developer: small rework of assets page (frontend)

ged-odoo:
This commit applies some remarks by Antoine.


AntoineVDV (approved):
Perfect :ok_hand: 

I was also bugged by the two sections about loading. I had the feeling that they could be organized differently but I wasn't sure if they applied to assets in general (then they should be H2) or if any of them was specific to bundles (then they should be an H3 of "Bundles"). I see that you sorted that out.

@robodoo r+

PR #1329, opened 2021-11-22
[ADD] odoo_sh: Mention Python dependency behavior in FAQ

raf-odoo:
Odoo.sh doesn't use the requirements.txt included in Odoo but Python
packages available in the Linux distribution instead. This is
non-obvious and can be confusing to users, especially in the cases where
the version available in the distribution and requirements.txt are
different.

Adding this section is intended to clarify that, and to provide a
reference for developers of custom modules on what to expect.

opw-2688205

### DIFF
--- a/content/administration/odoo_sh/advanced/frequent_technical_questions.rst
+++ b/content/administration/odoo_sh/advanced/frequent_technical_questions.rst
@@ -30,3 +30,26 @@ We advise that:
   `idempotent <https://stackoverflow.com/a/1077421/3332416>`_: they must not
   cause side-effects if they are started more often than expected.
 
+Why are the installed Python libraries different from Odoo's own requirements.txt?
+----------------------------------------------------------------------------------
+
+On Odoo.sh, Odoo uses the Python libraries available in the Linux distribution
+to fulfill its dependencies. Odoo is compatible with the packages in Debian
+stable and Ubuntu LTS releases. The `requirements.txt <https://github.com/odoo/odoo/blob/master/requirements.txt>`_
+included with Odoo is intended for use on different distributions where the
+same library versions might not be available. As a result, Odoo.sh does not use
+Odoo's own requirements.txt file and there might be occasional differences
+between them.
+
+For example, reportlab 3.5.34 is the version available on Ubuntu 20.04 as a
+distribution package. That distribution uses Python 3.8, in which case
+requirements.txt in `Odoo 14.0 <https://github.com/odoo/odoo/blob/2097c75dc38b310c4d9ddecd79e2218448a51a4c/requirements.txt#L46>`_
+specifies reportlab 3.5.55. Normally speaking the pin in requirements.txt
+should reflect that and be 3.5.34, but it turns out that this version only
+works if Ubuntu's patches are applied to it. Ubuntu LTS releases have
+backporting of bug fixes, which means that the behavior of a Python package in
+the distribution might not be the same as that same version when installed with
+pip. This is the case with reportlab, which is why the pin is different.
+
+If a custom module requires a more recent version of a package than is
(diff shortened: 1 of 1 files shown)


raf-odoo:
@jev-odoo @xmo-odoo Could you review this PR to confirm the technical aspects of what I'm saying are correct? After that I'll ask someone from the documentation team to review.

jev-odoo:
@raf-odoo actually, it is not only for Odoo.SH, but it is true for Saas server as well!

amigrave:
@raf-odoo I have some small comments about this but I'm currently busy with some server checking for the black friday, I'll review this beginning of next week

raf-odoo:
@amigrave You're right about the wording, it assumes requirements.txt is the norm. For Python projects in general people are going to assume that requirements.txt is leading and the SaaS / PaaS behavior relying on distro packages is non-standard from that perspective. I understand that from Odoo's perspective it's the other way around. The goal of this section is to clarify that, so that's why I start from those assumptions.

I'll try to tweak the wording a bit to make clear that the official dependencies for Odoo in general are the distro versions and that Odoo.sh is implementing that.

amigrave:
> I'll try to tweak the wording a bit to make clear that the official dependencies for Odoo in general are the distro versions and that Odoo.sh is implementing that.

Let me have a word with AL first, I'd like to check if anything is going to be clarified more officially, in such case the information should probably be in the administration/install section and Odoo.sh referring it in the FAQ.

amigrave:
@raf-odoo  AL confirmed that the official dependencies are the latest long term supported debia/ubuntu packages (using the lowest versions depending on their cycle). @d-fence will forge this in the documentation and will mention the documentation link in the `requirements.txt` itself as a comment. Once done we will evaluate if it's still useful to mention this a second time in the Odoo.sh documentation.

raf-odoo:
Ok, fine by me :-)

content/administration/odoo_sh/advanced/frequent_technical_questions.rst
```diff
@@ -30,3 +30,26 @@ We advise that:
...
+requirements.txt in `Odoo 14.0 <https://github.com/odoo/odoo/blob/2097c75dc38b310c4d9ddecd79e2218448a51a4c/requirements.txt#L46>`_
+specifies reportlab 3.5.55. Normally speaking the pin in requirements.txt
+should reflect that and be 3.5.34, but it turns out that this version only
+works if Ubuntu's patches are applied to it. Ubuntu LTS releases have
+backporting of bug fixes, which means that the behavior of a Python package in
+the distribution might not be the same as that same version when installed with
+pip. This is the case with reportlab, which is why the pin is different.
+
+If a custom module requires a more recent version of a package than is
+installed on Odoo.sh, you can specify it in your own requirements.txt file.
```
jcs-odoo, line 55: - I guess this could be written as a tip
- new phrasing suggestion
- "file" directive
```suggestion
.. tip::
   If a custom module requires a package version more recent than the one installed on Odoo.sh, you
   can specify it in your :file:`requirements.txt` file.
```

jcs-odoo (commented):
Hello, here are already a few comments even though I'm aware you want to rewrite this content.
I haven't reviewed the phrasing of the example yet, but I won't have the time to do it today.

Cheers,

raf-odoo:
@jcs-odoo Thanks for the feedback, I'll hold off on responding to it for the moment, since it's possible this PR might end up getting deleted.

amigrave:
Indeed, https://github.com/odoo/documentation/pull/1364 will keep the details about installation procedure and as soon as it is published I will link the Odoo.sh FAQ to the debian/ubuntu section as the official method for Odoo.sh instances

<https://github.com/odoo/documentation/pull/1364#issuecomment-993345784>

PR #1333, opened 2021-11-24
[FIX] conf.py: check odoo_dir is a real Odoo sources dir

remybar:
In `conf.py`, we try to find a Odoo sources folder among `odoo` and `../odoo` directories without really checking that they really are
Odoo sources folders, leading to doc generation error if it's not the case.

So, to improve that, this commit checks that the selected Odoo sources folder contains `odoo-bin`.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -58,34 +58,40 @@ extension_dir = Path('extensions')
 sys.path.insert(0, str(extension_dir.absolute()))
 
 # Search for the directory of odoo sources to know whether autodoc should be used on the dev doc
-odoo_dir = Path('odoo')
+odoo_sources_candidate_dirs = (Path('odoo'), Path('../odoo'))
+odoo_sources_dirs = [
+    d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()
+]
 odoo_dir_in_path = False
-if not odoo_dir.is_dir():
-    parent_odoo_dir = Path('../odoo')
-    if parent_odoo_dir.is_dir():
-        _logger.info('Using parent dir to find odoo sources')
-        odoo_dir = parent_odoo_dir
-if not odoo_dir.is_dir():
+
+if not odoo_sources_dirs:
     _logger.warning(
-        f"Could not find Odoo sources directory at {odoo_dir.absolute()}.\n"
-        f"The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
-        f"In order to fully build the 'Developer' documentation, clone the repository with "
-        f"`git clone https://github.com/odoo/odoo` or create a symbolink link."
+        "Could not find Odoo sources directory in neither of the following folders:\n"
+        "%(dir_list)s\n"
+        "The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
+        "In order to fully build the 'Developer' documentation, clone the repository with "
+        "`git clone https://github.com/odoo/odoo` or create a symbolic link.",
+        {'dir_list': '\n'.join([f'\t- {d.resolve()}' for d in odoo_sources_candidate_dirs])},
     )
 else:
-    sys.path.insert(0, str(odoo_dir.absolute()))
+    odoo_dir = odoo_sources_dirs[0].resolve()
+    sys.path.insert(0, str(odoo_dir))
     from odoo import release as odoo_release  # Don't collide with Sphinx's 'release' config option
     odoo_version = odoo_release.version if 'alpha' not in odoo_release.version else 'master'
     if release != odoo_version:
         _logger.warning(
-            f"Found Odoo sources directory but with version '{odoo_version}' incompatible with "
-            f"documentation version '{version}'.\n"
-            f"The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
-            f"In order to fully build the 'Developer' documentation, checkout the matching branch"
-            f" with `cd odoo && git checkout {version}`."
+            "Found Odoo sources in %(directory)s but with version '%(odoo_version)s' incompatible "
+            "with documentation version '%(doc_version)s'.\n"
+            "The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
+            "In order to fully build the 'Developer' documentation, checkout the matching branch"
+            " with `cd odoo && git checkout %(doc_version)s`.",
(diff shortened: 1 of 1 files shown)


conf.py
```diff
@@ -51,29 +51,34 @@
...
+# list of folders that may contain Odoo sources
+odoo_sources_candidate_dirs = (Path('odoo'), Path('../odoo'))
+
 #=== Extensions configuration ===#
 
 # Add extensions directory to PYTHONPATH
 extension_dir = Path('extensions')
 sys.path.insert(0, str(extension_dir.absolute()))
 
 # Search for the directory of odoo sources to know whether autodoc should be used on the dev doc
```
AntoineVDV, line 63: `odoo_sources_candidate_dirs` is only a local variable and not a Sphinx configuration var.
```suggestion
#=== Extensions configuration ===#

# Add extensions directory to PYTHONPATH
extension_dir = Path('extensions')
sys.path.insert(0, str(extension_dir.absolute()))

# Search for the directory of odoo sources to know whether autodoc should be used on the dev doc
odoo_sources_candidate_dirs = (Path('odoo'), Path('../odoo'))
```

conf.py
```diff
@@ -51,29 +51,34 @@
...
-odoo_dir = Path('odoo')
+odoo_sources_dirs = [d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()]
 odoo_dir_in_path = False
-if not odoo_dir.is_dir():
-    parent_odoo_dir = Path('../odoo')
-    if parent_odoo_dir.is_dir():
-        _logger.info('Using parent dir to find odoo sources')
-        odoo_dir = parent_odoo_dir
-if not odoo_dir.is_dir():
+odoo_dir = False
```
AntoineVDV, line 66:
```suggestion
```
now useless

conf.py
```diff
@@ -51,29 +51,34 @@
...
+
+if not odoo_sources_dirs:
     _logger.warning(
-        f"Could not find Odoo sources directory at {odoo_dir.absolute()}.\n"
-        f"The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
-        f"In order to fully build the 'Developer' documentation, clone the repository with "
-        f"`git clone https://github.com/odoo/odoo` or create a symbolink link."
```
AntoineVDV, line 73: "symbolink" was such a cool 'mot-valise' :D

conf.py
```diff
@@ -51,29 +51,34 @@
...
+        "Could not find Odoo sources directory in the following folders:\n"
+        "%s\n"
+        "The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
+        "In order to fully build the 'Developer' documentation, clone the repository with "
+        "`git clone https://github.com/odoo/odoo` or create a symbolic link.",
+        "\n".join([f"    - {d.resolve()}" for d in odoo_sources_candidate_dirs])
```
AntoineVDV, line 75:
```suggestion
        "\n".join([f"\t- {d.resolve()}" for d in odoo_sources_candidate_dirs])
```

AntoineVDV (changes requested):
I'm okay with the change but, since we keep patching the patch that allows building the doc on any setup, I also think it's time to introduce a config option to specify the odoo sources paths. It would look like this in the CLI: `make ODOO_SOURCE_PATH=../odoo`. If not passed, the default should be `odoo` to avoid breaking current setups. I'll create a task to do it later.

remybar:
> I'm okay with the change but, since we keep patching the patch that allows building the doc on any setup, I also think it's time to introduce a config option to specify the odoo sources paths. It would look like this in the CLI: `make ODOO_SOURCE_PATH=../odoo`. If not passed, the default should be `odoo` to avoid breaking current setups. I'll create a task to do it later.

I tried to do that but, as far as I know, in the `Makefile` we can just override some variables set by `conf.py` (using the `-D` option of `sphinx-build`). I didn't find any possibility to provide this `ODOO_SOURCE_PATH` argument to `sphinx-build` and being able to access to it inside `conf.py`.

Thanks for the review, I will update the PR ;-)

AntoineVDV:
> I tried to do that but, as far as I know, in the `Makefile` we can just override some variables set by `conf.py` (using the `-D` option of `sphinx-build`). I didn't find any possibility to provide this `ODOO_SOURCE_PATH` argument to `sphinx-build` and being able to access to it inside `conf.py`.

You need to declare it in `def setup(app)` like this: `app.add_config_value('odoo_source_path', None, 'env')`. Then, you can pass `-D odoo_source_path=$(ODOO_SOURCE_PATH)` to `sphinx-build` in the Makefile and read the config value with `app.config.odoo_source_path` in conf.py.

conf.py
remybar: done :+1:

conf.py
remybar: done :+1:

conf.py
remybar: done :+1:

remybar:
> > I tried to do that but, as far as I know, in the `Makefile` we can just override some variables set by `conf.py` (using the `-D` option of `sphinx-build`). I didn't find any possibility to provide this `ODOO_SOURCE_PATH` argument to `sphinx-build` and being able to access to it inside `conf.py`.
> 
> You need to declare it in `def setup(app)` like this: `app.add_config_value('odoo_source_path', None, 'env')`. Then, you can pass `-D odoo_source_path=$(ODOO_SOURCE_PATH)` to `sphinx-build` in the Makefile and read the config value with `app.config.odoo_source_path` in conf.py.

So, after some tests, unfortunately, it's not so easy to dynamically configure extensions in `conf.py` using a `ODOO_SOURCE_PATH` provided through `-D` option to `sphinx-build`, even using the `config-inited` signal.

But, in fact, we have 2 use cases:
1) users without Odoo sources: if not sources are found in `odoo` and `../odoo` a warning will be displayed and the doc will be correctly generated => OK.
2) users with Odoo sources: if sources are found in `odoo` or `../odoo`, they will be taken into account during doc generation. If not found, same than 1). If the user uses a "non standard" location for his Odoo sources, he can create a `odoo` symlink to refer to his source location.
So, at the end, no specific need to handle a custom odoo source path through a Makefile parameter.

AntoineVDV:
LGTM!

FYI: I took the liberty to improve the logging messages a bit, now that it's clear that we support alternative odoo sources paths.

@robodoo r+

AntoineVDV (approved):

PR #1334, opened 2021-11-24
[ADD] extensions: new autofield directive, [REF] i18n/localization.rst

william-andre:


content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+The solution is the usage of tags, one for each report line, to filter accounts like you want.
+
+Put the tags in `data/account_account_tag_data.xml` file:
+
+Then apply tags on some accounts (field `tag_ids`).
+
+.. example::
+  `addons/l10n_lt/data/account.account.template.csv <https://github.com/odoo/odoo/blob/15.0/addons/l10n_lt/data/account.account.template.csv>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_lt/data/account.account.template.csv
```
qdp-odoo, line 155: would probably present better with a `csv-table + file` directive. Can you use `start at`/`end-at` as well in that?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+   - :ref:`Account References <reference/account_account>`
+   - :doc:`/applications/finance/accounting/payables`
+   - :doc:`/applications/finance/accounting/receivables`
+
+Obviously, *Chart of Accounts* cannot exist without *Accounts*. You need to specify them in `data/account.account.template.csv`.
+
+.. example::
+  `addons/l10n_ch/data/account.account.template.csv <https://github.com/odoo/odoo/blob/15.0/addons/l10n_ch/data/account.account.template.csv>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/account.account.template.csv
```
qdp-odoo, line 99: would probably present better with a csv-table + file directive. Can you use start at/end-at as well in that?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/account_tax_group_data.xml
+    :language: xml
+    :start-after: <data
+    :end-before: </data>
+
+.. example::
+  `addons/l10n_uk/data/account.tax.group.csv <https://github.com/odoo/odoo/blob/15.0/addons/l10n_uk/data/account.tax.group.csv>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_uk/data/account.tax.group.csv
+    :language: csv
```
qdp-odoo, line 219: idem

qdp-odoo (approved):
just a minor cosmetic proposal/question. The rest is fine by me and can be improved latter on.

qdp-odoo:
@odoo/doc-review let us know your opinion on the changes proposed in https://github.com/odoo/documentation/pull/1334/commits/8753a51d61f12fecb6a027565cfda310cc90e41e

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+Obviously, *Chart of Accounts* cannot exist without *Accounts*. You need to specify them in `data/account.account.template.csv`.
+
+.. example::
+  `addons/l10n_ch/data/account.account.template.csv <https://github.com/odoo/odoo/blob/15.0/addons/l10n_ch/data/account.account.template.csv>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/account.account.template.csv
+    :language: csv
+    :end-at: ch_coa_1171
+
+XML is prefered but you may use xml format instead.
```
Levizar, line 103: Did you mean CSV is better ?

content/developer/reference/standard_modules/account/account_chart_template.rst
```diff
@@ -0,0 +1,62 @@
...
+
+---------------------------------
+Multi Language (`l10n_multilang`)
+---------------------------------
+
+* Multi language support for Chart of Accounts, Taxes, Tax Codes, Journals,
+  Accounting Templates, Analytic Chart of Accounts and Analytic Journals.
+* Setup wizard changes.
+    - Copy translations for COA, Tax, Tax Code and Fiscal Position from
+      templates to target objects.
```
oco-odoo, line 57: Second * is a bit gibberish.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/l10n_ch_chart_data.xml
+    :language: xml
+    :start-at: l10nch_chart_template
+    :end-at: </record>
+
+
+.. note::
+
+  Recommended **xmlid** for the record is `chart_template`.
```
oco-odoo, line 79: Funny that the example above doesn't respect that :p Should we even bother recommanding anything here ? It actually makes no difference from a technical point of view.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+Chart of Accounts
+=================
+
+Accounts
+--------
+
+.. seealso::
+   - :ref:`Account References <reference/account_account>`
+   - :doc:`/applications/finance/accounting/payables`
+   - :doc:`/applications/finance/accounting/receivables`
```
oco-odoo, line 92: Why linking the doc about payable and receivable accounts here? To me, they have nothing to do with the creation of an l10n module/CoA.

content/developer/howtos/localization.rst
oco-odoo, line 103: csv 
```suggestion
csv is preferred but you may use xml format instead.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+    :start-at: l10nch_chart_template
+    :end-at: </record>
+
+Account tags
+------------
+
+.. seealso::
+   :ref:`Account Tag References <reference/account_account_tag>`
+
+Tags is a way to sort accounts.
```
oco-odoo, line 144:
```suggestion
Tags are a way to sort accounts.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+For example, imagine you want to create a financial report having multiple lines but you have no way to find a rule to dispatch the accounts according their ``code`` or ``name``.
+The solution is the usage of tags, one for each report line, to filter accounts like you want.
+
+Put the tags in `data/account_account_tag_data.xml` file:
+
+Then apply tags on some accounts (field `tag_ids`).
```
oco-odoo, line 150:
```suggestion
Put the tags in `data/account_account_tag_data.xml` file, then apply the tags to some accounts (field `tag_ids`).
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+    :end-at: </record>
+
+
+Account groups
+--------------
+
+.. seealso::
+   :ref:`Account Group References <reference/account_group>`
+
+Account groups allows to describe structure of chart of accounts.
```
oco-odoo, line 174:
```suggestion
Account groups allow describing the structure of a chart of accounts.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+     :header-rows: 1
+
+Taxes
+-----
+
+.. seealso::
+   - :ref:`Tax References <reference/account_tax>`
+   - :doc:`/applications/finance/accounting/taxation/taxes/taxes`
+
+To add taxes you first need to specify tax groups. It has just two fields: name and country. Create file `data/account_tax_group_data.xml` and list the groups:
```
oco-odoo, line 191:
```suggestion
To add taxes you first need to specify tax groups. This model only has two required fields: name and country. Create file `data/account_tax_group_data.xml` and list the groups:
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+    <odoo>
+        <data noupdate="1">
+            <record id="tax_group_tva_0" model="account.tax.group">
+                <field name="name">TVA 0%</field>
+                <field name="country_id" ref="base.ch"/>
+            </record>
+
+            ...
+        </data>
+    </odoo>
```
oco-odoo, line 205: I don't think you need this with the example below.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+.. example::
+  `addons/l10n_ae/data/account_tax_template_data.xml <https://github.com/odoo/odoo/blob/15.0/addons/l10n_ae/data/account_tax_template_data.xml>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ae/data/account_tax_template_data.xml
+    :language: xml
+    :start-at: uae_sale_tax_5_dubai
+    :end-at: </record>
+
+If some accounts should use default taxes, you can set them up in `data/account_account_template_post_data.xml`
```
oco-odoo, line 233: Maybe link an example ?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+Tax Report
+----------
+
+.. seealso::
+   - :ref:`Tax Report Line References <reference/account_tax_report_line>`
+   - :doc:`/applications/finance/accounting/reporting/declarations/tax_returns`
```
oco-odoo, line 240: On financial reports, we explicitly state they are enterprise; we should do the same here. Only the model and the definition of its instances are community. Maybe just explaining it with 1-2 sentences here would be nice as well.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+In previous section you noticed fields `invoice_repartition_line_ids` / `refund_repartition_line_ids` and probably understand nothing about it. Good news: you are not alone on this incomprehension. Bad news: you have to figure it out a bit. The topic is complicated indeed:
```
oco-odoo, line 242:
```suggestion
In the previous section you noticed fields `invoice_repartition_line_ids` / `refund_repartition_line_ids` and probably understood nothing about them. Good news: you are not alone on this incomprehension. Bad news: you have to figure it out a bit. The topic is complicated indeed:
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+.. graphviz:: images/tax_report.dot
+
+
+But lucky we have a presentation explaining the tax reports in details:
```
oco-odoo, line 247:
```suggestion
But fortunately we have a presentation explaining the tax reports in details:
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+.. youtube:: PuXE_NyFRTM
+    :align: right
+    :width: 700
+    :height: 394
```
oco-odoo, line 252: As I said on the other PR, I'm not sure it's a good idea to link it here. This was done for v13, and the model has changed a little since then. Moreover, it essentially compares the financial reports with the new engine, and I'm not sure it'll help newcomers in any way.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+So, once you have properly configured taxes, you just need to add `data/account_tax_report_data.xml` file with a single record for `account.tax.report` in beginning:
```
oco-odoo, line 254:
```suggestion
So, once you have properly configured taxes, you just need to add `data/account_tax_report_data.xml` file with a record for your `account.tax.report` at the beginning:
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+    <odoo>
+        <record id="tax_report" model="account.tax.report">
+            <field name="name">Tax Report</field>
+            <field name="country_id" ref="base.XX"/>
+        </record>
+
+        ...
+    </odoo>
+
+... following by a list of records for `account.tax.report.line`.
```
oco-odoo, line 268:
```suggestion
... followed by the declaration of its lines, as `account.tax.report.line` records.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+
+Fiscal positions
+----------------
+
+.. seealso::
+   - :ref:`Fiscal Position References <reference/account_fiscal_position>`
+   - :doc:`/applications/finance/accounting/taxation/taxes/fiscal_positions`
+
+Specify fiscal positions in `data/account_fiscal_position_template_data.xml` file as following:
```
oco-odoo, line 287:
```suggestion
Specify fiscal positions in `data/account_fiscal_position_template_data.xml` file.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+You may have as many fiscal position as you need.
```
oco-odoo, line 289: Kind of obvious. I wouldn't say it.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_es/data/account_fiscal_position_template_data.xml
+    :language: xml
+    :start-at: fp_nacional
+    :end-before: fp_intra
+
+Final Steps
+===========
+
+The last step on installing Chart of Accounts is applying it on current company. File `data/account_chart_template_try_loading.xml` is responsible for that.
```
oco-odoo, line 302:
```suggestion
The last step when installing a localization module is to try applying its chart of accounts to the current company (if it does not already have one). 
File `data/account_chart_template_try_loading.xml` is responsible for that.
```

content/developer/howtos/images/tax_report.dot
```diff
@@ -0,0 +1,68 @@
...
+    account_tax_report_line;
+    account_tax_repartition_line_template;
+  }
+  account_tax -> account_tax_template [label="Creates when installing CoA", dir=back];
+  account_tax_repartition_line -> account_tax_repartition_line_template [label="Creates when installing CoA", dir=back];
+  account_tax_repartition_line_template -> account_account_tag [label="tag_ids"];
+  account_tax_report_line -> account_account_tag [label="Creates (+ and -)"];
+  account_tax_report_line -> res_country [label="0..1"];
+  account_account_tag -> res_country [label="0..1"];
+}
```
oco-odoo, line 68: It's missing account.tax.report.

We might also want to add account.tax.carryover.line to the drawing, if we really want something complete (not sure it's not too much here).

content/developer/reference/standard_modules/account/account_chart_template.rst
```diff
@@ -0,0 +1,62 @@
...
+
+==============
+Chart Template
+==============
+
+---------------------
+Invoicing (`account`)
+---------------------
+
+This is the main module for Invoicing.
```
oco-odoo, line 11: ?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,370 @@
...
+Tax Report
+----------
+
+.. seealso::
+   - :ref:`Tax Report Line References <reference/account_tax_report_line>`
+   - :doc:`/applications/finance/accounting/reporting/declarations/tax_returns`
+
+In previous section you noticed fields `invoice_repartition_line_ids` / `refund_repartition_line_ids` and probably understand nothing about it. Good news: you are not alone on this incomprehension. Bad news: you have to figure it out a bit. The topic is complicated indeed:
+
+.. graphviz:: images/tax_report.dot
```
oco-odoo, line 244: It's quite small on my screen and there doesn't seem to be a way of maximizing it ... can we do something ?

oco-odoo (changes requested):
Twice's the charm ! :partying_face:

content/developer/howtos/localization.rst
william-andre: They help devs to understand the account types when they have to chose the right ones

content/developer/howtos/localization.rst
william-andre: It is annoying to have `l10n_xx.l10nxx_chart_template`. (quite repetitive) Would be easier if it was uniform.

content/developer/howtos/localization.rst
william-andre: It is indeed explaining the difference (we don't care about that part), but also explaining how the new model works (we care about this)

content/developer/howtos/localization.rst
william-andre: Not that easy https://docutils.sourceforge.io/docs/ref/rst/directives.html#csv-table

I tried with a `.. include` and `.. csv-table` but without success

content/developer/reference/standard_modules/account/account_chart_template.rst
william-andre: ?

content/developer/howtos/localization.rst
william-andre: Is it better now?

redirects.txt
```diff
@@ -315,4 +315,7 @@ developer/reference/javascript/qweb.rst                          developer/refer
 developer/reference/backend/assets.rst                           developer/reference/frontend/assets.rst
 
 developer/reference/frontend/owl_component_system.rst            developer/reference/frontend/owl_components.rst
-developer/reference/frontend/generic_components.rst              developer/reference/frontend/owl_components.rst
\ No newline at end of file
+developer/reference/frontend/generic_components.rst              developer/reference/frontend/owl_components.rst
+
+developer/misc/i18n/localization.rst                    developer/howto/localization.rst
+developer/misc/i18n/translations.rst                    developer/howto/translations.rst
```
Feyensv, line 321: the move of the translations file does not really enter the scope of this PR (and should at least be in a separate commit)

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+Obviously, *Chart of Accounts* cannot exist without *Accounts*. You need to specify them in `data/account.account.template.csv`.
+
+.. example::
+  `addons/l10n_ch/data/account.account.template.csv <https://github.com/odoo/odoo/blob/15.0/addons/l10n_ch/data/account.account.template.csv>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/account.account.template.csv
+    :language: csv
+    :end-at: ch_coa_1171
+
+CSV is prefered but you may use XML format instead.
```
Feyensv, line 130: a little note on the perf diff?

Feyensv (changes requested):
I love the idea of an autofield documenter, but the rest of the "tech" changes do not bring much and break current compatibility & maintainability.

Somebody fetching only the documentation repository must be able to build the documentation (which is not the case with your changes).

Also, in all honesty, the current doc on "how to build a localization" is not a howto, but more a kind of "raw checklist" in its current state.
Unless you are already knowledgeable in the inner details of the accounting modules, I have the feeling that you are not really receiving much more help by reading the doc than by reading a localization module directly.

Did you ask newbies in your team to give their POV on this doc?
As developers, we easily believe everything is easy to understand in our own scope and to oversimplify the our explanations.

"Good news: you are not alone on this incomprehension. Bad news: you have to figure it out a bit." ---> What's the use of a documentation if it says "We leave you in your own mess" ?

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        elif field.type == 'boolean':
+            annotation[attrname] = bool
+        elif field.type in ('float', 'monetary'):
+            annotation[attrname] = float
+        elif field.type == 'integer':
+            annotation[attrname] = int
+        elif field.type == 'date':
+            annotation[attrname] = datetime.date
+        elif field.type == 'datetime':
+            annotation[attrname] = datetime.datetime
```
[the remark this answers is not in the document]
william-andre: What? No? That would return the `fields` class, not the type that we are expecting to input for that field.
When you do 
```python
sell.env['res.users'].write({'name': something})
```
You don't expect `something` to be an instance of `fields.Char`, but `str`

content/developer/reference/standard_modules/account/account_account.rst
```diff
@@ -0,0 +1,18 @@
+.. _reference/account_account:
+
+=======
+Account
+=======
+
+.. autoclass:: odoo.addons.account.models.chart_template.AccountAccountTemplate
+
```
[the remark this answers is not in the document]
william-andre: We will add some in the source code. But branches are not synchronized, and we need more time for that part.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+.. seealso::
+   :ref:`Chart Template References <reference/account_chart_template>`
+
+.. example::
+  `addons/l10n_ch/data/l10n_ch_chart_data.xml <https://github.com/odoo/odoo/blob/15.0/addons/l10n_ch/data/l10n_ch_chart_data.xml>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/l10n_ch_chart_data.xml
+    :language: xml
+    :start-at: l10nch_chart_template
+    :end-at: </record>
```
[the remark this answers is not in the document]
william-andre: Will be outdated sooner than later then. We don't care if there are no examples for ppl who build the doc without the source code.
I want the doc to raise a warning if the source code is available but the `start-at` raises a warning for instance.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+    This tutorial requires knowledges about how to build a module in Odoo (see
+    :doc:`/developer/howtos/backend`).
+
+
+How localization works
+======================
+
+On installing the `account <https://github.com/odoo/odoo/tree/15.0/addons/account>`__ module, the localization module corresponding to the country code of the company is installed automatically.
+In case of no country code set or no localization module found, the `l10n_generic_coa <https://github.com/odoo/odoo/tree/15.0/addons/l10n_generic_coa>`__ (US) localization module is installed by default.
+Check `post init hook <https://github.com/odoo/odoo/blob/15.0/addons/account/__init__.py>`__ for details.
```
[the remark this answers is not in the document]
william-andre: It doesn't target devs. And trust me, it is not that easy to find.

To quote you:
> Did you ask newbies in your team to give their POV on this doc?
As developers, we easily believe everything is easy to understand in our own scope and to oversimplify the our explanations.

And yes. I asked newbies.

content/developer/howtos/localization.rst
william-andre: There is none (tested one or two years ago, after 13.0). 
It is just for readability.

content/developer/howtos/localization.rst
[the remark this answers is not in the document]
Feyensv: If it doesn't target devs, why is it in the `developer/` section then ?

conf.py
```diff
@@ -98,10 +105,14 @@
...
     # Parse Python docstrings (autodoc, automodule, autoattribute directives)
     'sphinx.ext.autodoc' if odoo_dir_in_path else 'autodoc_placeholder',
+    'autodoc_field',
 
     # Link sources in other projects (used to build the reference doc)
     'sphinx.ext.intersphinx',
 
+    # Embed Graphviz graphs
+    'sphinx.ext.graphviz',
+
```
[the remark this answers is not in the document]
william-andre: You are asking me to do changes. So yes, this is proof that it is better to have the source code available instead of a static picture. Now anyone can edit it without needing to redo everything.

content/developer/howtos/localization.rst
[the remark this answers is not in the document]
william-andre: It targets devs, but some are more "hardcore" than others. We often have PRs of ppl who can barely use a version control system for instance. Or devs du dimanche who want to help building a localization for their country, who know how to write xml but are afraid of digging in the code.

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import List
+
+import odoo
+import datetime
+
+MODEL_REFERENCE = {
+    'account.account.type': 'addons/account/data/data_account_type.xml',
+    'res.country': 'odoo/addons/base/data/res_country_data.xml',
+    'res.currency': 'odoo/addons/base/data/res_currency_data.xml',
+}
```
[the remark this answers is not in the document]
william-andre: > The dev can easily grep and find the file generating the base data.

Let's delete the docs then? Devs can find everything from the source code.

extensions/autodoc_field/__init__.py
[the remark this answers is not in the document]
Feyensv: Yes, I expect `something` to be a `char` but we're documenting Odoo fields.  I want the users to know what's behind a field, not only as value, but also as behavior, implied orm logic, choices of values, ...

If the display of a selection, reference, char, text & html fields is the same, then the autofield documenter doesn't do its job right...
Furthermore, we are in the dev doc, I expect the readers to know how to set the different types of fields (and if they don't, clicking on the class reference will redirect them to the needed documentation, even if some improvements are probably needed in the orm documentation).

content/developer/howtos/localization.rst
Feyensv: IIRC, csv records are created in batch, which isn't the case for xml records :thinking:

extensions/autodoc_field/__init__.py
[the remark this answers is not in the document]
william-andre: I completely disagree with that. I can add some better typing using the python module with the same name for fields if you want, but I am sure that I don't want `fields.Char`.
The end user, which might not be a real dev with a lot of experience, but just someone looking to do some RPC calls or to import/export data with the tools, doesn't care about the details of the field.

conf.py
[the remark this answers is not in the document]
oco-odoo: I agree with @william-andre ; if people have a warning when generating the doc and decide they don't care, it's their responsability. Having it directly in the code (and being generated automatically with the rest) makes the modifications of the graph way easier for everyone.

content/developer/howtos/localization.rst
oco-odoo: But on the other hand, it's already partly outdated, as account.tax.report objects didn't exist back then. Running exactly the examples from the presentation won't work. I'm not 100% against keeping it, but we need to keep that in mind. And I'm not sure we can say it explicitly here either, like "oh, we know we're giving you something a bit outdated, but who cares, we didn't want to make a new one" :p Maybe there is a more diplomatic way, though.

content/developer/howtos/localization.rst
oco-odoo: Hmmm, not convinced; to me it's giving a little too much details regarding what we are explaining here.

content/developer/reference/standard_modules/account/account_chart_template.rst
```diff
@@ -0,0 +1,62 @@
...
+    .. autofield:: name
+    .. autofield:: parent_id
+    .. autofield:: code_digits
+    .. autofield:: visible
+    .. autofield:: currency_id
+    .. autofield:: country_id
+    .. autofield:: use_anglo_saxon
+    .. autofield:: bank_account_code_prefix
+    .. autofield:: cash_account_code_prefix
+    .. autofield:: transfer_account_code_prefix
```
yelizariev, line 24: You lost my notes about those fields. The information from autofield is not enough

# Current version

image

# My version

image

yelizariev: https://github.com/odoo/documentation/pull/1301

content/developer/reference/standard_modules/account/account_chart_template.rst
```diff
@@ -0,0 +1,62 @@
...
+    .. autofield:: account_ids
+    .. autofield:: income_currency_exchange_account_id
+    .. autofield:: expense_currency_exchange_account_id
+    .. autofield:: account_journal_suspense_account_id
+    .. autofield:: account_journal_payment_debit_account_id
+    .. autofield:: account_journal_payment_credit_account_id
```
yelizariev, line 30: The same about suspense accounts. Don't make conspiracy about it. People should know the truth

# Current version
image

# My version

image

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+
+.. todo add reference to account_id in CoA
+
+.. example::
+  `addons/l10n_ch/data/l10n_ch_chart_post_data.xml <https://github.com/odoo/odoo/blob/15.0/addons/l10n_ch/data/l10n_ch_chart_post_data.xml>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_ch/data/l10n_ch_chart_post_data.xml
+    :language: xml
+    :start-at: l10nch_chart_template
+    :end-at: </record>
```
yelizariev, line 163: The example gives 7 fields only, while it's 21 fields

```xml

  <!-- Receivable/payable accounts. -->
  <field name="property_account_receivable_id" ref="..."/>
  <field name="property_account_payable_id" ref="..."/>

  <!-- Default Income/Expense Account for a Product Template. -->
  <field name="property_account_income_id" ref="..."/>
  <field name="property_account_expense_id" ref="..."/>

  <!-- Default Income/Expense Account for a Product Category. -->
  <field name="property_account_income_categ_id" ref="..."/>
  <field name="property_account_expense_categ_id" ref="..."/>

  <!-- Cash loss/gain accounts. They correspond profit_account_id / loss_account_id fields -->
  <field name="default_cash_difference_income_account_id" ref="..."/>
  <field name="default_cash_difference_expense_account_id" ref="..."/>

  <!-- Loss/gain exchange rate accounts. -->
  <field name="income_currency_exchange_account_id" ref="..."/>
  <field name="expense_currency_exchange_account_id" ref="..."/>

  <!-- Stock valuation accounts. -->
  <field name="property_stock_account_input_categ_id" ref="..."/>
  <field name="property_stock_account_output_categ_id" ref="..."/>
  <field name="property_stock_valuation_account_id" ref="..."/>

  <!-- Tax accounts to  balance current/advance tax payments. -->
  <!-- It's used in `env['account.generic.tax.report']._generate_tax_closing_entries` -->
  <field name="property_tax_payable_account_id" ref="..."/>
  <field name="property_tax_receivable_account_id" ref="..."/>
  <field name="property_advance_tax_payment_account_id" ref="..."/>


  <!-- Base Tax Received Account -->
  <!-- Corresponds to `account_cash_basis_base_account_id` field. -->
  <!-- It's used for taxes with `tax_exigibility` equal to `on_payment` -->
  <field name="property_cash_basis_base_account_id" ref="..."/>

  <!-- POS account. -->
  <field name="default_pos_receivable_account_id" ref="..."/>

  <!-- Payment accounts. -->
  <!--
       Suspense account is used to register an unreconiled bank entries created via bank statement.
       Once it's reconciled, the suspense account in `account.move` will be replaced with a proper account.
       Suspense account is an exceptional case when account is supposed to be replaced in the transaction.
       The new account depends on `account.payment` records. There are two cases
       1. `account.payment' exists before reconciliation aka Blue lines in reconciliation wizard
       2. `account.payment' is created on reconciliation aka regular Black lines in reconciliation wizard

       For case n.1 the new account is Outstanding Receipts/Payments account and the workflow is following:

       Initial transactions:
       * Invoice/bill: Payable/Receivable vs Expenses/Sales
       * Payment:             Outstanding vs Payable/Receivable
       After creating bank statement
       * Bank entry:                 Bank vs Suspense
       After reconciliation:
       * Bank entry:                 Bank vs Outstanding

       For case n.2 the new account is Payable/Receivable account from invoice/bill and the workflow is following:

       Initial transactions:
       * Invoice/bill: Payable/Receivable vs Expenses/Sales
       After creating bank statement
       * Bank entry:                 Bank vs Suspense
       After reconciliation:
       * Bank entry:                 Bank vs Payable/Receivable
  -->
  <field name="account_journal_suspense_account_id" ref="..."/>
  <!-- Outstanding Receipts Account -->
  <field name="account_journal_payment_debit_account_id" ref="..."/>
  <!-- Outstanding Payments Account -->
  <field name="account_journal_payment_credit_account_id" ref="..."/>

```

yelizariev: # Current version
image

content/developer/reference/standard_modules/account/account_chart_template.rst
william-andre: We will create a PR that adds `help` on the fields. It wouldn't be synchronized anyway.

content/developer/howtos/localization.rst
william-andre: Because you can't put them all at the same time, you first need to create the accounts, look below in the HowTo, where the steps are done in the right order

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+     :header-rows: 1
+
+Taxes
+-----
+
+.. seealso::
+   - :ref:`Tax References <reference/account_tax>`
+   - :doc:`/applications/finance/accounting/taxation/taxes/taxes`
+
+To add taxes you first need to specify tax groups. This model only has two required fields: name and country. Create file `data/account_tax_group_data.xml` and list the groups:
```
jco-odoo, line 189:
```suggestion
To add taxes you first need to specify tax groups. You normally need just one tax group for every tax rate, except for the 0% as you need to often distinguish between exempt, 0%, not subject, ... taxes. 
 This model only has two required fields: name and country. Create file `data/account_tax_group_data.xml` and list the groups:
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+    :end-before: </data>
+
+.. example::
+  `addons/l10n_uk/data/account.tax.group.csv <https://github.com/odoo/odoo/blob/15.0/addons/l10n_uk/data/account.tax.group.csv>`__.
+
+  .. literalinclude:: {ODOO_RELPATH}/addons/l10n_uk/data/account.tax.group.csv
+    :language: csv
+
+
+Now you can add the taxes via `data/account_tax_template_data.xml` file.
```
jco-odoo, line 220:
```suggestion
Now you can add the taxes via `data/account_tax_template_data.xml` file.  The first tax you define that is purchase/sale also becomes the default purchase/sale tax for your products.  
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,374 @@
...
+   - :ref:`Tax Report Line References <reference/account_tax_report_line>`
+   - :doc:`/applications/finance/accounting/reporting/declarations/tax_returns`
+
+In the previous section you noticed fields `invoice_repartition_line_ids` / `refund_repartition_line_ids` and probably understood nothing about them. Good news: you are not alone on this incomprehension. Bad news: you have to figure it out a bit. The topic is complicated indeed:
+
+.. graphviz:: images/tax_report.dot
+    :class: overflow-auto
+
+
+But fortunately we have a presentation explaining the tax reports (as in version 13.0) in details:
```
jco-odoo, line 252: @william-andre I have already commented about this in the other PR as well:  

```suggestion
The simple version is that in the tax template, you indicate in the invoice/refund repartition lines whether base or a percentage of the tax needs to be reported in which report line (through the minus/plus_report_line_ids).  It becomes clear also when you check the tax configuration in the Odoo interface (or check the docs (ref))  And fortunately we have a presentation explaining the tax reports (as in version 13.0) in details:
```

jco-odoo (commented):
There is a lot still missing.  The transfer and suspense accounts would be good to explain in a more functional documentation as well. (I like the way it is explained by @yelizariev and you can not just add that easily in the code)

redirects.txt
yelizariev, line 321:
```suggestion
developer/misc/i18n/localization.rst                    developer/howtos/localization.rst
developer/misc/i18n/translations.rst                    developer/howtos/translations.rst
```

content/developer/howtos/localization.rst
yelizariev: You probably misunderstand me. I'm talking about post-data file, not about the first file `account_chart_template_data`

content/developer/reference/standard_modules/account/account_chart_template.rst
yelizariev: Why it won't be synchronized ? You plan to add help to master branch, right?
We either need such PR to odoo/odoo v15 or keep fields description in odoo/documentation v15 for a while (and replace it  with generated docs in v16). I don't like the idea to hide the knowledge and enlighten new developers during special ceremony with accounting gurus.

content/developer/reference/standard_modules/account/account_chart_template.rst
yelizariev: https://github.com/odoo/documentation/pull/1334#discussion_r759242928

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,104 @@
+from sphinx.ext.autodoc import AttributeDocumenter
+from typing import List
+
+import odoo
```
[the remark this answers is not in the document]
william-andre: It should be fixed now, the extension is only loaded when the repo is found

extensions/autodoc_field/__init__.py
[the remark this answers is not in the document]
william-andre: I put it in the config

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.docstringable = (
+            odoo.addons.account.models.account_account_tag.AccountAccountTag,
+            odoo.addons.account.models.account_tax_report.AccountTaxReportLine,
+            odoo.addons.account.models.chart_template.AccountAccountTemplate,
+            odoo.addons.account.models.chart_template.AccountChartTemplate,
+            odoo.addons.account.models.chart_template.AccountFiscalPositionTemplate,
+            odoo.addons.account.models.chart_template.AccountGroupTemplate,
+            odoo.addons.account.models.chart_template.AccountTaxTemplate,
```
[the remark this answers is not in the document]
william-andre: I changed the reference system using `:ref:` and `.. autoodooclass:`

content/developer/reference/standard_modules/account/account_chart_template.rst
william-andre: It isn't synchronized because runbot team says it is too difficult to do to have a batch containing both the code and the doc

content/developer/howtos/localization.rst
william-andre: You have the reference for that, no need to put an example with every single account set. I'm not even sure it exists.

content/developer/howtos/localization.rst
william-andre: The only case where it's done doesn't follow this guideline, in l10n_de (and makes the code a mess)

content/developer/reference/standard_modules/account.rst
```diff
@@ -0,0 +1,16 @@
+:nosearch:
+
+==========
+Accounting
+==========
```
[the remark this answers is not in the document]
william-andre: Yes but no, we are also talking about `l10n*`. Devs who read the docs probably don't know the technical name of the module (yet).
They might also be just ppl who want to import/export with the right fields.

content/developer/howtos/images/tax_report.dot
william-andre: Is it good like this?

conf.py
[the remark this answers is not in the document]
william-andre: Using graphviz:
* Enables proper VCS:
  * track changes more easily
  * can collaborate on the same document, everyone has the code
  * conflict management
* No blob in the repo, lighter `.git`
* No need to restart everything during reviews

Using static file:
* Shorter compile time
* No need to install graphviz

Installing graphviz, and the time to compile the graph isn't long.
If you didn't install graphviz, you still get the document generated but without the graphs, that is totally fine. Exactly like compiling without the `odoo` repo.

content/developer/howtos/localization.rst
william-andre: I changed the link to the doc, maybe you'll find it more relevant

content/developer/howtos/localization.rst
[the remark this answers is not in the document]
william-andre: Fixed using a `condition` option

qdp-odoo:
@Feyensv all good for us, can you give a r+ ?

thanks

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,391 @@
...
+.. seealso::
+   - :ref:`Tax References <reference/account_tax>`
+   - :doc:`/applications/finance/accounting/taxation/taxes/taxes`
+
+To add taxes you first need to specify tax groups. You normally need just one tax group for every tax rate, except for the 0% as you need to often distinguish between exempt, 0%, not subject, ... taxes. 
+This model only has two required fields: name and country. Create file `data/account_tax_group_data.xml` and list the groups:
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="utf-8"?>
```
xmo-odoo, line 203: Leave out, it's unnecessary syntactic overhead (applies for every literal XML document)

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,391 @@
...
+    :end-at: </record>
+
+If some accounts should use default taxes, you can set them up in `data/account_account_template_post_data.xml`
+
+Tax Report
+----------
+
+.. raw:: html
+
+   <div><span class="badge" style="background-color:#AD5E99">Enterprise feature</span><div>
```
xmo-odoo, line 251: Might be a good use-case for an `rst_epilog` and a "standard" (project-wide) substitution.

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,391 @@
...
+.. seealso::
+   - :ref:`Tax Report Line References <reference/account_tax_report_line>`
+   - :doc:`/applications/finance/accounting/reporting/declarations/tax_returns`
+
+In the previous section you noticed fields `invoice_repartition_line_ids` / `refund_repartition_line_ids` and probably understood nothing about them. Good news: you are not alone on this incomprehension. Bad news: you have to figure it out a bit. The topic is complicated indeed:
+
+.. graphviz:: images/tax_report.dot
+    :class: overflow-auto
+
+The simple version is that in the tax template, you indicate in the invoice/refund repartition lines whether base or a percentage of the tax needs to be reported in which report line (through the minus/plus_report_line_ids).  It becomes clear also when you check the tax configuration in the Odoo interface (or check the docs (ref))
```
xmo-odoo, line 264: I don't understand a word of this. Also what's with the doubly-nested `(ref)`?

extensions/autodoc_placeholder/__init__.py
```diff
@@ -19,6 +19,8 @@ def setup(app):
     directives.register_directive('autodata', PlaceHolder)
     directives.register_directive('automethod', PlaceHolder)
     directives.register_directive('autoattribute', PlaceHolder)
+    directives.register_directive('autofield', PlaceHolder)
+    directives.register_directive('autoodooclass', PlaceHolder)
```
xmo-odoo, line 23: Since this uses the Odoo lingo of "field" for the class attributes, why not make this `automodel` rather than the difficult to read `autoodooclass`?

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,115 @@
+from sphinx.ext.autodoc import AttributeDocumenter, ClassDocumenter
+from typing import Sequence
+
+import odoo
+import datetime
+
+
+class OdooClassDocumenter(ClassDocumenter):
```
xmo-odoo, line 8: The printing of the class name is really gross:
Screenshot 2021-12-07 at 13-16-17 Account Tag — Odoo 15 0 documentation

And it doesn't seem super useful given users don't interact with it at all (whether to use or extend the model).

Maybe derive `class` as well, change the label to `model`, and that way we don't need to override `directivetype` because the objtype can match it?

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+                    self.add_line(f"|     `{tech}`: {nice}", source_name)
+        if field.type in ('many2one', 'one2many', 'many2many'):
+            comodel_name = field.comodel_name
+            string = f"| **Comodel:** :ref:`{comodel_name} <model-{comodel_name.replace('.', '-')}>`."
+            self.add_line(string, source_name)
+            reference = self.config.model_reference.get(comodel_name)
+            if reference:
+                self.add_line(f"| **Possible values:** `{reference} <{self.config.ultimate_replacements['{GITHUB_PATH}']}/{reference}>`__.", source_name)
+        if field.default:
+            self.add_line(f"| **Default:** {field.default(odoo.models.Model)}", source_name)
```
xmo-odoo, line 82: Seems like things we should use [fields lists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html?highlight=info%20fields) for, they nearly work OOTB (only issue I can see is value-less fields still have a `:` which is not great), and might be more amenable to flexible styling than hard-coding all the display properties in the generator.

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        if field.help:
+            for line in field.help.split('\n'):
+                self.add_line(f"| {line}", source_name)
```
xmo-odoo, line 85: Seems like a waste as it prevents marking up the help text with anything other than inline tags.

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        self.add_line(f"| **Name:** {field.string}", source_name)
+        if field.readonly:
+            self.add_line(f"| **Read-only:** this field is not supposed to/cannot be set manually", source_name)
+        if not field.store:
+            self.add_line(f"| **Not stored:** this field is there only for technical reasons", source_name)
+        if field.type == 'selection':
+            if isinstance(field.selection, (list, tuple)):
+                self.add_line(f"| **Possible values:**", source_name)
+                for tech, nice in field.selection:
+                    self.add_line(f"|     `{tech}`: {nice}", source_name)
```
xmo-odoo, line 73:
```suggestion
                    self.add_line(f"|     ``{tech}``: {nice}", source_name)
```
While leveraging the default-role might be acceptable for content written up directly, it seems quite unnecessary for  generated content.

extensions/autodoc_field/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        return res
+
+def disable_warn_missing_reference(app, domain, node):
+    if (domain and domain.name != 'std') or node['reftype'] != 'ref':
+        return None
+    else:
+        target = node['reftarget']
+        if target.startswith('model-'):
+            node['reftype'] = 'odoo_missing_ref'
+            return True
```
xmo-odoo, line 103:
```suggestion
    if (not domain or domain.name == 'std') and node['reftype'] == 'ref':
        target = node['reftarget']
        if target.startswith('model-'):
            node['reftype'] = 'odoo_missing_ref'
            return True
```
Seems simpler and to work the same? Though the first part of teh guard seems a bit weird.

Also a docstring to more clearly explain the purpose of this handler would be nice.

xmo-odoo (changes requested):

conf.py
```diff
@@ -75,7 +80,11 @@
     )
 else:
     odoo_dir = odoo_sources_dirs[0].resolve()
+    ultimate_replacements['{ODOO_RELPATH}'] = str(Path('/../' + str(odoo_sources_dirs[0])))
```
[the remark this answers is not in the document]
william-andre: It is an "absolute" related path, that starts with a `/`

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude
> The file name is usually relative to the current file’s path. However, if it is absolute (starting with /), it is relative to the top source directory.

conf.py
```diff
@@ -75,7 +80,11 @@
...
+    ultimate_replacements['{ODOO_ABSPATH}'] = str(odoo_dir)
     sys.path.insert(0, str(odoo_dir))
+    import odoo.addons
+    odoo.addons.__path__ += [str(Path(str(odoo_dir) + '/addons'))]
```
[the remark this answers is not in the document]
william-andre: Is it safe for every OS?

extensions/autodoc_field/__init__.py
william-andre: I'm not sure I understand what you mean

conf.py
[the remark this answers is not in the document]
william-andre: SVG is a lot less easy to manipulate though

conf.py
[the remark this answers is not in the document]
xmo-odoo: ...

ok?

conf.py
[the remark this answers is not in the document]
xmo-odoo: It's certainly no less safe than the original, but if you fear semantics judt remove the outer `str()` and `Path` calls, they do nothing.

conf.py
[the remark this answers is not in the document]
william-andre: What I meant is that this is not a "related path" per se.

And is it safe to not use `Path`, regarding OS differences?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,391 @@
...
+
+Account tags
+------------
+
+.. seealso::
+   :ref:`Account Tag References <reference/account_account_tag>`
+
+Tags are a way to sort accounts.
+For example, imagine you want to create a financial report having multiple lines but you have no way to find a rule to dispatch the accounts according their ``code``.
+The solution is the usage of tags, one for each report line, to filter accounts like you want.
```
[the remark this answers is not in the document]
xmo-odoo: I know how rST works. This just looks schisophrenic.

Either the breaks are necessary for readability and they should also be in the rendered output, or they're not and they don't belong in the source. If the goal was to avoid overlong lines (which seems unlikely as the second line is >150) then it should be fill-paragraph'd.

conf.py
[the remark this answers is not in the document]
xmo-odoo: > What I meant is that this is not a "related path" per se.

Sure.

> And is it safe to not use Path, regarding OS differences?

Considering this use of `Path` does absolutely nothing, it's certainly no loss in safety.

extensions/autodoc_field/__init__.py
xmo-odoo: Which part?

* showing the classname is not useful since odoo developers don't interact with class(names) directly
* the class name is misrendered (cf boxes separating module and class names)
* the `class` moniker (leading keyword) is confusing as in odoo terms this is a model
* both could be resolved by overriding the class directive with one using the `model` moniker and having the auto-documenter give it the model name as "class name".

content/developer/reference/standard_modules/account/account_account.rst
```diff
@@ -0,0 +1,18 @@
...
+    .. autofield:: name
+    .. autofield:: currency_id
+    .. autofield:: code
+    .. autofield:: user_type_id
+    .. autofield:: reconcile
+    .. autofield:: note
+    .. autofield:: tax_ids
+    .. autofield:: nocreate
+    .. autofield:: chart_template_id
+    .. autofield:: tag_ids
```
xmo-odoo, line 18: Do these have to be listed as autofields, can't `:members:` do the job? That would make the *auto* part of autoodooclass a lot truer.

extensions/autodoc_field/__init__.py
william-andre: I changed the label to "model", and I also changed the class module/name, but it didn't look good with the link to the sources it makes sense to have the full class name.
It is also good to see in an eye what Odoo module we are referring to.

content/developer/reference/standard_modules/account/account_account.rst
william-andre: `:members:` is not that good of an idea to me because
* we rarely want all the fields (we don't care about computed and technical fields in this API)
* I like to have one field per line for VCS reasons

xmo-odoo: > `:members:` is not that good of an idea to me because
> 
>  * we rarely want all the fields (we don't care about computed and technical fields in this API)

That can be handled through selection flags in exactly the way autodoc normally works, and `:members:` supports whitelisting.

> * I like to have one field per line for VCS reasons

:shrug:

william-andre: Ok, but what's the difference, apart from more development? If we always need to whitelist, then what's the point?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+     :header-rows: 1
+
+Taxes
+-----
+
+.. seealso::
+   - :ref:`Tax References <reference/account_tax>`
+   - :doc:`/applications/finance/accounting/taxation/taxes/taxes`
+
+To add taxes you first need to specify tax groups. You normally need just one tax group for every tax rate, except for the 0% as you need to often distinguish between exempt, 0%, not subject, ... taxes.
```
poma-odoo, line 198: Is it useful to mention tax groups' subtotals are displayed below the invoice, but tags are used in tax reports?

william-andre: Probably, you are welcome to suggest some changes! :)

jco-odoo: There are plenty ideas for changes, but we thought to have this one merged first

william-andre: I want to merge it as soon as it is good on the tech pov for everyone.
While it isn't, anyone can already add some functional things, no need to delay that.

edi-odoo (approved):
LGTM

qdp-odoo:
@odoo/doc-review can we find an agreement on the graphviz thingy? If the only problem you have with it is the fact that it was logging a warning in case it wasn't installed, I think it has been taken care of. If it's still a friction point for you, can you say again why so that we can carry on?

All the points related to the accounting/functional content itself will be addressed later on, but your opinion on the technical part of this PR really matters to us and you'll truly be gentlemen to point us where we still have effort to make to convince you ;-) :pray: 

thanks

AntoineVDV (changes requested):
> https://github.com/orgs/odoo/teams/doc-review can we find an agreement on the graphviz thingy? If the only problem you have with it is the fact that it was logging a warning in case it wasn't installed, I think it has been taken care of. If it's still a friction point for you, can you say again why so that we can carry on?

Hello, it seems like all the blocking points have been dealt with so it's good for us :)
I spoke with @d-fence who is okay to install graphviz on runbot's docker but you still have to convince @mart-e to do the same for internal 😅

> All the points related to the accounting/functional content itself will be addressed later on, but your opinion on the technical part of this PR really matters to us and you'll truly be gentlemen to point us where we still have effort to make to convince you ;-) pray

I'm posting all my comments that are *not* related to the content *per se* now because GitHub is dans les choux today and it's a PITA to write comments until it gets better. Another pass of review focused on the RST of the content will follow soon.

> ho, also if you could guide us in fixing the codeowner errors from ci/documentation that would be super nice of you grin

I will create a codeowner team and assign it to the new documentation paths. Who should I put in this team? They would be responsible for maintaining the content added with this PR and handle community PRs targeting this content (I doubt there will be any).

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+    :condition: odoo_dir_in_path
+    :language: xml
+    :start-at: fp_nacional
+    :end-before: fp_intra
+
+Final Steps
+===========
+
+The last step when installing a localization module is to try applying its chart of accounts to the current company (if it does not already have one).
+File `data/account_chart_template_try_loading.xml` is responsible for that.
```
AntoineVDV, line 319:
```suggestion
The file :file:`data/account_chart_template_try_loading.xml` is responsible for that.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+==================
+
+.. raw:: html
+
+   <div><span class="badge" style="background-color:#AD5E99">Enterprise feature</span><div>
+
+.. seealso::
+  :doc:`/applications/finance/accounting/reporting/overview`
+
+Accounting reports should be added via a separate module **l10n_XX_reports** that should go to the `enterprise repository <https://github.com/odoo/enterprise>`__.
```
AntoineVDV, line 351:
```suggestion
Accounting reports should be added via a separate module `l10n_XX_reports` that should go to the `enterprise repository <https://github.com/odoo/enterprise>`_.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+        ],
+        "auto_install": True,
+    }
+
+
+Functional overview of financial reports is here: :doc:`/applications/finance/accounting/reporting/overview/main_reports`.
+
+Some good examples:
+
+* `l10n_ch_reports/data/account_financial_html_report_data.xml <https://github.com/odoo/enterprise/blob/15.0/l10n_ch_reports/data/account_financial_html_report_data.xml>`__
```
AntoineVDV, line 377: Wouldn't that make sense to have a replacement for enterprise URLs too?

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+* `l10n_be_reports/data/account_financial_html_report_data.xml <https://github.com/odoo/enterprise/blob/15.0/l10n_be_reports/data/account_financial_html_report_data.xml>`__
+
+For the fields meaning dive directly to the source:
+
+* `account.financial.html.report (v15) <https://github.com/odoo/enterprise/blob/d4eff9d39469cf3fe18589a1547cb0cdb93f4ae9/account_reports/models/account_financial_report.py#L59-L75>`__
+* `account.financial.html.report.line (v15) <https://github.com/odoo/enterprise/blob/d4eff9d39469cf3fe18589a1547cb0cdb93f4ae9/account_reports/models/account_financial_report.py#L931-L964>`__
+
+Menu for the new report is created automatically. By default it's located under *Accounting >> Reporting* menu.
```
AntoineVDV, line 385:
```suggestion
The menu for the new report is created automatically. By default, it is located under :menuselection:`Accounting -> Reporting`.
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+
+
+Functional overview of financial reports is here: :doc:`/applications/finance/accounting/reporting/overview/main_reports`.
+
+Some good examples:
+
+* `l10n_ch_reports/data/account_financial_html_report_data.xml <https://github.com/odoo/enterprise/blob/15.0/l10n_ch_reports/data/account_financial_html_report_data.xml>`__
+* `l10n_be_reports/data/account_financial_html_report_data.xml <https://github.com/odoo/enterprise/blob/15.0/l10n_be_reports/data/account_financial_html_report_data.xml>`__
+
+For the fields meaning dive directly to the source:
```
AntoineVDV, line 380:
```suggestion
For the fields' meaning, dive directly to the source:
```

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+
+* `account.financial.html.report (v15) <https://github.com/odoo/enterprise/blob/d4eff9d39469cf3fe18589a1547cb0cdb93f4ae9/account_reports/models/account_financial_report.py#L59-L75>`__
+* `account.financial.html.report.line (v15) <https://github.com/odoo/enterprise/blob/d4eff9d39469cf3fe18589a1547cb0cdb93f4ae9/account_reports/models/account_financial_report.py#L931-L964>`__
+
+Menu for the new report is created automatically. By default it's located under *Accounting >> Reporting* menu.
+To create a dedicated section in Reporting menu, you need to create new `ir.ui.menu` record (usually in the main `l10n_XX` module) and set it as `parent_id` field in `account.financial.html.report` model. Example for Belgium localization:
```
AntoineVDV, line 386:
```suggestion
To create a dedicated section in the :guilabel:`Reporting` menu, you need to create a new `ir.ui.menu` record (usually in the main `l10n_XX` module) and set it as `parent_id` field in the `account.financial.html.report` model. Example for the Belgian localization:
```

content/developer/reference/standard_modules/account/account_tax_report_line.rst
```diff
@@ -0,0 +1,24 @@
+.. _reference/account_tax_report_line:
+
+===============
+Tax Report line
```
AntoineVDV, line 4:
```suggestion
Tax Report Line
```

content/developer/reference/standard_modules.rst
```diff
@@ -0,0 +1,10 @@
+:nosearch:
+
+================
+Standard Modules
```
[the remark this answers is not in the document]
william-andre: Ok, done.

But better than the *title case* for titles?
https://en.wikipedia.org/wiki/Title_case

conf.py
```diff
@@ -1,9 +1,12 @@
 import re
 import os
+import shutil
 import sys
 from pathlib import Path
 
+import docutils
```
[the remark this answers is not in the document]
william-andre: I added the same version as in sphinx 
https://github.com/sphinx-doc/sphinx/blob/335bf513e85eb2727322a2be9a731af427584fe0/setup.py#L23

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
...
+    :end-at: </record>
+
+Accounts
+--------
+
+.. seealso::
+   - :ref:`Account References <reference/account_account>`
+   - :doc:`/applications/finance/accounting/getting_started/initial_configuration/chart_of_accounts`
+
+Obviously, *Chart of Accounts* cannot exist without *Accounts*. You need to specify them in `data/account.account.template.csv`.
```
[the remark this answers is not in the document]
william-andre: Is there a documentation to where we can find all the roles (like `guilabel` or `file`) that we should use?

content/developer/reference/standard_modules/account/account_tax_report_line.rst
william-andre: I'm confused. Do you want sentence case or title case for the titles?

conf.py
```diff
@@ -98,15 +109,34 @@
...
         odoo_dir_in_path = True
 
+
+# Fail silently if `dot` is not installed and
+original_render_dot = graphviz.render_dot
+def patched_render_dot(*args, **kwargs):
+    if shutil.which('dot'):
+        return original_render_dot(*args, **kwargs)
+    return (None, None)
+graphviz.render_dot = patched_render_dot
```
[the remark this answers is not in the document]
william-andre: Cleaner indeed :+1:

AntoineVDV (commented):
I noticed that you missed a few review comments:
- [One about moving the standard module references into 'backend'](https://github.com/odoo/documentation/pull/1334#discussion_r856297924). I pushed a fixup commit doing just that to see how it looks. You can resolve the comment and squash the commit if you're ok with it.
- [Another about using external hyperlinks instead of anchors](https://github.com/odoo/documentation/pull/1334#discussion_r858746757). This one is not up for debate so I fixed the issue. I also amended some commits to make additional trivial changes.
- [This one about re-using the tooling for GitHub links for the enterprise repository](https://github.com/odoo/documentation/pull/1334#discussion_r858762553).

content/developer/reference/standard_modules.rst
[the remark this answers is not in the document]
AntoineVDV: That'd be a question for @jcs-odoo as I didn't pick this guideline, I just make sure we merge diffs that are consistent with it ;)

content/developer/howtos/localization.rst
```diff
@@ -0,0 +1,389 @@
+
+=======================
+Accounting Localization
```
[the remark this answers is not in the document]
AntoineVDV: My bad, I didn't know about this rule.

content/developer/howtos/localization.rst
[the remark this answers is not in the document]
AntoineVDV: Not *yet*: https://www.odoo.com/web#id=2810468&cids=1&menu_id=5195&action=333&active_id=3835&model=project.task&view_type=form

content/developer/reference/standard_modules/account/account_tax_report_line.rst
AntoineVDV: I suggested this change because it's an Odoo model name. The [guidelines](https://www.odoo.com/documentation/master/contributing/documentation/content_guidelines.html#titles-and-headings) recommend capitalizing *feature names* so I'm not sure about this one. I'd go for the same naming convention as in Odoo... @jcs-odoo what do you think?

content/developer/reference.rst
```diff
@@ -10,3 +10,4 @@ Reference
 
     reference/backend
     reference/frontend
+    reference/standard_modules
```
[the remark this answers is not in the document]
william-andre: It is not part of the backend though. The backend is the ORM.

This is a functional (from a tech PoV) thing, where we document models, not the API of how Odoo works.
It is destined to people who want to create/update data, not write python code.

So I personally do not agree

content/developer/reference/standard_modules/account/account_tax_report_line.rst
jcs-odoo: @william-andre @AntoineVDV 

Indeed, the guidelines are too blurry about the exceptions to the sentence case.
https://www.odoo.com/documentation/master/contributing/documentation/content_guidelines.html#titles-and-headings

I intend to change them slightly by removing the exception about feature names (but keeping the other exceptions). This should remove 99,9% of doubts and make it easier for everyone as it would be with sentence case pretty much everywhere :)

Here is an example.
image

Actually, I'll make a PR for that right now :)

jcs-odoo: Here it is ^^
https://github.com/odoo/documentation/pull/1910

content/developer/howtos/localization.rst
william-andre: Done

AntoineVDV (approved):
Waiting for confirmation that it will build fine on odoo.com before merging.

william-andre:
@robodoo rebase-ff ? :pray:

jco-odoo:
@AntoineVDV

AntoineVDV:
@robodoo rebase-ff

PR #1345, opened 2021-11-30
Typo fix

PermanAtayev:


jcs-odoo (changes requested):
Thank you for your contribution!
However, the target can be 13.0 instead of 15.0.
I'll create a new PR with your commit right now.

jcs-odoo:
backported in #1394

PR #1349, opened 2021-12-03
[FIX] rdtraining: fix wrong suggested folder for views

hokolomopo:
The training suggests to put the views in the data/ folder instead of the views/ folder


AntoineVDV:
@robodoo r+

PR #1352, opened 2021-12-06
Improvements and fixes to the API documentation

xmo-odoo:
Somewhat related to odoo/odoo#74604  as these are issues I found while fixing the other one (and the first commit is quite literally support for a change in odoo/odoo#74604).

### DIFF
--- a/content/developer/reference/backend/orm.rst
+++ b/content/developer/reference/backend/orm.rst
@@ -1,4 +1,3 @@
-
 .. _reference/orm:
 
 =======
@@ -388,12 +387,14 @@ dependencies are modified.
       f_ids = fields.Many2many(related="m2m_ids.m2m_ids")
       g_ids = fields.One2many(related="o2m_ids.o2m_ids")
 
+.. currentmodule:: odoo.models
+
 .. _reference/fields/automatic:
 
 Automatic fields
 ----------------
 
-.. attribute:: id
+.. attribute:: Model.id
 
     Identifier :class:`field <odoo.fields.Field>`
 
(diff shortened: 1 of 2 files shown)


xmo-odoo:
Also hit a weird issue but didn't manage to debug it and ended up ignoring it: since 2.4 `py:attribute` is supposed to [have a `type` field](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-attribute).

It works fine when creating a new sphix project, but in the odoo doc project it fails and weirdly it seems to fail *during parsing* (the content which reaches the translator is not correct) but I didn't find any place where we'd override the parsing, only the translation?

xmo-odoo:
Also the CI says it doesn't find "build succeeded" in the log file, but when looking at the logfile it's right there O_o

Feyensv:
It doesn't like the "4 warnings" part :P

xmo-odoo:
> It doesn't like the "4 warnings" part :P

Oh I'd missed that there were actual warnings during the build, I only saw the ones at the top and figured they had nothing to do with me.

It's probably unhappy because those are docstrings I fixed in odoo/odoo#74604 and it always (I assume) builds based on the current master.

content/developer/reference/backend/orm.rst
```diff
@@ -445,44 +446,50 @@ A few field names are reserved for pre-defined behaviors beyond that of
...
-  necessary.
+   default value for :attr:`~odoo.models.BaseModel._rec_name`, used to
+   display records in context where a representative "naming" is
+   necessary.
 
-  :class:`~odoo.fields.Char`
+   :class:`~odoo.fields.Char`
 
-.. attribute:: active
+.. data:: Mode.active
```
AntoineVDV, line 457: L

AntoineVDV (approved):
LGTM (except for the typo).

I'm flagging the PR as a draft until the odoo/odoo counterpart to ease the follow-up.

@robodoo delegate+

xmo-odoo:
@robodoo r+

xmo-odoo:
@robodoo r+

xmo-odoo:
@robodoo rebase-ff

PR #1353, opened 2021-12-06
Class name changed

luxiaochuang:
For the learner who wants to test the code in the IDE, the code should be executable.
I changed the Class name to the self. 
For reference.


AntoineVDV (commented):
Hi, this PR is refused because, if anything, it adds more confusion to the training. The current excerpt of code is not meant to be copy-pasted as-is in a method on the `res.partner` model but to clearly show the relationship between the `test.model` and `res.partner` models.

PR #1354, opened 2021-12-06
We recommand that the many2one is created Manually

luxiaochuang:
Although the odoo system is smart enough. However. we should obey the database rules about the relationships among the tables. for example database NFs. otherwise . the beginners will be confused.

### DIFF
--- a/content/developer/howtos/rdtraining/08_relations.rst
+++ b/content/developer/howtos/rdtraining/08_relations.rst
@@ -257,9 +257,8 @@ exercise: an offer is always accessed through a property.
 
 Second, despite the fact that the ``property_id`` field is required, we did not include it in the
 views. How does Odoo know which property our offer is linked to? Well that's part of the
-magic of using the Odoo framework: sometimes things are defined implicitly. When we create
-a record through a one2many field, the corresponding many2one is populated automatically
-for convenience.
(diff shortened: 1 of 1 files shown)


Feyensv:
Do you mind clarifying what you mean by "Creating the many2one manually" ?
Do you mean using `record = create(m2o=self.id)` instead of `self.x2m = [(0, 0, {}]` ?
Honestly, in some cases, the use of `create` is wrong (E.g. in onchanges/computes).

luxiaochuang:
@Feyensv, 
Hi, Good morning.
The original document said that  “When we create a record through an one2many field, the corresponding many2one is populated automatically for convenience”, which is located at the end of chapter 8.

However. when we define the relationship between the 2 tables, we should tell the learner that the One2many field and the Many2one field are needed separately, although the odoo is smart and powerful.

The benefit is obviously that everyone who learned the database system concepts about the Normal Forms can easily understand and develop before. after which, we can introduce the above new feature of the odoo. For example, self.x2m = [(0, 0, {}],that you mentioned is the further knowledge for the learner.

Just the suggestion for your reference.

Thanks.

AntoineVDV:
Hi, @luxiaochuang thank you for your suggestion but I believe that, if anything, it adds more confusion. We'll soon clarify the pain points of the technical training in a separate PR.

PR #1359, opened 2021-12-07
[FIX] developer/reference/javascript_reference

Julien00859:
The "qweb key of the manifest" file was still mentionned even if it has
been replaced by the assets bundles.

See related task: 2352566

See also: odoo/odoo#80988

### DIFF
--- a/content/developer/reference/frontend/javascript_reference.rst
+++ b/content/developer/reference/frontend/javascript_reference.rst
@@ -311,8 +311,9 @@ Here is an example of a basic counter widget:
     });
 
 For this example, assume that the template *some.template* (and is properly
-loaded: the template is in a file, which is properly defined in the *qweb* key
-in the module manifest) is given by:
+loaded: the template is in a file, which is properly defined in the assets of
+the module manifest ``'assets': {'web.assets_qweb': [...]}``,  see
+:ref:`assets <reference/assets>`.) is given by:
 
 .. code-block:: xml
 


lordkrandel:
What about a link to https://www.odoo.com/documentation/15.0/developer/reference/frontend/assets.html#reference-assets ?

Julien00859:
The page itself doesn't bring useful information regarding qweb, not sure it is worth to link it

lordkrandel:
I disagree. New users use the Javascript Reference almost like a tutorial to learn how to make a Widget. In this case, it says that the template you want has to be loaded before you can use it, and that it must be present in the `__manifest__.py`.
immagine

And the would-be linked page actually tells you that the templates are part of a bundle, and informs you about what key is to be used in `__manifest__.py`.

immagine

Julien00859:
Tu veux bien trancher Antoine stp ? J'ai bien compris l'argument de Paolo et il a également compris le mien. On est reste chacun convaincu de notre position donc on a besoin d'une 3e personne pour trancher :D

AntoineVDV:
@Julien00859 I agree with @lordkrandel on this one. The page on assets doesn't say much about QWeb but we don't want it to. What matters is that:
- If developers struggle to add their template into a bundle, the linked doc should help them do it.
- If developers are curious about this particular instruction, they should be able to learn right away about assets bundles. To me, this is somewhat similar to my first ever coding lesson (in C++) when I was told to ignore 90% of the code because I'd figure it out later (`using namespace std;`, etc.).

Julien00859:
Done :)

Julien00859:
@robodoo r+

AntoineVDV:
@Julien00859 Since @ged-odoo was pinged I'll let him have a look at this PR before merging it ;)

ged-odoo (approved):

AntoineVDV:
@robodoo r+

PR #1360, opened 2021-12-07
[FIX] remove enterprise tag from Gantt documentation

frln:
Gantt view is implemented in community, see https://github.com/odoo/odoo/blob/c3673e63064c7f331754ef0e9d02cbbe7b1927f4/odoo/addons/base/models/ir_ui_view.py#L246

### DIFF
--- a/content/developer/reference/backend/views.rst
+++ b/content/developer/reference/backend/views.rst
@@ -956,10 +956,6 @@ Generic structure
 Gantt
 -----
 
-.. raw:: html
-
-   <span class="badge" style="background-color:#AD5E99">Enterprise feature</span>
-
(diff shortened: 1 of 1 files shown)


AntoineVDV:
Hi, @ged-odoo could you double-check this? The type is indeed available in community but I guess that it's a little 'hack' to avoid overriding the selection in enterprise since the module `web_gantt` is in enterprise.

mart-e:
Yes, I don't think it's actually available in community, the implementation is in `web_gantt`

frln:
Yeah never mind I checked and the view is defined on the work orders action in manufacturing and it doesn't show up so it's not available in community, I spoke too soon sorry!

PR #1364, opened 2021-12-08
[IMP] install/setup: install dependencies from debian packages

d-fence:
This commit changes the way that Odoo python dependencies should be
installed.

Before this commit, the described way to install Odoo dependencies was
by using the `pip` utility.

Now, we explicitly recommend to use a Debian based system and install
Odoo dependencies from their packaging tool.

The main reasons are:
* Odoo productions instances are, most of the time, run on Debian based
  systems with those package already installed
* That way, a developer cannot mistakenly use a feature from a
  dependency that does not exists on those production envirorments
* Avoid mixes of python packages on the developer/user 's system

### DIFF
--- a/content/administration/install/install.rst
+++ b/content/administration/install/install.rst
@@ -505,50 +505,75 @@ create a new PostgreSQL user:
 .. note:: Because your PostgreSQL user has the same name as your Unix login, you will be able to
           connect to the database without password.
 
-Dependencies
-^^^^^^^^^^^^
+.. _install/python-dependencies:
 
-For libraries using native code, it is necessary to install development tools and native
-dependencies before the Python dependencies of Odoo. They are available in `-dev` or `-devel`
-packages for Python, PostgreSQL, libxml2, libxslt1, libevent, libsasl2 and libldap2.
+Python Dependencies
+^^^^^^^^^^^^^^^^^^^
 
-On Debian/Unbuntu, the following command should install all the required libraries:
+Debian/Ubuntu
+~~~~~~~~~~~~~
+
+Using your **distribution packages** is the preferred way of installing dependencies.
+
+For Debian-based systems, these packages are listed in the `debian/control
+<https://github.com/odoo/odoo/blob/master/debian/control>`_ file of the Odoo sources.
+
+On Debian/Ubuntu, the following commands should install the required packages:
 
 .. code-block:: console
 
-    $ sudo apt install python3-dev libxml2-dev libxslt1-dev libldap2-dev libsasl2-dev \
-        libtiff5-dev libjpeg8-dev libopenjp2-7-dev zlib1g-dev libfreetype6-dev \
-        liblcms2-dev libwebp-dev libharfbuzz-dev libfribidi-dev libxcb1-dev libpq-dev
+   $ cd /CommunityPath
+   $ sed -n -e '/^Depends:/,/^Pre/ s/ python3-\(.*\),/python3-\1/p' debian/control | sudo xargs apt-get install -y
 
-Odoo dependencies are listed in the `requirements.txt` file located at the root of the Odoo
+Install with pip
+~~~~~~~~~~~~~~~~
+
+Alternatively, you can use **pip** to install python dependencies. As some of the packages need a
+compilation step, they require system libraries to be installed.
+
+On Debian/Ubuntu-based systems, the following command should install these required libraries:
+
+.. code-block:: console
+
+   $ sudo apt install python3-pip libldap2-dev libpq-dev libsasl2-dev
+
+Odoo dependencies are listed in the :file:`requirements.txt` file located at the root of the Odoo
 community directory.
 
-.. tip:: It can be preferable to not mix python modules packages between different instances of Odoo
-         or with your system. You can use virtualenv_ to create isolated Python environments.
+.. note::
+   The python packages in :file:`requirements.txt` are based on their stable/LTS Debian/Ubuntu
+   corresponding version at the moment of the Odoo release.
 
-Navigate to the path of your Odoo Community installation (`CommunityPath`) and run **pip**
-on the requirements file:
+   E.g., for Odoo 15.0, the `python3-babel` package version is 2.8.0 in Debian Bullseye and 2.6.0
+   in Ubuntu Focal. The lowest version is then chosen in the :file:`requirements.txt`.
+
+Navigate to the path of your Odoo Community installation (:file:`CommunityPath`) and run **pip** on
+the requirements file to install the requirements for the current user.
 
 .. code-block:: console
 
-    $ cd /CommunityPath
-    $ pip3 install setuptools wheel
-    $ pip3 install -r requirements.txt
+   $ cd /CommunityPath
+   $ pip install -r requirements.txt
 
-.. warning:: `wkhtmltopdf` is not installed through **pip** and must be installed manually in
-             version `0.12.5 <the wkhtmltopdf download page_>`_ for it to support headers and
-             footers. See our `wiki <https://github.com/odoo/odoo/wiki/Wkhtmltopdf>`_ for more
-             details on the various versions.
+.. tip::
+   It can be preferable not to mix python modules packages between different instances of Odoo or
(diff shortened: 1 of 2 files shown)


content/administration/install/install.rst
```diff
@@ -508,37 +508,27 @@ create a new PostgreSQL user:
...
-        liblcms2-dev libwebp-dev libharfbuzz-dev libfribidi-dev libxcb1-dev libpq-dev
-
-Odoo dependencies are listed in the `requirements.txt` file located at the root of the Odoo
-community directory.
+    $ cd /CommunityPath
+    $ sed -n -e '/^Depends:/,/^Pre/ s/ python3-\(.*\),/python3-\1/p' debian/control | sudo xargs apt-get install -y
 
 .. tip:: It can be preferable to not mix python modules packages between different instances of Odoo
          or with your system. You can use virtualenv_ to create isolated Python environments.
+         In that case, the `requirements.txt` file can be used to install the dependencies.
```
Xavier-Do, line 522: I think that we should specify here that -dev packages are required. It is sad to loose this list.

d-fence: That means bringing back the whole pip explanation in the tip ... what do you think @AntoineVDV

d-fence:
Linked to odoo/documentation#1329

amigrave (commented):
2 cents

AntoineVDV (changes requested):
I find it weird to keep only *some* of the explanations for the `pip` setup while removing the others (list of `-dev` packages, for example). Either we recommend installing the dependencies through `apt` and we properly explain how to do it through `pip` for those who prefer, or we only mention the Debian setup. If we go for the first option, I suggest that we properly split the instructions for the two setups in two different headings, and find a nicer-looking solution later on (tabs, maybe).

I'm a bit confused about why we'd want to discourage the `pip` setup. Maybe @amigrave could enlighten me? And, if we do that, why do we keep (and maintain) a `requirements.txt` file? Also, are Debian dependencies listed in the `debian/control` file compatible with any supported version of Odoo?

content/developer/howtos/rdtraining/02_setup.rst
```diff
@@ -170,24 +170,16 @@ You can check your Python version with:
...
 
+The preferred way of installing dependencies is by using your distribution packages.
+For Debian based systems, those packages are listed in the `debian/control` file of the odoo sources.
+On Debian/Ubuntu, the following command should install the required packages:
+
 .. code-block:: console
 
-    $ cd $HOME/src/odoo
-    $ pip3 install -r requirements.txt
+    $ sed -n -e '/^Depends:/,/^Pre/ s/ python3-\(.*\),/python3-\1/p' debian/control | sudo xargs apt-get install -y
```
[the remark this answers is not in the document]
d-fence: Why is it going to get out of sync quickly ? We use that kind of stuff for a long time in our runbot Docker files.
`sed` is only 4 years younger than me and will probably survive me.
Joking aside, the `debian/control` file is kept in sync in Odoo, so it's probably the best thing to do to stay in sync with Debian dependencies.

d-fence:
@AntoineVDV I splitted the installation of `Python dependencies` into `debian/Ubuntu` and `The pip way`.
I hope that it will not bring more confusion in the chaos...
And I'm not sure that @amigrave will agree

content/developer/howtos/rdtraining/02_setup.rst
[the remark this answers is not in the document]
amigrave: @d-fence I guess that Antoine means that this duplicated content will get out of sync compared to the install page and that we could take the opportunity to remove some sections of this rdtraining setup doc in favor of a link to the install page. I might be extrapolating Antoine's comment a bit though :)

amigrave:
> @AntoineVDV I splitted the installation of `Python dependencies` into `debian/Ubuntu` and `The pip way`. I hope that it will not bring more confusion in the chaos... And I'm not sure that @amigrave will agree

No, I'm ok with that. I'll link from the Odoo.sh FAQ to the "preferred way" of installation.

AntoineVDV:
> > @AntoineVDV I splitted the installation of `Python dependencies` into `debian/Ubuntu` and `The pip way`. I hope that it will not bring more confusion in the chaos... And I'm not sure that @amigrave will agree
> 
> No, I'm ok with that. I'll link from the Odoo.sh FAQ to the "preferred way" of installation. Maybe mention that the Odoo platforms (Odoo Online and Odoo.sh) uses the debian/ubuntu installation method ? Not mandatory though

I don't think that it adds any value from the POV of the doc.

AntoineVDV (approved):
I took the liberty to make some fine-tuning changes:
- RST cleanup.
- Some wording fixes.
- Removed the sentence "In that case, the requirements.txt file can be used to install the dependencies." from the virtualenv tip as I believe it's a leftover from when you didn't want to suggest pip anymore.
- The `wkhtmltopdf` warning block didn't make sense anymore, I replaced it by a bullet point.

I believe that it's clear enough to be merged but maybe @jcs-odoo wants to do a last pass? Otherwise,
@robodoo delegate=jcs-odoo

AntoineVDV:
@d-fence I just noticed two changes that I'm not too sure about:
- You replaced `pip3` with `pip` in command blocks. Does `pip` correctly refer to `pip3` on the version of Debian/Ubuntu that we support? Because it was not the case before and users had to figure that `pip3` was the correct command to call.
- You removed `pip3 install setuptools wheel` which I had to add at some point because these packages were not pre-installed with `pip` on the brand new Linux Mint machines of our newbies. Is this no longer a problem?

d-fence:
> @d-fence I just noticed two changes that I'm not too sure about:
> 
> * You replaced `pip3` with `pip` in command blocks. Does `pip` correctly refer to `pip3` on the version of Debian/Ubuntu that we support? Because it was not the case before and users had to figure that `pip3` was the correct command to call.

AFAIK, Buster and Focal only provides `python3-pip` where pip and pip3 are the same.

> * You removed `pip3 install setuptools wheel` which I had to add at some point because these packages were not pre-installed with `pip` on the brand new Linux Mint machines of our newbies. Is this no longer a problem?

In the required libraries, we now ask to install [`python3-pip`](https://packages.ubuntu.com/focal/python3-pip) which depends and recommends all the necessary packages (`setuptools`, `build-essential` ...). And we provide a Mint based on Focal AFAIK.

jcs-odoo:
> I took the liberty to make some fine-tuning changes:
> 
> * RST cleanup.
> * Some wording fixes.
> * Removed the sentence "In that case, the requirements.txt file can be used to install the dependencies." from the virtualenv tip as I believe it's a leftover from when you didn't want to suggest pip anymore.
> * The `wkhtmltopdf` warning block didn't make sense anymore, I replaced it by a bullet point.
> 
> I believe that it's clear enough to be merged but maybe @jcs-odoo wants to do a last pass? Otherwise, @robodoo delegate=jcs-odoo

I'm adding a few small fixes then it's good to go.

content/administration/install/install.rst
```diff
@@ -505,50 +505,71 @@ create a new PostgreSQL user:
...
+   with your system. You can use virtualenv_ to create isolated Python environments.
 
-For languages with right-to-left interface (such as Arabic or Hebrew), the package `rtlcss` is
-needed:
+Other Dependencies
+^^^^^^^^^^^^^^^^^^
 
-#. Download and install **nodejs** and **npm** with your package manager.
-#. Install `rtlcss`:
+- `wkhtmltopdf` must be installed manually in version `0.12.5 <the wkhtmltopdf download page_>`_ for
```
jcs-odoo, line 561: @d-fence @AntoineVDV 

I'm putting this url https://wkhtmltopdf.org/downloads.html instead of the placeholder URL.
Or should it be  https://github.com/wkhtmltopdf/wkhtmltopdf/releases/0.12.5/   ?

AntoineVDV: The later.

jcs-odoo: Okay

jcs-odoo:
@robodoo r+

PR #1369, opened 2021-12-08
[FIX] howtos: remove useless <group/> tag from training example

luxiaochuang:
</group> tag should be deleted


AntoineVDV (approved):
Thanks for your contribution :)

@robodoo r+

PR #1397, opened 2021-12-10
[IMP] general/payment_acquirers: authorize support refund from odoo

hote-odoo:
Add 'x' to make evident that Authorize.net supports
refunds directly from odoo.

Task - 2712287
Task (on payment) - 2678757


hote-odoo:
Closed after change on payment.

PR #1421, opened 2021-12-15
[IMP] developer/*: replace occurrences of "access rule" by "record rule"

AntoineVDV:
Both the functional and technical name of the `ir.rule` model is "Record
Rule". This commit makes sure that all occurrences of "Access Rule" are
replaced by the correct name "Record Rule" as it was easily confused
with "Access Rights".

Original PR: https://github.com/odoo/documentation/pull/1118


Feyensv:
@robodoo r+

PR #1427, opened 2021-12-16
[IMP] developer/modules: deprecate 'active' and detail 'auto_install'

AntoineVDV:
1. `active` is just a deprecated field that means exactly the same as
   `auto_install`.
2. `auto_install` can be a list.

See https://github.com/odoo/odoo/blob/6cb3140ea8f62a265b76ae494295f466ca6a00c4/odoo/modules/module.py#L348-L352

Original PR: https://github.com/odoo/documentation/pull/1109

Co-authored-by: Antoine Vandevenne (anv) <anv@odoo.com>


AntoineVDV:
@robodoo r+

PR #1448, opened 2021-12-22
[IMP] `ssl on` is depricated

obayit:


jcs-odoo:
@d-fence is this correct, and if so, wouldn't we need to target 13.0 so that we can forward-port this change to all branches instead?

Thanks!

d-fence:
> @d-fence is this correct, and if so, wouldn't we need to target 13.0 so that we can forward-port this change to all branches instead?
> 
> Thanks!

Looks good to me.

Now, from the packaging point of view, it was deprecated in [nginx 15.0](http://nginx.org/en/CHANGES)
* Odoo 13.0 and Odoo 14.0 are supposed to run on Debian Buster that provides [Nginx 1.14](https://packages.debian.org/buster/nginx) 
* Odoo 15.0 on the other hand is shipping [Nginx 1.18](https://packages.debian.org/bullseye/nginx)

But as the `listen ssl` was introduced in Nginx 0.7.14, I agree that this PR should target 13.0 and be forward-ported.

jcs-odoo:
Backported this commit to 13.0 in #1612

PR #1455, opened 2021-12-29
[FIX] Update docs for static outgoing mail addresses

Yajo:
This changed for Odoo 15.0 in https://github.com/odoo/odoo/pull/61853 and the docs didn't reflect that change.

MT-52


content/developer/misc/other/cmdline.rst
```diff
@@ -259,6 +259,10 @@ Emails
 
     Email address used as <FROM> when Odoo needs to send mails
 
+.. option:: --from-filter <address or domain>
+
+    Email address or domain authorized to send outgoing mails
```
std-odoo, line 264: It's not exactly that, it's more
> Email address or domain that the mail server is configured for. If the email of the sender do not match this value, it might be encapsulated into `mail.default.from`

(if the email address do not match this, and if you have only 1 mail server, the email will be sent anyway, so the word "authorized" is not 100% correct)

content/applications/productivity/discuss/advanced/email_servers.rst
```diff
@@ -136,20 +136,34 @@ Instead, you can also configure Odoo to do something similar by itself:
...
+
+    .. image:: media/outgoing_server_from_filter.png
+       :align: center
+
+    * Use a domain (such as ``mycompany.example.com``) to keep the original "From"
+      address for mails that come from that domain.
+
+    * Use an address (such as ``outgoing@mycompany.example.com``) to allow only
+      that outgoing address.
+
```
std-odoo, line 153: maybe you can add
> Keep it empty to use this server for any email address

content/applications/productivity/discuss/advanced/email_servers.rst
```diff
@@ -136,20 +136,34 @@ Instead, you can also configure Odoo to do something similar by itself:
...
 
-    * If you want to keep the original "From" address for emails that use your
-      same domain, but change it for emails that use a different domain, use
-      the key ``mail.dynamic.smtp.from`` and set as value the email address
-      that should be used in those cases (such as
-      ``outgoing@mycompany.example.com``).
+      For example, if your mail server is configured to allow mails from 
+      ``mycompany.example.com``, and you want that any email that comes from another
+      domain gets sent "From" ``outgoing@mycompany.example.com``, you should set
+      ``mail.default.from`` to ``outgoing``.
```
std-odoo, line 166: it will not take the domain of the from-filter, but the domain in the system parameter `mail.catchall.domain` (so your example is good if `mail.catchall.domain` is `mycompany.example.com` but it must be written I think)

So the default from is `<mail.default.from>@<mail.catchall.domain>`

std-odoo (commented):
Hi :)

I added some comment, you can ping me once it's done so I can re-check

Thank you

content/applications/productivity/discuss/advanced/email_servers.rst
std-odoo: > set the email local part in ``mail.default.from``.
so this sentence is not true

std-odoo:
@Yajo It's good for me like this :)

content/applications/productivity/discuss/advanced/email_servers.rst
```diff
@@ -136,20 +136,38 @@ Instead, you can also configure Odoo to do something similar by itself:
...
+      that outgoing address.
+      
+    * Keep it empty to use this server for any email address.
+
+With this configuration in place, if Odoo sends an email that doesn't match any
+of the *from filters*, it will alter the email's "From" before sending it to
+the MTA.
+
+It will use the default outgoing email address, composed like this:
+``{mail.default.from}@{mail.catchall.domain}``.
```
AntoineVDV, line 161: Indent this block to avoid breaking the numbering.
```suggestion
    With this configuration in place, if Odoo sends an email that doesn't match any
    of the *from filters*, it will alter the email's "From" before sending it to
    the MTA.

    It will use the default outgoing email address, composed like this:
    ``{mail.default.from}@{mail.catchall.domain}``.
```

AntoineVDV:
Hi, @Yajo could you please squash your commits into a single commit with title and body:
```
[FIX] discuss, cmdline: update for static outgoing mail addresses

This changed for Odoo 15.0 in odoo/odoo#61853 and the docs didn't
reflect that change.
```
Then we're good to go :)

Yajo:
Can't you just hit "Merge and squash"? 🤔

AntoineVDV (approved):
@robodoo r+

AntoineVDV:
> Can't you just hit "Merge and squash"? thinking
> 
> done, BTW

No, because we use @robodoo's commands to merge and automatically forward-port PRs and, at the moment, we have no way to squash+rebase commits :/

PR #1490, opened 2022-01-26
[FIX] rdtraining: use "deprecate" rather than "depreciate"

jennafer-factor:
Fixed one typo - "depreciated" to "deprecated"


AntoineVDV (approved):
Hi, @jennafer-factor thanks for your contribution :)

@robodoo r+ squash

PR #1502, opened 2022-01-28
[FIX] subscription: replace an `admonition` with `example`

AntoineVDV:


AntoineVDV:
@fw-bot ignore

AntoineVDV:
@robodoo r+

PR #1528, opened 2022-02-03
[IMP] contributing/doc, install: add an extension for content tabs

AntoineVDV:
In many places in the docs, we need to document several flows that
differ only by a few details. For example, the steps to install Odoo
from the sources are almost identical for Windows, Linux, and Mac OS.
In other places, we'd want to suggest a command in different
environments or programming languages, etc.

This commit thus adds an extension for content tabs (tabs within the
content) to help reduce content repetition in the documentation. As a
proof of concept, the "Packaged installers" and "Source install"
sections of the "Installing Odoo" page are reworked to present their
content within tabs whenever it helps clarify the instructions.

A new python dependency is added as of this commit: `sphinx-tabs`

task-2713983

### DIFF
--- a/content/contributing/documentation/rst_cheat_sheet.rst
+++ b/content/contributing/documentation/rst_cheat_sheet.rst
@@ -502,6 +502,219 @@ Render
 
    Customize this admonition with a **Title** of your choice.
 
+.. _contributing/tabs:
+
+Content tabs
+============
+
+.. caution::
+   The `tabs` directive may not work well in some situations. In particular:
+
+   - The tabs' headers cannot be translated.
+   - A tab cannot contain :ref:`headings <contributing/headings>`.
+   - An :ref:`admonition <contributing/admonitions>` cannot contain tabs.
+   - A tab cannot contain :ref:`internal hyperlink targets
+     <contributing/internal-hyperlink-targets>`.
+
+Basic tabs
+----------
+
+RST
+~~~
+
+.. code-block:: rst
+
+   .. tabs::
+
+      .. tab:: Odoo Online
+
+         Content dedicated to Odoo Online users.
+
+      .. tab:: Odoo.sh
+
+         Alternative for Odoo.sh users.
+
+      .. tab:: On-premise
+
+         Third version for On-premise users.
+
+Render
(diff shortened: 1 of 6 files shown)


content/contributing/documentation/rst_cheat_sheet.rst
```diff
@@ -502,6 +502,219 @@ Render
...
+
+            The North Star.
+
+   .. tab:: Moons
+
+      .. tabs::
+
+         .. tab:: The Moon
+
+            Orbits the Earth
```
jcs-odoo, line 623:
```suggestion
            Orbits the Earth.
```

content/contributing/documentation/rst_cheat_sheet.rst
```diff
@@ -502,6 +502,219 @@ Render
...
+
+         .. tab:: Titan
+
+            Orbits Jupiter
```
jcs-odoo, line 627:
```suggestion
            Orbits Jupiter.
```

content/developer/howtos/rdtraining/02_setup.rst
```diff
@@ -184,7 +184,7 @@ You can check your Python version with:
 Install required packages
 -------------------------
 
-See :ref:`install/python-dependencies`
+See :ref:`install/dependencies`.
```
jcs-odoo, line 187: well spotted

jcs-odoo (approved):
Here is the rendering on a mobile device
image

Would there be a way to have all tabs on a same line that we can scroll? it wouldn't affect the page's width but we could "swipe" the tabs in that invisible block.

AntoineVDV:
> Would there be a way to have all tabs on a same line that we can scroll? it wouldn't affect the page's width but we could "swipe" the tabs in that invisible block.

@jcs-odoo I created a task for the design team. A proper style will come later ;)

@robodoo r+

jcs-odoo:
@AntoineVDV Any reason it's not deployed to all versions?

AntoineVDV:
> @AntoineVDV Any reason it's not deployed to all versions?

It adds a python dependency.

PR #1529, opened 2022-02-03
[ADD] developper: add gantt dependency fields

IT-Ideas:
task-2691454

see also:
- https://github.com/odoo/enterprise/pull/23370

### DIFF
--- a/content/developer/reference/backend/views.rst
+++ b/content/developer/reference/backend/views.rst
@@ -971,6 +971,15 @@ take the following attributes:
 ``date_stop`` (required)
   name of the field providing the end duration of the event for each
   record.
+``dependency_field``
+  name of the ``many2many`` field that provides the dependency relation between two records.
+  If B depends on A, ``dependency_field`` is the field that allows getting A
+  from B. Both this field and ``dependency_inverted_field`` field are used to
+  draw dependency arrows between pills and reschedule them.
+``dependency_inverted_field`` (required if ``dependency_field`` is provided)
+  name of the ``many2many`` field that provides the invert dependency relation than
+  ``dependency_field``. If B depends on A, ``dependency_inverted_field`` is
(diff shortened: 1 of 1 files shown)


Feyensv:
Back to draft until dev PR's are merged :)

IT-Ideas:
Dev PRs are merged.

content/developer/reference/backend/views.rst
```diff
@@ -971,6 +971,15 @@ take the following attributes:
...
   record.
+``dependency_field``
+  name of the field that provides the dependency relation between two records.
+  If B depends on A, ``dependency_field`` is the field that allows getting A
+  from B. Both this field and ``dependency_inverted_field`` field are used to
+  draw dependency arrows between pills and reschedule them.
+``dependency_inverted_field`` (required if ``dependency_field`` is provided)
+  name of the field that provides the invert dependency relation than
+  ``dependency_field``. If B depends on A, ``dependency_inverted_field`` is
+  the field that allows getting B from A.
```
Feyensv, line 982: Might be interesting to specify which kind of fields are supported/expected for those two attributes ;).
Is it only supported for m2m, or also x2m & m2o ?

IT-Ideas: I don't quite see any valid functional example of m2o or o2m. That would be quite odd to me...
I will add that we expect m2m, the code has been made making this assessment, thanks.

IT-Ideas: Here you go, I emphasized the many2many, is it ok for you?

Feyensv: Parent-children relationships, if we want to show the dependencies between partners, parent/children tasks, ...
It would be more a tree of dependencies, but one might think it can be done with this feature ^^.

Feyensv (approved):
robodoo r+

PR #1539, opened 2022-02-04
[FIX] accounting: fix module name to install in intrastat

jcs-odoo:


jcs-odoo:
cc @chiaraprattico

AntoineVDV:
@robodoo r+ squash

PR #1545, opened 2022-02-04
[IMP] developer: clarify options for many2one widget

jdkirkwood:
Only `no_create` is required to hide both the 'Create "xxx"' and the 'Create and Edit...' dropdown menu options.


AntoineVDV:
Hi @jdkirkwood, your PR's build on runbot is red. You may want to fix that so that someone can review your PR ;)

AntoineVDV:
Hello! Given that there has been no review in a while I will go ahead and merge as-is. Thanks for the demystification of those options by the way!

@robodoo r+

PR #1092, opened 2021-08-06
[ADD] email_communication: new  documentation regarding general emailing

Abridbus:
Creation of a new submenu under Applications/General for the Email communication.
Moved advanced of dicuss to this new submenu + adding of content regarding
frequently asked question in support.

Redirection has been done on other files pointing to Discuss/Advanced to match
the new folder email_communication.

Modified content in:
- Email servers: split and move configuration part under email_domain
- Email domain: covering information about SPF DKIM and DMARC configurations by JQU
- Email templates: (how to use the functions) by GOR
- Email common: answering frequently asked questions + common issues by ALA

PR #1547, opened 2022-02-07
[FIX] redirections: duplicate rule + misplaced rules

jcs-odoo:
- removes a duplicate rule:
  applications/finance/sign/overview/signature_validity.rst
  applications/finance/sign.rst
- moves two rules closer to their related redirections
- adds comments to these two rules
- updates redirection to "email_servers" with the new path (line 203)

This PR fixes issues in redirects.txt added with PR #1092 
cc @Abridbus


Abridbus:
Thanks for the fix :)

AntoineVDV (approved):
@robodoo delegate+

I let you call `@fw-bot up to xx.y`.

PR #1242, opened 2021-10-30
[FIX] install/maintain: update versions in "Installing Odoo"

StanvanHoorn:
Correct version from 13.0 to 15.0

PR #1559, opened 2022-02-09
Update Odoo version references on `install.rst` (13.0 -> 15.0)

AdrianTeng:
Not sure why these haven't been updated for so long -- The official website's package install instruction is still pointing to 13.0!


jcs-odoo:
Thanks for the PR @AdrianTeng !

I thought it had been updated already with #1242  but I see that the changes requested by the reviewer were not applied. I'll do it right now.

Have a great day!

Jonathan

PR #1560, opened 2022-02-09
Adds note about dev mode toggle

javl:
Adds note warning users they need at least one app installed to be able to see the 'enable dev mode' option. Took me a while to figure this out as it doesn't seem to be mentioned anywhere (in these docs or in the `Development Environment Set-up`article).

### DIFF
--- a/content/applications/general/developer_mode.rst
+++ b/content/applications/general/developer_mode.rst
@@ -11,6 +11,9 @@ Activate through the Settings
 
 Go to :menuselection:`Settings --> Activate the developer mode`.
 
+.. note::
+   You need to have at least one app installed for this menu option to appear.
+
 .. image:: settings.png
    :align: center
    :alt: Overview of the debug options under settings in Odoo


tiku-odoo:
@javl  I wasn't able to replicate this behavior in a test database. I received an error (Module 'saas_trial' cannot be uninstalled), while trying to uninstall the last app, the discuss app. What is the process to have no apps installed? Does this only occur in a subscription activated database? 

For wording can we adjust the wording to: 

Note: At least one application needs to be installed for the developer menu option to appear.

javl:
@tiku-odoo 

To replicate:

1. Create a new, bare bones, installation: `./odoo-bin --addons-path="addons" -d rd-tmp`

2. Visit `localhost:8069` in your browser. This redirects to the apps page.
3. Open the top left menu and select settings
4. You arrive at the user settings, no other settings are avaible.
5. Navigate back to the apps and install 'contacts'
6. Open the settings again, a 'general settings' page will appear showing the 'enable developer mode' option at the bottom.

(You can however enable dev mode at any time using the ctrl+k controlpanel, even without the extra app installed)

tiku-odoo:
@javl Thanks for the quick response. I notice this is for on-premise only, I tried to replicate on SAAS but I wasn't able to. Is this setting only unavailable to on-premise or local installs and not on SH/SAAS? If so, we will need to adjust the wording a bit, it should be a pretty easy change though! :) Thanks, Tim

javl:
@tiku-odoo For me this was indeed a problem on a local installation (which I think would be the first place to run Odoo for a lot of users?). I'm not sure about SAAS. I guess you'll always run Enterprise on there which might already add some extra apps? Or at least it adds the improved home screen which has a message at the top. The only SAAS installations I have access to we're set up by someone else and I can't really go around messing with them to try this ;)

I'd argue the real problem here is that settings menu needing another app to appear, which seems silly, but I assume there's some technical reason for that and just having that extra note in the text should be enough to prevent people from getting too confused (like I was).

content/applications/general/developer_mode.rst
```diff
@@ -11,6 +11,9 @@ Activate through the Settings
 
 Go to :menuselection:`Settings --> Activate the developer mode`.
 
+.. note::
+   You need to have at least one app installed for this menu option to appear.
```
tiku-odoo, line 15: Hi! Can we change this line to: 

`Note: At least one application needs to be installed for the developer menu option to appear.`

Let's remove the familiar second person language and make it less personal.

tiku-odoo (changes requested):
@javl Just a small change, see below.

content/applications/general/developer_mode.rst
tiku-odoo: @javl Are you okay with the above changes? If so, can you update the RST and we will get the doc updated :) Thanks, Tim

tiku-odoo, line 15:
```suggestion
   At least one application needs to be installed for the developer menu option to appear.
```

tiku-odoo (commented):
@odoo/doc-review I believe this employee has left Odoo. I've made a couple grammatical corrections. Thank you!

AntoineVDV (commented):
Hello @tiku-odoo you need to push your change to the PR's branch if you want to merge your version of the note. Since the `patch-2` branch was created from a fork of our documentation repository, it grants us the right to push on javl's branch directly. See bottom of the page for instructions on how to push to a different remote:
image

Alternatively, you can cherry-pick his commit in a new branch of your own and create a new PR. The authorship on the commit will be kept.

tiku-odoo:
Closing out PR #1560 as this PR was cherry-picked and made PR #2485 

CC: @StraubCreative @mivu-odoo

PR #1561, opened 2022-02-09
Add command+note on needing web_enterprise

javl:
Following this documentation I was unable to enable the enterprise version of Odoo 15.0. Seems you need to install the `web_enterprise` module for it to work, so I updated the command.

### DIFF
--- a/content/developer/howtos/rdtraining/02_setup.rst
+++ b/content/developer/howtos/rdtraining/02_setup.rst
@@ -264,8 +264,11 @@ Once all dependencies are set up, Odoo can be launched by running odoo-bin, the 
 .. code-block:: console
 
     $ cd $HOME/src/odoo/
+    $ ./odoo-bin --addons-path="addons/,../enterprise/" -d rd-demo -i web_enterprise --stop-after-init
     $ ./odoo-bin --addons-path="addons/,../enterprise/" -d rd-demo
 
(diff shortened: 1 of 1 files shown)


AntoineVDV (commented):
Hi @javl, could you double-check that your setup is correct? Indeed, you don't need to "force" the installation of `web_enterprise` as [it depends on `web`](https://github.com/odoo/enterprise/blob/15.0/web_enterprise/__manifest__.py#L14) which [depends on `base`](https://github.com/odoo/odoo/blob/15.0/addons/web/__manifest__.py#L14) which [has no dependency](https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/__manifest__.py#L13). All these modules are automatically installed once their dependencies are met.

I'm closing the PR since the suggested change is misleading, but feel free to reply should you have something to add ;)

javl:
Hi @AntoineVDV 
Apart from using a different repo for odoo-enterprise from my colleague I followed all of the steps in [this tutorial](https://www.odoo.com/documentation/15.0/developer/howtos/rdtraining/02_setup.html).
`./odoo-bin` Didn't throw any errors on `addons-path` so it was able to find the enterprise directory, it just didn't load anything until I ran it with the `-i web_enterprise` argument.

I did try with enterprise 14.0 initially, maybe that threw Odoo off. I'll try again on my other computer (which hasn't had Odoo on it yet) to make sure and report back.

PR #1566, opened 2022-02-09
[IMP] Correction on the webinar link

ren-odoo:
The current webinar link is for another version, with this commit we're adding the newest webinar link.


jcs-odoo (approved):
Good but the build seems to fail, for some reason. I'll see what I can do.

Thanks!

jcs-odoo (approved):

AntoineVDV (approved):
@robodoo r+

PR #1572, opened 2022-02-10
[REF] hooks: useAutofocus new API

Polymorphe57:


Feyensv (commented):
commit should be an IMP or FIX I suppose, and do you mind referencing the community PR causing this change in the commit message ?

@robodoo delegate+

Polymorphe57:
> commit should be an IMP or FIX I suppose, and do you mind referencing the community PR causing this change in the commit message ?
> 
> @robodoo delegate+

done

PR #1588, opened 2022-02-14
[IMP] odoo_theme: introduce fallback URLs for the page switchers

mart-e:
When a user clicks on the link of an alternate page in the version or
language switcher, we now check if the page referenced by the target URL
exists or not. If not, we generate a series of fallback URLs from the
target URL and check whether the targeted resource exists or not, until
we read the root of the documentation. As soon as we find a valid URL,
we redirect the user to it.

This is inspired by the behaviour of docs.python.org's version and
language switchers.

task-2534669


AntoineVDV (approved):
@robodoo r+

PR #1589, opened 2022-02-14
[IMP] switch admonitions' display type to inline-block

edi-odoo:
When placing an image before an admonition block and setting its alignment to left or right, the admonition block would hide the image. The reason for this is the use of "float" on the image. In order for the image to reappear, we change the admonition's `display: block` to `display: inline-block`.

task-2582954


AntoineVDV:
Updated the commit message to give it scope, description (from the PR), and task id.

AntoineVDV (approved):
Works just fine with both `:align:left` and `:align:right`.

@robodoo r+

PR #1668, opened 2022-03-08
[IMP] scss: alert and tabs refining

edi-odoo:
Before, the alerts would all have different widths according to their content, some stretching and getting pushed under the
"on this page" menu. This way all alerts are aligned.
This branch also removes unnecessary negative margins and makes code-blocks inside content tabs lose their bottom margin and border.


AntoineVDV (approved):
@robodoo r+

PR #1675, opened 2022-03-09
[FIX] developer/mixins: mail.thread method names

nle-odoo:
In odoo/odoo@d46e358d3e88aea9f9289c4845c6c87ab9491c55 some mail.thread method
names were changed: it should be changed in documentation too.

### DIFF
--- a/content/developer/reference/addons/mixins.rst
+++ b/content/developer/reference/addons/mixins.rst
@@ -342,9 +342,9 @@ ways:
   Expense Manager, etc.)
 
 These buttons settings can be applied to different groups that you can define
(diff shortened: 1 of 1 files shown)


nle-odoo:
(sorry, I had assigned antoine vdv when i checked who has merged PR before, but I saw that there was a doc-review group after)

Feyensv (approved):
@robodoo delegate+

cc: @tde-banana-odoo FYI, when the mail API is modified, the doc should be updated as well :D.

AntoineVDV (approved):
@robodoo r+

PR #1693, opened 2022-03-14
[IMP] developer/cli: mention the different ways of calling the CLI

AntoineVDV:
All examples on the page suggest calling the CLI with "odoo-bin" while
it is recommended to call it with "odoo" when Odoo was installed from a
distribution package. It also failed to mention the location of
"odoo-bin" relative to the source files.

The chance is also taken to rename the somewhat unclear page title to
something more clear and generic.

See also:
- https://github.com/odoo/odoo/pull/85878


AntoineVDV:
@robodoo r+

Feyensv (approved):

jcs-odoo:
> @robodoo r+

okay

image

PR #1723, opened 2022-03-18
Update terms_of_sale.rst

gorecrds:


odony:
"No description provided"

Anyway, there are tons of pricing details that aren't written down in contracts (starting with the prices!) and that's on purpose. Lots of things are standard practice and discussed during the sale. At that point if the customer doesn't want to accept the commercial conditions, just don't close the sale 🤷 

Adding more fine print is not gonna help. Either the customer is honest and they will understand our conditions and comply (because they're fair!), or they're cheaters and whatever you do, you're wasting your time.

gorecrds:
This has nothing to do with cheating or someone that refuses to pay. Would
you want to know that if you are late with your credit card payment there
is a penalty fee? Moreover, would you want this to be explicit, written
somewhere, and how much the fee is?

PR #1728, opened 2022-03-18
[ADD] accounting/l10n_mx: Warn not to create separate payments for invoices in USD

antoine162:
Mexican companies very commonly issue invoices in USD, to be fulfiled
in MXN, at the official exchange rate defined by the Banco de Mexico
on the day of payment.

(Presumably, this is to insure against the volatility of the Mexican
peso.)

Odoo supports this workflow, but only if the payment is registered
directly on the invoice using the 'Register Payment' button.

If the payment is created separately, and then reconciled manually
with the invoice, a whole host of problems occur:
- the payment typically can't be reconciled fully with the invoice,
  (even though that can usually be solved by manually creating
  an exchange move)
- but more problematically, the amounts on the payment CFDI will
  be wrong, and even manually creating an exchange move won't solve
  that.

So, we absolutely need to warn users not to try to do that.

(We've been encountering lots of tickets lately in the tech-support
pipe because of users who tried this and then wonder why it doesn't
work.)

This is currently an issue in 14.0, 15.0 and master.

### DIFF
--- a/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
+++ b/content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
@@ -496,6 +496,14 @@ payment, 15 days, 21 days, all falling within the current month).
 Payments
 --------
 
+.. warning::
+   - When creating a payment in MXN for an invoice in USD, the payment **must** be created using the
+     'Register Payment' button on the invoice view and not separately as a payment. Otherwise, the
+     payment CFDI will not be correctly generated.
+   - As such, a payment in MXN cannot be used to pay multiple invoices in USD. Rather, the payment
+     should be separated into multiple payments, and each payment created using the 'Register Payment'
+     button on its corresponding invoice.
+
 `According to the SAT documentation
 <https://www.sat.gob.mx/consultas/92764/comprobante-de-recepcion-de-pagos>`_, there may be 2 types
 of payments: **PUE** or **PPD**. In both cases the payment process in Odoo is the same, the


antoine162:
@smetl since you're also well aware of the technical roots of this issue, could you please just confirm for the marketing team that it indeed is an issue?

jcs-odoo:
cc @vbe-odoo .

content/applications/finance/accounting/fiscal_localizations/localizations/mexico.rst
```diff
@@ -496,6 +496,14 @@ payment, 15 days, 21 days, all falling within the current month).
 Payments
 --------
 
+.. warning::
+   - When creating a payment in MXN for an invoice in USD, the payment **must** be created using the
+     'Register Payment' button on the invoice view and not separately as a payment. Otherwise, the
+     payment CFDI will not be correctly generated.
+   - As such, a payment in MXN cannot be used to pay multiple invoices in USD. Rather, the payment
+     should be separated into multiple payments, and each payment created using the 'Register Payment'
+     button on its corresponding invoice.
```
jcs-odoo, line 505:
```suggestion
.. warning::
   - When making a payment in MXN for an invoice in USD, the payment must be created using the
     :guilabel:`Register Payment` button **on the invoice view** and not separately as a payment.
     Otherwise, the payment CFDI is not correctly generated.
   - As such, a payment in MXN cannot be used to pay multiple invoices in USD. Rather, the payment
     should be separated into multiple payments created using the :guilabel:`Register Payment`
     button on the corresponding invoices.
```

jcs-odoo (changes requested):
Thanks for this PR @antoine162 !
I hope it will help reduce the amount of errors and tickets the support gets :)

- I wouldn't put this warning at the beginning of this section as we want to explain the "normal" behaviour first.
- if you want to emphasize something with bold, better put it on relevant information rather than on "must".
- Use the present tense for normal behavior (not the future) see [guidelines](https://www.odoo.com/documentation/15.0/contributing/documentation/content_guidelines.html#grammatical-tenses)
-  `'` doesn't work to highlight a button. prefer using `:guilabel:` or double backticks  `` for "technical" display. (see suggestion)
- 100th characters max per line (see [rst guidelines](https://www.odoo.com/documentation/15.0/contributing/documentation/rst_guidelines.html#start-a-new-line-before-the-100th-character))

cc @vbe-odoo since it's your content.

vbe-odoo:
Hi, @jcs-odoo @antoine162 many thanks for the PR with the changes and the review.
LGTM in terms of the text and the description of the issue being solved here.
Thanks!
cc' @ren-odoo @fvz-odoo

jcs-odoo:
> @smetl since you're well aware of the technical roots of this issue, could you just please confirm for the marketing team that it is indeed an issue?

@antoine162 We won't check all localizations behaviors, so I assume that it was needed if you made this PR. Let me know once you've updated your commit.

Thanks :)

smetl:
@antoine162 Yeah ... I confirm I confirm!

antoine162:
@jcs-odoo all my apologies for my lateness in getting back to this PR, and thank you so much for educating me on good practices! I've changed the wording as you suggested, and put the warning at the end of the section on PUE payments, after the explanation of the normal flow. (The warning applies only to PUE payments, since these are the only payments that require a CFDI.)

AntoineVDV (approved):
@robodoo r+ squash

antoine162:
Merci beaucoup @jcs-odoo @AntoineVDV

PR #950, opened 2021-05-03
[IMP] manufacturing: add warning regarding scrap location

nouraellm:
The destination location in manufacturing order shouldn't be a scrap location. Otherwise the user won't be able to complete manufacturing orders.

PR #1736, opened 2022-03-24
[IMP] manufacturing: scrap location warning fw of #950

nouraellm:
The destination location in manufacturing order shouldn't be a scrap location.
Otherwise, the user won't be able to complete manufacturing orders.

fw of #950


AntoineVDV (approved):
@robodoo r+

AntoineVDV:
@robodoo r+

PR #1741, opened 2022-03-25
[imp] add and remove tags for position=attributes

poma-odoo:
The tags`add`, `remove` and `separator` are not included in the documentation

### DIFF
--- a/content/developer/reference/backend/views.rst
+++ b/content/developer/reference/backend/views.rst
@@ -297,6 +297,10 @@ how the matched node should be altered:
     * if the ``attribute`` element has no body, the attribute named after
       its ``name`` is removed from the matched node. If no such attribute
       exists, an error is raised
+    * if the ``attribute`` element has an ``add`` attribute, a ``remove`` attribute, or both, the
+      value of the matched node's attribute named after ``name`` is recomputed to include the
+      value(s) of ``add`` (separated by ``separator``) and delete the value(s) of ``remove``
+      (separated by ``separator``). If ``separator`` is not provided, ``,`` is used instead.
 
     .. code-block:: xml
 
@@ -305,6 +309,7 @@ how the matched node should be altered:
         <attribute name="attrs">
           {'invisible': [('sale_ok', '=', False)], 'readonly': [('editable', '=', False)]}
         </attribute>
+        <attribute name="class" add="mt-1 mb-1" remove="mt-2 mb-2" separator=" "/>
       </field>
 
 ``move``


content/developer/reference/backend/views.rst
```diff
@@ -297,16 +297,19 @@ how the matched node should be altered:
...
+      (default: ``,``) and the value given by remove is earased from attribute values
+    
     .. code-block:: xml
 
       <field name="sale_information" position="attributes">
         <attribute name="invisible">0</attribute>
         <attribute name="attrs">
           {'invisible': [('sale_ok', '=', False)], 'readonly': [('editable', '=', False)]}
+        <attribute name="class" add="mt-1" separator=" "/>
         </attribute>
```
AntoineVDV, line 311:
```suggestion
        </attribute>
        <attribute name="class" add="mt-1" separator=" "/>
```

AntoineVDV (changes requested):
Nice, I didn't know about those!

Please find below a suggestion that follows more closely the other bullet points' style and helps make the usage of `add` and `remove` clearer (IMO).
Also, the `</attribute>` closing tag is wrongly positionned.

content/developer/reference/backend/views.rst
```diff
@@ -305,8 +311,9 @@ how the matched node should be altered:
         <attribute name="attrs">
           {'invisible': [('sale_ok', '=', False)], 'readonly': [('editable', '=', False)]}
         </attribute>
+        <attribute name="class" remove="mt-2" separator=" "/>
```
AntoineVDV, line 314: I see you added this line. Is it correct to specify the `separator` attribute for `remove`? If you do need to remove `separator`, please also swap this line with the one showcasing `add` so we follow the specs order.

william-andre: You need separator for both add and remove
https://github.com/odoo/odoo/blob/be43710ab8da3be5677ae4cd3fd2e3dbe1be637b/odoo/tools/template_inheritance.py#L198-L202

But `add` and `remove` are also often used together
https://github.com/odoo/odoo/blob/74a594bb2e68662b22477fcce24d9620b2dd0b2a/addons/website/views/website_templates.xml#L136

AntoineVDV: Ok! It's indeed clearer with a bit of code :p

It appears I misinterpreted what @poma-odoo originally wrote ("the values given by `add` is added to the values separated by `separator`") as "...added to the values _**,**_ separated by `separator`". The difference is subtle but it's an entirely different story, even if `separator` is also used to append `add`'s values.

I will rephrase this and push a new version to clarify it. Let's make it crystal clear.

PR #1752, opened 2022-03-28
[FIX] upgrade: small fixes

Donapi:
task-2684744

### DIFF
--- a/extensions/odoo_theme/layout_templates/homepage.html
+++ b/extensions/odoo_theme/layout_templates/homepage.html
@@ -42,13 +42,13 @@
                         </a>
                     </li>
                     <li>
-                        <a href="{{ pathto('administration/deployment/deploy') }}" class="stretched-link">
-                            {{ _("System configuration") }}
+                        <a href="{{ pathto('administration/maintain/update') }}" class="stretched-link">
+                            {{ _("Bugfix updates") }}
                         </a>
                     </li>
                     <li>
-                        <a href="{{ pathto('administration/maintain/update') }}" class="stretched-link">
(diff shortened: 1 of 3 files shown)


AntoineVDV (approved):
@robodoo delegate=jcs-odoo

extensions/odoo_theme/layout_templates/homepage.html
```diff
@@ -42,13 +42,13 @@ <h5 class="text-muted pt-3 text-uppercase fw_semibold">{{ _("Top Links") }}</h5>
...
-                            {{ _("System configuration") }}
+                        <a href="{{ pathto('administration/maintain/update') }}" class="stretched-link">
+                            {{ _("Bugfix updates") }}
                         </a>
                     </li>
                     <li>
-                        <a href="{{ pathto('administration/maintain/update') }}" class="stretched-link">
-                            {{ _("Bugfix updates") }}
+                        <a href="{{ pathto('administration/upgrade') }}" class="stretched-link">
+                            {{ _("Upgrading Odoo") }}
```
jcs-odoo, line 51: Okay why not indeed. I agree that upgrading may be more interesting than deploying ^^

jcs-odoo (changes requested):

extensions/odoo_theme/layout_templates/homepage.html
AntoineVDV: The idea is to bring back the "maintain" section on the homepage when it is reworked to have a bit more structure and a general intro.

jcs-odoo (approved):

jcs-odoo:
@robodoo r+

PR #1771, opened 2022-03-30
[REF] general: move images in their respective page's folder

AntoineVDV:
Codeowner regexes need images to be in the correct location, otherwise some
image files might not be assigned to a GitHub team when the regex
covering the parent folder is split into multiple codeowner rules.


AntoineVDV:
@robodoo r+

PR #1795, opened 2022-04-04
Update RD training chapter 2

bso-odoo:
At the point when the developer mode is introduced in the training, it is confusing because the documentation page describes the option inside the Settings page, but that page only appears if at least one application is installed. (Otherwise, the Settings app shows the Users instead)


bso-odoo:
Doing it in a proper PR instead. #1826

PR #1797, opened 2022-04-05
[ADD] last_build: add an orphan page informing on the last build date

AntoineVDV:


content/last_build.rst
```diff
@@ -0,0 +1,7 @@
+:orphan:
+:nosearch:
+
+Last build date
+===============
+
+The last time the documentation was build in version |version| was on |today|.
```
jcs-odoo, line 7:
```suggestion
The last time the documentation was built in version |version| was on |today|.
```

jcs-odoo (changes requested):
Wouldn't there be a way to change last_build.rst to last.rst and include something like this:
the last update was on [commit merge date] with [commit].
?
@AntoineVDV

jcs-odoo (approved):

AntoineVDV:
> Wouldn't there be a way to change last_build.rst to last.rst

Last what? The page is not indexed so I'd rather have an obvious file name.

> and include something like this: the last update was on [commit merge date] with [commit].

Nope, not without an extension.

jcs-odoo:
@AntoineVDV okei alor

AntoineVDV:
@robodoo r+

PR #1804, opened 2022-04-06
[FIX] supported_versions: correct naming of recent saas~x versions

KangOl:
Since 11.0, the naming scheme changed from `(x.)saas~y` to `saas~x.y`.


AntoineVDV (approved):
@robodoo r+

PR #1816, opened 2022-04-09
[FIX] developer/rdtraining: fix typo

altela:
Fixed `name = field.Char(required=True)` to
`name = fields.Char(required=True)`.

PR #1817, opened 2022-04-09
[FIX] 15.0 Fixed field to fields in Developer Documentation

altela:
Fixed line 234

`name = field.Char(required=True)` to `name = fields.Char(required=True)`


AntoineVDV:
The fix will be handled in the automatic forward-port of https://github.com/odoo/documentation/pull/1816.

PR #1826, opened 2022-04-13
[IMP] Clarify some possible confusing points from rdtraining

bso-odoo:
This PR clarifies two points in the rdtraining:
- in chapter 2, makes it clear that the Settings page can only be accessed if at least one app is installed,
- in chapter 13, makes it clear that the task described in the transition sentence will be explained in its own section.

task-2822582


AntoineVDV (approved):
@robodoo r+ rebase-ff

AntoineVDV:
@robodoo retry

PR #1846, opened 2022-04-22
[FIX] developer: fix duplicate action availability

Bertrand2:
The duplicate action is not available for list views


AntoineVDV (approved):
@robodoo r+

PR #1868, opened 2022-04-25
[FIX] developer/rd-training: fix commands

adrienmilis:
Changed commands that were still referring to Odoo 14.0 to 15.0


AntoineVDV (approved):
@robodoo r+

PR #1902, opened 2022-05-02
[IMP] payment_ogone: add possibility to use sha256

fdamhaut:
SHA1 is going to be deprecated by ogone.
This change try to keep current behaviour while adding
the possibility to use SHA256 and SHA512

As this change as to be done in stable version,
We use the length of the key to know which version of SHA
we should use.

The goal is to switch add an selection to master to ensure
better code modularity.

The change was requested by PDE/ANV.

opw-2766648

See also:
- https://github.com/odoo/odoo/pull/88628
- https://github.com/odoo/upgrade/pull/3490

### DIFF
--- a/content/applications/finance/payment_acquirers/ogone.rst
+++ b/content/applications/finance/payment_acquirers/ogone.rst
@@ -65,6 +65,17 @@ API key as you'll not be allowed to get it later without generating a new one.
    If you are trying Ogone as a test, with the Test Account, change the **State** to *Test Mode*. We
    recommend doing this on a test Odoo database, rather than on your main database.
 
+Hash function
+~~~~~~~~~~~~~
+
+The hash function used to communicate with Ogone. It should be similar to the one set on Ogone.
+
+For new Odoo database, it's set to 'SHA512' by default which is Ogone default.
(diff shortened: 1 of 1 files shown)


AntoineVDV:
@toaa-odoo

jcs-odoo (changes requested):
Hello @fdamhaut and @toaa-odoo 

About this PR, I'd like to know what's the use of documenting something that is in developer mode? If Odoo is meant to recognize the SHA version according to the length of the key (as you explain in the commit message), then the users shouldn't be bothered with this setting.

Next, is this change applied to all versions, or only from master? It seems it's being applied to all versions from 13.0, though. If so, the base of this PR seems wrong.

It's a not a full review, but here are a few important points:
- "Odoo" and "Ogone" aren't correctly capitalized, which is grammatically incorrect.
- best to use "Odoo database" instead of "instance"
- "debug mode" is called "developer mode" and should include a ref link to the doc explaining how to activate it. (check how it's written in other user doc pages)

In Tom's PR #2117 , lines 57 to 61, Tom explains how to manually set the SHA type. But I guess all there is to do, once you've merged your PR, is to remove this explanation?

Thanks :)

cc @AntoineVDV

PR #1908, opened 2022-05-03
[FIX] developer/orm: fields_view_get is now deprecated

Feyensv:
Since odoo/odoo#87522, the method `fields_view_get` is deprecated and
`get_view` should be used and documented instead.

Community PR: https://github.com/odoo/odoo/pull/90372 + https://github.com/odoo/odoo/pull/90634

### DIFF
--- a/content/developer/reference/backend/orm.rst
+++ b/content/developer/reference/backend/orm.rst
@@ -785,6 +785,8 @@ Fields/Views
 
 .. automethod:: Model.fields_get
 
+.. automethod:: Model.get_view
+
 .. automethod:: Model.fields_view_get
 
 .. _reference/orm/domains:


Feyensv:
@robodoo r+

AntoineVDV:
Since it still very much exists, wouldn't it be better to document both methods and flag the old one as *deprecated*?

content/developer/reference/backend/orm.rst
```diff
@@ -785,7 +785,13 @@ Fields/Views
...
 .. automethod:: Model.fields_get
 
-.. automethod:: Model.fields_view_get
+.. automethod:: Model.get_view
+
+.. versionadded:: saas-15.4
+
+    :meth:`~odoo.models.Model.fields_view_get()` is now deprecated.
+    It returns the same content, but is not used anymore by the web client.
+    Please use :meth:`~odoo.models.Model.get_view()` instead.
```
Feyensv, line 794: @AntoineVDV warning or versionadded ? :D

AntoineVDV: Can't we do something [like this](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated)? We don't currently use `versionadded` because we would do it so inconsistently it's better not to mention it at all.

Feyensv: It's a bit the same display as versionadded, but the text is more adequate :+1:

AntoineVDV: Sure but it's relative to the deprecated method, which makes it easier to maintain IMO.

AntoineVDV (commented):
I'm not sure about the final look... I'd typically expect to find the _deprecated_ tag after the method name and not before, but it wouldn't be much clearer that it's related to the huge body of text above either.

Instead, could you try inserting the directive inside the docstring directly?

AntoineVDV (approved):
*Much* better! Just replace the colon with a dot in the docstring.

Feyensv:
@robodoo r+

PR #1923, opened 2022-05-06
[FIX] developer/howtos: missing parentheses in print statement

altela:
Added parentheses in a print function


Feyensv (approved):
@robodoo r+ squash

PR #1924, opened 2022-05-06
[FIX] 14.0 - Added Missing Parenthesis in Print() Function

altela:
Added parentheses in a print function


Feyensv:
Managed by #1926

PR #1932, opened 2022-05-06
[IMP] install/maintain: add seealso links in domain names doc

jcs-odoo:


jcs-odoo:
@robodoo r+

Lau-Odoo:
> pour info @Lau-Odoo

Thanks for the info!
Great idea to add the link to the tutorial!

PR #1949, opened 2022-05-09
[FIX] conf: allow building the documentation with Python 3.6.

AntoineVDV:
Commit 3d358fdb introduced a new import of Odoo source files which
complain about the minimal Python version if the hack added by 57e1e314
is not executed early enough.


mart-e (approved):

AntoineVDV:
@robodoo r+

PR #2004, opened 2022-05-16
[IMP] core: additionnal country db for geoip

Julien00859:
Task: 2848206

Related to odoo/odoo#91337


Julien00859:
We need to merge community/enterprise first before merging this one as the dependencies changed

PR #2006, opened 2022-05-17
[MOV] content/*: move resource files into their related page's directory

AntoineVDV:
Since odoo/documentation#903, the guideline for the location of new
resource (images, downloadable files, RST includes...) files is to place
those inside the directory of the RST page that references them.

For example, if `doc1.rst` has a reference to `image.png` and to
`download.zip`, the file structure should look like this:

├── parent_doc/
│     └── doc1/
│     │     └── image.png
│     │     └── download.zip
│     └── doc1.rst
│     └── doc2.rst
├── parent_doc.rst

Before this commit, most of the resource files were still located inside
'media' directories holding all the resource files referenced by RST
pages located at the same level as these directories. In the example
above, a single 'media' directory would hold all the resource files
referenced by both `doc1.rst` and `doc2.rst`. Doing so prevented us from
figuring out easily which resource file was referenced by which RST page
and, thus, lead to unused resource files piling up in the repository. It                                                                                                                                                                                            
also made it more complicated to define codeowners regex rules because a                                                                                                                                                                                            
team could not simply be assigned to `/some_page.*` but needed to be                                                                                                                                                                                                
assigned to both `/some_page\.rst` and to the location of 'media'. 

In order to help new content writers figure out the guideline when
taking examples from other RST pages, this commit retroactively applies
the guideline to existing resource files and 'media' directories. The
left-over resource files that are not referenced by any RST page are
removed.

task-2497965


jcs-odoo (commented):
Hi @AntoineVDV !

Since this PR is to adapt the media filenames according to the guidelines, may I suggest...
- replace underscores with hyphens
- remove the common prefixes in the filenames that still had a numbering instead of meaningful names (example_doc01.png --> 01.png, maybe? Not too sure, and it doesn't matter that much tbh)

AntoineVDV:
That would fall quite out of the scope of this task and would require several changes in the script that allowed to generate this diff... I leave this to content writers :sweat_smile:

jcs-odoo:
> That would fall quite out of the scope of this task and would require several changes in the script that allowed to generate this diff... I leave that to content writers 😅

No problem. Thanks already for this colossal work.
I'd say: merge it asap so we can limit the number of branches that will be in conflict ^^

Good training to learn how to fix conflicts 😈

AntoineVDV:
@robodoo r+

AntoineVDV:
@fw-bot ignore

AntoineVDV:
Manually forward-ported with:
- https://github.com/odoo/documentation/pull/2026
- https://github.com/odoo/documentation/pull/2064
- https://github.com/odoo/documentation/pull/2065
- https://github.com/odoo/documentation/pull/2066
- https://github.com/odoo/documentation/pull/2067
- https://github.com/odoo/documentation/pull/2068

PR #2026, opened 2022-05-19
[MOV] content/*: move resource files into their related page's directory

AntoineVDV:
Since odoo/documentation#903, the guideline for the location of new
resource (images, downloadable files, RST includes...) files is to place
those inside the directory of the RST page that references them.

For example, if `doc1.rst` has a reference to `image.png` and to
`download.zip`, the file structure should look like this:

├── parent_doc/
│     └── doc1/
│     │     └── image.png
│     │     └── download.zip
│     └── doc1.rst
│     └── doc2.rst
├── parent_doc.rst

Before this commit, most of the resource files were still located inside
'media' directories holding all the resource files referenced by RST
pages located at the same level as these directories. In the example
above, a single 'media' directory would hold all the resource files
referenced by both `doc1.rst` and `doc2.rst`. Doing so prevented us from
figuring out easily which resource file was referenced by which RST page
and, thus, lead to unused resource files piling up in the repository. It
also made it more complicated to define codeowners regex rules because a
team could not simply be assigned to `/some_page.*` but needed to be
assigned to both `/some_page\.rst` and to the location of 'media'.

In order to help new content writers figure out the guideline when
taking examples from other RST pages, this commit retroactively applies
the guideline to existing resource files and 'media' directories. The
left-over resource files that are not referenced by any RST page are
removed.

task-2497965


AntoineVDV:
@fw-bot ignore
@robodoo r+

AntoineVDV:
@robodoo rebase-ff

PR #2063, opened 2022-05-20
[ADD] payment_providers: explain partial capture

Demesmaeker:
Adyen now support the partial capture

task-2728768

See also:
- https://github.com/odoo/enterprise/pull/35205
- https://github.com/odoo/odoo/pull/87251

### DIFF
--- a/content/applications/finance/payment_providers.rst
+++ b/content/applications/finance/payment_providers.rst
@@ -73,7 +73,7 @@ Online payment providers
    * - :doc:`Adyen <payment_providers/adyen>`
      - Odoo
      - |V|
-     - Full only
+     - Full and partial
      - Full and partial
      -
      -
@@ -235,6 +235,10 @@ and by ticking the :guilabel:`Capture Amount Manually` checkbox.
 To capture the payment after it was authorized, go to the related sales order or invoice and click
 on the :guilabel:`CAPTURE TRANSACTION` button. To release the funds, click on the :guilabel:`VOID
 TRANSACTION` button.
+
+If your provider supports partial capture, you can select to capture only a **partial amount** of
+the full amount. The **remaining (uncaptured) amount** can either be voided or partially captured
+again later.
 
 .. note::
    - Some payment providers support capturing only part of the authorized amount. The remaining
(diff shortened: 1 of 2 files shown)


chevalierv (approved):

toaa-odoo (approved):

content/applications/finance/payment_providers.rst
```diff
@@ -236,6 +236,9 @@ To capture the payment after it was authorized, go to the related sales order or
 on the :guilabel:`CAPTURE TRANSACTION` button. To release the funds, click on the :guilabel:`VOID
 TRANSACTION` button.
 
+If your provider supports the partial capture, you may choose the amount you want to capture, and
+tick whether the remaining amount has to be voided or wait for another partial capture.
```
AntoineVDV, line 240: I'm not sure if that'd be clear for a customer. It is to me because I wrote the spec, but it sounds odd IMO.
@toaa-odoo Do you have a better alternative?

content/applications/finance/payment_providers/adyen.rst
```diff
@@ -127,12 +127,12 @@ Then, open your Adyen Merchant Account, go to :menuselection:`Account --> Settin
...
-   - Odoo doesn't support the partial capture yet. Be aware that if you make a partial capture from
-     Adyen's interface, Odoo will manage it as if it was a full capture.
 
 .. note::
-   After **7 days**, if the transaction hasn't been captured yet, the customer has the right to
-   **revoke** it.
+   - After **7 days**, if the transaction has not been captured yet, the customer has the right to
+     **revoke** it.
+   - Your Adyen account is set to only allow single partial capture by default, reach Adyen support
+     to enable multiple partial capture if needed.
```
AntoineVDV, line 135: Customers will never read this when they need to. This should rather be a configuration step appearing earlier in the page so that the account is always correctly configured.

AntoineVDV (changes requested):

content/applications/finance/payment_providers.rst
toaa-odoo: If your provider supports partial capture, you can select to capture only a **partial amount** of the full amount. The **remaining (uncaptured) amount** can either be voided or be partially captured again at a later date.

Not sure how the "wait for another partial capture" part works, but this is my suggestion based on what I understood :)

AntoineVDV (approved):
@robodoo r+

PR #2070, opened 2022-05-20
Cannot use the Odoo SMTP in combination with another outgoing mail se…

jqu-odoo:
…rver

Customer may want to use their own SMTP for email marketing, and Odoo default SMTP for everything else. There is no good way to do this. Previously we suggested adding Odoo SMTP as an outgoing email server but we don't want to share credentials anymore.
R&D is working on a solution (e.g. showing the Odoo default email server in the list rather than keeping it as an "hidden feature") but for the time being, if you use an outgoing email server, you give up using the Odoo default SMTP.


Abridbus:
Is there a reason to target 14.0 only? 
I mean we can retarget to v13 and fp to master :)

PR #1646, opened 2022-03-02
[IMP] payment_acquirers/stripe: support refund

chevalierv:
Add a check to indicate that Stripe supports refunds.

See also:
- https://github.com/odoo/odoo/pull/79271

PR #2086, opened 2022-05-25
[IMP] payment_acquirers/stripe: support refund

chevalierv:
Add a check to indicate that Stripe supports refunds.

Continuation of https://github.com/odoo/documentation/pull/1646

See also:
- https://github.com/odoo/odoo/pull/92235


AntoineVDV (approved):
@robodoo r+

PR #2167, opened 2022-06-08
[IMP] design: add an icon behind external links

edi-odoo:
The external links' target attribute is now also updated to open the
link in a new tab. In addition, the "[source]" link of autodoc's models
is moved right next to the name of the model to correctly position the
icon.

task-2790244


AntoineVDV (approved):
@robodoo r+

PR #2169, opened 2022-06-09
[FIX] requirements: pin docutils to 0.16.0

AntoineVDV:
Since sphinx-tabs 3.2.0 has the requirement docutils==0.16.0 and sphinx
3.5.4 has the requirement docutils>=0.12,<0.17, this commit pins
docutils to version 0.16.0.

While we're at it, the dependency to sphinx-tabs is also explicitly
listed, although it was already imported in the extensions. It causes
no trouble to install it from pip alongside the imported extension.


Feyensv (approved):
Tested in master as well, and It didn't change anything in my venv because I already had docutils 16 & sphinx-tabs 3.2.0

AntoineVDV:
@robodoo r+

PR #2173, opened 2022-06-09
[IMP] accounting: indonesian localization modification

toaa-odoo:
Added a warning note to indicate the " sign can mess with data.

task - 2879785


jcs-odoo (changes requested):

toaa-odoo:
Down to bugfix/R&D rather than documentation. See task id: 287978 chatter for more info.

PR #2179, opened 2022-06-10
[IMP] odoo_theme: show the "On this page" section in mobile

AntoineVDV:
When the page gets too small for the "On this page" section (local tree
of content) to fit on the page, the section is now moved above the menu
(global tree of content) rather than being hidden.

task-2800970


AntoineVDV (commented):
@robodoo r+

One issue remaining (to be fixed with task 2800937): On mobile and on pages with a large local TOC, the user needs to scroll the page to the top to be able to navigate the menu.

PR #2181, opened 2022-06-10
[IMP] rdtraining: tip about -i/-u for odoo-bin

Julien00859:
It is never shown to the trainee that he can install/update a module
right from the CLI. Multiple developpers still use the app manager to
install/update their modules multiple months into the job.


Feyensv (approved):
@robodoo r+

* backported to 14.0
* `comma serated list` --> `comma separated list`
* little correction in the indentation of previous `.. tip` command

PR #2206, opened 2022-06-14
[MERGE] odoo_theme: revamp header and menu for better mobile support

edi-odoo:
Previously the switchers were hidden on mobile. These have been moved
to a sub-nav which also contains the button for the sidenav, leaving
more space for the searchbox so it can remain in the header. Like MDN
web docs, the navigation now slides in from the left instead of the
top.

See PR's commits for more details.

task-2800937


AntoineVDV (approved):
@robodoo r+ merge

PR #2214, opened 2022-06-14
[ADD] Accounting: Carryover feature documentation added

Donapi:
task - 2523121

### DIFF
--- a/content/applications/finance/accounting/taxation/taxes/carryover.rst
+++ b/content/applications/finance/accounting/taxation/taxes/carryover.rst
@@ -1,0 +1,59 @@
+===============
+Taxes carryover
+===============
+
+Under certain legal conditions, some tax report lines must have their amount transferred to the
+following period without creating additional accounting entries. The taxes carryover functionality
+responds to this specific need.
+
+.. example::
+   In the Belgian tax report, grid 81 may contain a negative amount at the end of the tax period,
+   but it must be declared to the government as zero, and the negative amount should be carried over
+   to the next period.
+
+Configuration
+=============
+
+The carryover option can be configured for automatic use or manually for a specific line, which is
+less frequent.
+
+.. note::
+   Activate the :ref:`Developer mode <developer-mode>` to access the :guilabel:`Tax Report` from
+   :menuselection:`Configuration --> Management --> Tax Report`.
+
+Automatic use
+-------------
+
+To activate the carryover option from the Accounting application, go to
+:menuselection:`Configuration --> Tax reports`.
+
+Open a tax report, :guilabel:`Edit` and select a specific line. The following fields should be
+filled in:
+
+:guilabel:`Method`: Determines if and how the line should be carried over.
+
+:guilabel:`Persistent`: Defines how this report lines creates carry-over lines when performing tax
+closing:
+
+- If true, the amounts carried over are always added on top of each other: for example, a report
+  line with a balance of 10 with an existing carryover of 50 adds an additional 10 to it when doing
+  the closing, making a total of 60.
+- If false, the total carried-over amount is forced to the total of this report line: a report
+  line with a balance of 10 with an existing carryover of 50 creates a new carryover line of -40
+  so the total carryover becomes 10.
+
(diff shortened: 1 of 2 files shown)


content/applications/finance/accounting/taxation/taxes/carryover.rst
```diff
@@ -0,0 +1,62 @@
...
+To activate the carryover option from the Accounting application, go to
+:menuselection:`Configuration --> Tax reports`.
+
+.. note::
+   :ref:`Debug mode <developer-mode>` should be activated to access the :guilabel:`Tax Report` menu.
+
+Open a tax report, :guilabel:`Edit` and select a specific line. The following fields should be
+filled in:
+
+:guilabel:`Method`: Determines if the line should be carried over.
```
oco-odoo, line 33:
```suggestion
:guilabel:`Method`: Determines if and how the line should be carried over.
```

content/applications/finance/accounting/taxation/taxes/carryover.rst
```diff
@@ -0,0 +1,62 @@
...
+  closing, making a total of 60.
+- If false, the total carried over amount is forced to the total of this report line: a report
+  line with a balance of 10 with an existing carryover of 50 creates a new carryover line of -40
+  so the total carryover becomes 10.
+
+:guilabel:`Destination`: Defines the line to which the value of this line will be carried over to
+if needed. If left empty the line will carry over to itself.
+
+:guilabel:`Used in line balance`: If set, the carryover amount for this line will be used when
+calculating its balance in the report. This means that the carryover could affect other lines of
```
oco-odoo, line 49:
```suggestion
calculating its balance in the report. This means that the carryover could affect other lines if
```

oco-odoo (changes requested):
Very small things :)

All in all, it's really short, but I think it's best not to give too many details here, and just stick to a simple and easy-to-get explanation that will just give people what they need to try it on runbot. I'm ok with it :)

toaa-odoo (approved):

content/applications/finance/accounting/taxation/taxes/carryover.rst
```diff
@@ -0,0 +1,58 @@
...
+It responds to a specific need: under certain legal conditions, some tax report lines must have
+their amount transferred to the following period without creating additional accounting entries.
+
+.. example::
+   In the Belgian tax report, grid 81 may contain a negative amount at the end of the tax period,
+   but it must be declared to the government as zero, and the negative amount should be carried over
+   to the next period.
+
+Configuration
+=============
```
jcs-odoo, line 15: I don't understand in which cases someone would need to configure this in the hidden Config/tax reports.
cc @oco-odoo ?

We should explain the default use first, and - unless I'm mistaken - it is all pre-configured correctly according to your localization's needs?

jcs-odoo (changes requested):

content/applications/finance/accounting/taxation/taxes/carryover.rst
oco-odoo: Indeed, normally you won't have to configure it yourself as a customer. I guess the interest of a doc about it is more for whose localization does not natively support carryover in Odoo. For partners nedding it for their customers, for example. Maybe even to get an idea on how to configure it before making a PR for our Community repo. 

In our current code base, carryover is only supported in Belgium, France and Italy. I guess making it clearer for everyone might help having the other cases covered :)

Though, I don't know if it's the best time to do this, as we're currently entirely revamping Odoo's reports for v16, and the carryover is configured entirely differently in the new model (it's overall clearer than before).

Donapi: Thanks for your feedback @oco-odoo, so in the intro, we could clarify that per default carryover is supported for companies located in BE, FR, IT and that this doc is there to help for customers located in another location who might need this feature too?
As the doc is almost done, let's publish it for V15, what do you think @jcs-odoo ?

Donapi:
PR closed as we will target V16 and the content will be updated accordingly

PR #2215, opened 2022-06-14
[FIX] odoo_theme: fix highlight blocks in field-list

edi-odoo:
On pages such as /developer/reference/backend/orm.html#fields, the
highlight blocks inside `dl.field-list` items were pushing the width of
the page to go outside the screen on mobile. This is fixed by changing
the `.field-list`'s `display: grid` into `block` on mobile.


AntoineVDV (approved):
@robodoo r+

PR #2239, opened 2022-06-16
[IMP] accounting: added specification on payment flow for payment acquirer

toaa-odoo:
V13 and V14 do not allow customers with no account on Odoo.com portal to pay through payment acquirers with payment flow "Payments from Odoo"                                                                                                                                                                                                                            

task - 2711962

### DIFF
--- a/content/applications/finance/payment_acquirers.rst
+++ b/content/applications/finance/payment_acquirers.rst
@@ -169,6 +169,12 @@ authenticate the payment, or if to stay on the current page and authenticate the
 .. note::
    Some features are available only if you select *Redirection to the acquirer website*.
 
+.. warning::
+   When configuring a payment acquirer with the payment flow :guilabel:`Payment from Odoo`, the
+   payment acquirer is visible only to people who have an account and are logged in on the related
+   database. To add a payment acquirer visible to everyone, the payment flow has to be set to
+   :guilabel:`Redirection to the acquirer website`.
+
 .. _payment_acquirers/countries:
 
 Countries


content/applications/finance/payment_acquirers.rst
```diff
@@ -168,6 +168,11 @@ authenticate the payment, or if to stay on the current page and authenticate the
 
 .. note::
    Some features are available only if you select *Redirection to the acquirer website*.
+.. warning::
+   When configuring a payment acquirer with the payment flow **Payment from Odoo**, the payment
+   acquirer will not be visible to people who do not have an account on the Odoo portal. To add a
```
chiaraprattico, line 173: When configuring a payment acquirer with the payment flow **Payment from Odoo, such acquirer will not be visible to people who do not have an account on the Odoo portal.

chiaraprattico (changes requested):

log.txt
```diff
@@ -0,0 +1 @@
+2022/06/16 15:43:05 Micro started
```
AntoineVDV, line 1: This file should be deleted.

AntoineVDV (changes requested):
Please update your commit message to replace the link to the task by its id. Something like that: `task - 1234`.

FYI, there exists a script to allow clicking on task ids: https://gist.github.com/KangOl/2d1b2f710346be87d760aaa2c245914c cc @jcs-odoo

jcs-odoo:
> Please update your commit message to replace the link to the task by its id. Something like that: `[task - 1234](https://www.odoo.com/web#view_type=form&model=project.task&id=1234)`.
> 
> FYI, there exists a script to allow clicking on task ids: https://gist.github.com/KangOl/2d1b2f710346be87d760aaa2c245914c cc @jcs-odoo

Hey that's great, I didn't know!
so task-#### is the right way to do it.
Thanks

jcs-odoo (changes requested):
Hello @toaa-odoo and @chiaraprattico 

see comments in the diff.

- The commit main message must be less than 70 character
- no need to explain the exact same content than in your diff haha
- apparently, the right way to write the odoo task number is task-xxxxx

Cheers :)

toaa-odoo:
@jcs-odoo I checked with @AntoineVDV and it is an account to the Odoo.com portal, not the customer's DB

content/applications/finance/payment_acquirers.rst
```diff
@@ -169,6 +169,12 @@ authenticate the payment, or if to stay on the current page and authenticate the
 .. note::
    Some features are available only if you select *Redirection to the acquirer website*.
 
+.. warning::
+   When configuring a payment acquirer with the payment flow :guilabel:`Payment from Odoo`, the
+   payment acquirer is visible only to people who have an account on the `Odoo.com <https://www.odoo
```
AntoineVDV, line 174: Would it be clearer to simply say "users logged into their account"?

toaa-odoo: If I understood correctly, users who do not have an account on the Odoo.com portal cannot see the payment acquirer, including both non-logged users and users who do not have an account. I find "users logged into their account" implies the doc/warning only concerns users having an Odoo account not logged in, and not those who do not have an account at all.

AntoineVDV: Oh, it's actually totally unrelated to the odoo.com database! I didn't read that well. The only thing that matters is whether the user is logged in to their (internal or portal) account on the merchant's database (i.e. the one they're trying to make a payment to).

toaa-odoo: Done!

jcs-odoo (approved):

PR #2264, opened 2022-06-22
[IMP][FIX][CLN] odoo_theme: styling fixes and tweaks and general cleanup

edi-odoo:
**Special terms restyling:**
A general restyling of special terms has been made for better legibility in text, and to differentiate directives (ie: `:command:`,`:file:`,`:dfn:`) so writers can use these instead of italic and bold styling. `:guilabel:` has not been changed because it's use is too vast. 
[task-2790284](https://www.odoo.com/web/#id=2790284&menu_id=4720&cids=1&action=327&model=project.task&view_type=form)

**Reference pages have a new look:** 
`code` elements have been restyled to remove unnecessary spaces (ex: [`odoo.models` `.BaseModel` ](https://www.odoo.com/documentation/15.0/developer/reference/backend/orm.html#odoo.models.BaseModel))
and include visual aid of parent/children relations (left border).
[task-2755298](https://www.odoo.com/web#id=2755298&cids=1&menu_id=4720&action=333&active_id=3835&model=project.task&view_type=form)

**URL styling:**
When an image has an external link, the `external-link` icon appears under the image. This has been removed.
URLs with icons such as external links or "Edit on Github" links have been modified to remove the underline under the icons.
On mobile, long URLs would push the content outside the viewport. `overflow-wrap: anywhere` has been added to all URLs on mobile devices.

**No more horizontal scroll on Windows thanks to css grid:**
Windows and Mac don't work the same (no kidding). 
Windows would add ±`1rem` of horizontal scroll when Mac doesn't.
Adding that `1rem` to the content's `max-width` calculations would make Mac unhappy (it left a `1rem` gap).
The solution for everyone to be happy was to change the display type to `grid`. 
By doing this, the max-width calculations and the left positioning according to the side-nav were not necessary anymore.
Great Success. Very Nice.

**Scrolling To:**
With the new subnav, the scroll position was wrong. This is fixed by changing the padding/margin top values in the 
`o-scroll-padding` mixin.

**Replace "Hide Search Matches" with close icon:**
The title says it all.

**Forgotten scrollbar styling**
The scrollbar stylings for the side-nav should have been moved to `o_side_nav-inner` during the previous restyling of the header.

**Clean up the mess:**
All these changes made me want to clean up a bit.
By doing so I realised that a lot of the utility mixins that were used in the old *old* doc were still present here, but aren't being used. Same for some variables.
Some stying was using values instead of their variable counterparts, so these have been replaced.


AntoineVDV (approved):
@robodoo r+ merge

PR #2288, opened 2022-06-23
[FIX] update menu ref from "Master Data" to "Products"

jedgalbraith:
This updates the reference to the name of the Inventory sub menu item from "Master Data" to "Products". It changed between v14 & v15.


jirocha (approved):
Approved small edit.

PR #2324, opened 2022-06-27
[IMP] developer/views: apply the new API references style

AntoineVDV:
The documentation of view attributes was lacking visual structure to
help the reader figure out which attribute relates to which parent
element. Since this documentation was relying on RST block quotes to
represent the hierarchy of view attributes, we can apply the same new
styling that was added with commit 934f45cf and designed for autodoc,
and which relies on the exact same HTML structure as block quotes.

task-2755298

### DIFF
--- a/extensions/odoo_theme/static/style.scss
+++ b/extensions/odoo_theme/static/style.scss
@@ -666,7 +666,7 @@ header {
                 @extend code;
             }
 
-            dl.py, dl.js {
+            dl.py, dl.js, dl.o-definition-list {
                 > dd {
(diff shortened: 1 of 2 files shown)


extensions/odoo_theme/static/style.scss
```diff
@@ -666,7 +666,7 @@ header {
                 @extend code;
             }
 
-            dl.py, dl.js {
+            dl.py, dl.js, dl.o-definition-list {
```
Xavier-Do, line 669: <3

AntoineVDV:
@robodoo r+

PR #2367, opened 2022-07-04
[IMP] developer/javascript: unify the styling of API references

AntoineVDV:
Now that a new styling for API references was added with commit 934f45c,
we can remove the extra CSS that was crafted specifically for JavaScript
API references with commit 05a0a49a. While we're at it, this commit also
applies the new styling to definition lists that were missing it, like
it was done for the page on views with commit 120ae785.


ged-odoo:
looks good to me :)
robodoo r+

ged-odoo (approved):

AntoineVDV:
@robodoo r+

PR #2385, opened 2022-07-05
[IMP] accounting: updated foreign bank page v15

toaa-odoo:
Previous page was based on V13.

task - 2904025


xpl-odoo:
See PR targetting v14 to be forward ported to v15: https://github.com/odoo/documentation/pull/2383

PR #2387, opened 2022-07-06
[IMP] pos: add a warning about Six unusability in July

LoredanaLrpz:
Task ID:3835
Since Six payment terminal won't be usable starting in July 2022, I
added this warning. Also, the line lenght wasn't respected at all,
that's why they're changes in the whole doc.


jaae-odoo (changes requested):
@LoredanaLrpz  I only changed the order of some words in a sentence in the warning, up to you if you want to keep it that way or not :) For the rest, everything is A-ok

jcs-odoo (approved):
@robodoo r+

PR #2427, opened 2022-07-08
[IMP] content/developer/webservices

Julien00859:
It is hard to find and learn about RPC in Odoo, users hardly know about
JSON-RPC as the only place where we talk about it in deep down in the
*How to make a Module* tutorial.

In the commit, we isolated the tutorial about RPC in a dedicated file,
that file is better referenced thanks to cross-references in some others
places that are talking about RPC

Task: 2884559


Julien00859:
Somehow the two

    :ref:`link <howto/webservices>`

don't work... Help wanted :(

content/developer/howtos/webservices.rst
```diff
@@ -0,0 +1,150 @@
...
+            # 3.create a new session for the "Functional" course
+            course_id = call('openacademy.course', 'search', [('name','ilike','Functional')])[0]
+            session_id = call('openacademy.session', 'create', {
+                'name' : 'My session',
+                'course_id' : course_id,
+            })
+
+.. note::
+
+    There is an in-depth tutorial on XML-RPC with examples spanning over multiple programming
```
AntoineVDV, line 89:
```suggestion
    There is an in-depth tutorial on XML-RPC with examples spanning multiple programming
```
But this should rather be a `seealso` too. This is generally preferable when you mention the existence of another page that might interest the reader. Notes are more for additional comments on the content of the current page.
```rst
.. seealso::
   - :doc:`../api/external_api`: The in-depth tutorial on XML-RPC with examples spanning multiple programming languages.
```

AntoineVDV (changes requested):
Good idea :+1: 

However, could you please rebase the branch and retarget the PR onto 15.0? Since the content is the same, it would be easier for us to maintain a single version of the code spanning multiple branches than only changing the code in master. This would also benefit immediately to a lot more readers.

Julien00859:
It continues on the referenced v15 PR

PR #2438, opened 2022-07-11
[FIX] legal: fix the page layout

AntoineVDV:


content/legal.rst
```diff
@@ -22,121 +22,74 @@ Licenses
...
-    :target: legal/terms/i18n/partnership_es.html
+.. list-table::
+   :header-rows: 1
+
+   * -
+     - English
+     - Français
+     - Nederlands
+     - Deutsch
+     - Español
```
odony, line 33: the headers are aligned differently than the rows inside the table, which sometimes makes it difficult to find why language the files correspond to (in a wide browser window)

Seems like we use custom CSS[^1] for that, so it's easy to fix:
- either we remove the custom CSS and make everything left-aligned
- or we center all rows, and make `text-align: center` also apply to the TDs


[^1]: https://github.com/odoo/documentation/blob/0d474685b551f3139a1efe4d6ab096fb523a41d3/static/css/legal.css#L7-L9

extensions/odoo_theme/static/js/layout.js
```diff
@@ -8,24 +8,26 @@
...
-                         <button type="button" class="btn-close btn-close-white" data-bs-dismiss="modal" aria-label="Close"></button>
+            // Add a modal to each image that has no target.
+            if (image.parentElement.tagName !== 'A') {
+                const modalContainer = document.createElement('div');
+                modalContainer.classList.add('o_image_modal');
+                modalContainer.innerHTML = `<div class="modal fade" id="modal-${imageModalId}">
+                       <div class="modal-dialog modal-dialog-centered">
+                         <div class="modal-content">
+                           <div class="modal-body">
+                             <img src="${image.src}" alt="${image.alt}"/>
```
odony, line 19: This was already the case before, but the `alt` text is only for screenreaders. Yet in our case, it would be useful as a `title` too, because it's something everybody should see.

Could we add that too?

BTW, `img.src` and `image.alt` are data/text, you can't inject that safely inside HTML markup, they should be escaped.
Fortunately you have `$u.escape()`  ;-)

content/legal.rst
AntoineVDV: Less CSS is better CSS.

extensions/odoo_theme/static/js/layout.js
AntoineVDV: I'm not sure it would be a good idea to set the original image's `alt` as `title` of the modal image. As you said, `alt` texts are meant for screenreaders, so they're usually written as a description of what the user sees on screen (e.g., "A sales order with two order lines."). I don't think that it has any added value to show that on-screen too.

AntoineVDV:
@robodoo r+ rebase-ff

PR #2445, opened 2022-07-12
[IMP] developer: rename reference directories to "Py/Js framework"

AntoineVDV:
It was confusing to label the directory for pages related to the Python
framework "Backend" as, in Odoo, the backend is the web client, and the
frontend is the portal/website. It also led to the "Standard modules"
directory to be placed within the "Backend" directory as of `saas-15.1`
to indicate that they were part of the backend too, but that was a
mistake. Indeed, most standard modules comprise JavaScript methods, and
we could want to document these in the reference at some point.

### DIFF
--- a/content/developer/reference.rst
+++ b/content/developer/reference.rst
@@ -1,5 +1,4 @@
 :nosearch:
-:types: reference
 
 =========
 Reference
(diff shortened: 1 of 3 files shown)


AntoineVDV:
@robodoo r+

GED's review:
image

PR #2495, opened 2022-07-18
[IMP] install: troubleshoot typical problems for Mac users

bso-odoo:
As mentioned in [1], because of [2], some python implementation for Mac
fail on startup with default parameters.

Also, when installing requirements, the psycopg requirement can fail
because the $PATH does not contain the postgresql tools installed with
the .app.

This commit provides instructions to avoid those issues.

[1]: https://github.com/odoo/odoo/issues/79112
[2]: https://bugs.python.org/issue40518

task-2922249

### DIFF
--- a/content/administration/install/install.rst
+++ b/content/administration/install/install.rst
@@ -700,6 +700,11 @@ PostgreSQL
 Odoo uses PostgreSQL as database management system. Use `postgres.app <https://postgresapp.com>`_
 to download and install PostgreSQL (supported version: 10.0 and later).
 
+.. tip::
+   To make the command line tools bundled with `postgres.app` available, make sure to setup your
+   `$PATH` variable by following the `Postgres.app CLI Tools Instructions
+   <https://postgresapp.com/documentation/cli-tools.html>`_.
+
 By default, the only user is `postgres` but Odoo forbids connecting as `postgres`, so you need to
 create a new PostgreSQL user:
 
@@ -784,6 +789,11 @@ A typical way to run the server would be:
 
 Where `CommunityPath` is the path of the Odoo Community installation
 and `mydb` is the default database to serve on `localhost:8069`.
+
+.. warning::
+   Some versions of python on Mac might lead to the following error message: `current limit exceeds
+   maximum limit`. In that case, add the following parameter `--limit-memory-hard 0` to avoid the
+   problem.
 
 .. seealso::
    - :doc:`The exhaustive list of arguments for odoo-bin </developer/cli>`.


content/administration/install/install.rst
```diff
@@ -700,6 +700,10 @@ PostgreSQL
 Odoo uses PostgreSQL as database management system. Use `postgres.app <https://postgresapp.com>`_
 to download and install PostgreSQL (supported version: 10.0 and later).
 
+.. warning:: For the command line tools bundled with `postgres.app` to be available, make sure to
+          setup your `$PATH` variable by following
+          `these instructions <https://postgresapp.com/documentation/cli-tools.html>`_.
```
jcs-odoo, line 705: - it isn't a warning. Maybe a note or a tip ?
- We start the content after the admonition command (see [RST cheat sheet: tip alert block](https://www.odoo.com/documentation/15.0/contributing/documentation/rst_cheat_sheet.html#tip))
- You can go to the line anywhere there is a space, even inside the URL. (see [RST guidelines](https://www.odoo.com/documentation/15.0/contributing/documentation/rst_guidelines.html#start-a-new-line-before-the-100th-character))
- it's better to write a relevant tag for your hyperlink, rather than "this link", "here", or "these instructions

```suggestion
.. tip::
   To make the command line tools bundled with `postgres.app` available, make sure to setup your
   `$PATH` variable by following the `Postgress.app CLI Tools instructions
   <https://postgresapp.com/documentation/cli-tools.html>`_.
```

content/administration/install/install.rst
```diff
@@ -785,6 +789,10 @@ A typical way to run the server would be:
 Where `CommunityPath` is the path of the Odoo Community installation
 and `mydb` is the default database to serve on `localhost:8069`.
 
+.. warning:: Some versions of python on Mac might lead to the following error message:
+          `current limit exceeds maximum limit`. In that case, add the following
+          parameter to avoid the problem: `--limit-memory-hard 0`.
```
jcs-odoo, line 794:
```suggestion
.. warning::
   Some versions of python on Mac might lead to the following error message: `current limit exceeds
   maximum limit`. In that case, add the following parameter `--limit-memory-hard 0` to avoid the
   problem.
```

jcs-odoo (changes requested):
Hi @bso-odoo !
Thanks for the contribution to the documentation.
See my comments in the diff.

jcs-odoo (approved):

content/administration/install/install.rst
```diff
@@ -700,6 +700,11 @@ PostgreSQL
 Odoo uses PostgreSQL as database management system. Use `postgres.app <https://postgresapp.com>`_
 to download and install PostgreSQL (supported version: 10.0 and later).
 
+.. tip::
+   To make de command line tools bundled with `postgres.app` available, make sure to setup your
```
Feyensv, line 704:
```suggestion
   To make the command line tools bundled with `postgres.app` available, make sure to setup your
```
?

AntoineVDV:
@bso-odoo Any reason to target 14.0 in particular? Also, should this be forward-ported up to master or 15.4 only?

bso-odoo:
> @bso-odoo Any reason to target 14.0 in particular? Also, should this be forward-ported up to master or 15.4 only?

I targetted 14.0 because I thought nobody will install 13.0 anymore before it expires. Do you think I should target 13.0 instead ?
As far as I know these remain true for master: they are related to third party products, not to a specific version of odoo itself.

AntoineVDV:
Indeed, but this PR claims to have fixed the issue as of saas-15.5. Is the documentation change still relevant then?
Alright, let's merge from 14.0 then :)

bso-odoo:
> Indeed, but [this PR](https://github.com/odoo/odoo/pull/93381) claims to have fixed the issue as of saas-15.5. Is the documentation change still relevant then? Alright, let's merge from 14.0 then :)

Then I should remove that part in the master forward-port, but still keep the one about postgres ?

AntoineVDV (approved):
> Then I should remove that part in the master forward-port, but still keep the one about postgres ?

Sounds like a good plan to me! Note that you will need to fix a conflict in saas-15.3 because the page was heavily refactored. The per-OS sections have been replaced by content tabs.

@robodoo r+

PR #2507, opened 2022-07-20
[IMP] rdtraining: update float utils information

bso-odoo:
The warning about float utils does not state where to find them, and
they are preceded by an example in the code that infringes the warning's
recommendation.

This commit indicates where to find the float utils and replaces the
example by one that uses them.

task-2925402


AntoineVDV (approved):
@robodoo r+

PR #2514, opened 2022-07-21
[l10n_mx] Update exempt tax configuration.

fvz-odoo:
In Mexico there are two 0% taxes, 0% VAT and 0% exempt VAT. This change makes the tip clearer so users do not confuse both taxes.


AntoineVDV:
I'm closing this one since it's being done in 14.0 (and automatically forward-ported in 15.0) in https://github.com/odoo/documentation/pull/2515

PR #2516, opened 2022-07-22
[ADD] frontend: simple documentation for the editor's Powerbox

Zinston:
This adds a simple documentation for the editor's Powerbox and creates the base for the editor's own documentation in general.

Community PR: https://github.com/odoo/odoo/pull/96287
Enterprise PR: https://github.com/odoo/enterprise/pull/29649


dmo-odoo:
@robodoo r+

dmo-odoo:
💔

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+*/web_editor/tests*.
+
+Powerbox
+========
+
+The Powerbox is a piece of user interface that contains *commands* organized
+into *categories*. The commands can be filtered when the user inputs text, and
+navigated with the arrow keys.
+
+.. image:: editor/powerbox.png
```
AntoineVDV, line 29: Please compress the image with [pngquant](https://pngquant.org/).
It's not incredibly useful here, but it's a good habit to take since we'll soon test that images are indeed compressed.

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+.. highlight:: javascript
+
+.. default-domain:: js
+
+===========
+Odoo Editor
+===========
+
+Odoo Editor is Odoo's own rich text editor. It has replaced the Summernote
+library since Odoo 15.0. It can be found at
```
AntoineVDV, line 11: We generally avoid mentioning past (or current) versions of Odoo in the documentation since it should be self-sufficient in the currently browsed version.
```suggestion
Odoo Editor is Odoo's own rich text editor.
It can be found at
```

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+*web_editor/static/src/js/editor/odoo-editor*.
+
+Full documentation will come soon.
```
AntoineVDV, line 14: Better not mention that at all. Simply add the documentation when it is ready.
```suggestion
```

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+
+.. default-domain:: js
+
+===========
+Odoo Editor
+===========
+
+Odoo Editor is Odoo's own rich text editor. It has replaced the Summernote
+library since Odoo 15.0. It can be found at
+*web_editor/static/src/js/editor/odoo-editor*.
```
AntoineVDV, line 12:
```suggestion
library since Odoo 15.0. Its sources can be found in the `odoo-editor directory
<{GITHUB_PATH}/addons/web_editor/static/src/js/editor/odoo-editor>`_.
```

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Full documentation will come soon.
+
+Tests
+=====
+
+The editor comes with its own series of unit tests that can be seen at the route
+*/web_editor/tests*.
```
AntoineVDV, line 20: I'm not too sure about mentioning this. Does it add any value to the page, considering that most Odoo modules have unit tests too?

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
+
+.. highlight:: javascript
+
+.. default-domain:: js
+
```
AntoineVDV, line 5: This has no impact on the current page's content.
```suggestion
```

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+The Powerbox is a piece of user interface that contains *commands* organized
+into *categories*. The commands can be filtered when the user inputs text, and
+navigated with the arrow keys.
+
+.. image:: editor/powerbox.png
+
+Instantiation
+-------------
+
+*OdooEditor.js* instantiates the Powerbox in its own constructor and stores it
```
AntoineVDV, line 34: If it's meant as a file name, it should use the `file` role:
```suggestion
:file:`OdooEditor.js` instantiates the Powerbox in its own constructor and stores it
```

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -0,0 +1,120 @@
...
+
+There are three ways to filter commands:
+
+1. Via the `powerboxFilters` Powerbox option. It's a *function* that takes the
+   current commands and returns them filtered.
+2. Via a given command's `isDisabled` entry. It's a *function* that simply
+   returns `true` when the command should be disabled.
+3. The user can filter commands by simply typing text after opening the
+   Powerbox. It will fuzzy-match that text with the names of the categories and
+   commands.
```
AntoineVDV, line 108: Prefer [automatic numbering](https://www.odoo.com/documentation/master/contributing/documentation/rst_cheat_sheet.html#numbered-list) for numbered lists.
```suggestion
#. Via the `powerboxFilters` Powerbox option. It's a *function* that takes the
   current commands and returns them filtered.
#. Via a given command's `isDisabled` entry. It's a *function* that simply
   returns `true` when the command should be disabled.
#. The user can filter commands by simply typing text after opening the
   Powerbox. It will fuzzy-match that text with the names of the categories and
   commands.
```

AntoineVDV (changes requested):
I only posted purely technical comments in this review, but I'd like to discuss the purpose of this documentation page with you tomorrow if you don't mind.

Indeed, to me, it looks like a manifest of how you implemented the editor, and a series of pointers to interesting methods to programmatically interact with the editor. Ideally, it should rather be structured as a true reference page for the powerbox's "API".

content/developer/reference/frontend/odoo_editor.rst
Zinston: Why do you put `odoo-editor directory` in code along with the path? Also I'm guessing the trailing `_` is not intended. I had put the path between `*` because that's what I had seen in other pages of the doc. Is that not the guideline then?

content/developer/reference/frontend/odoo_editor.rst
Zinston: This is the only module that has a separate page for its unit tests. Other unit tests are at `/web/tests` but the editor has its tests at `/web_editor/tests` and (unlike all the others) doesn't use QUnit.

AntoineVDV: Oh ok, then it should probably be mentioned since it's the interesting part of the information ;)

Zinston:
Closed in order to merge separately from the Community and Enterprise PR at a later date.

PR #2533, opened 2022-07-29
update uom.rst

tonylu608:
add a conclusion to tell us the basic effect when using the units convert.


StraubCreative:
Hi there @tonylu608 👋 

We reviewed the proposed change and decided to not incorporate into this doc given the batch of changes that took place on #2556. That PR included a significant content update + restructuring to `uom.rst` which more or less covered the information you're requesting be added to the doc, in detail.

If you have further changes you'd like to be added here or elsewhere in the documentation please feel free to open up another pull request.

Thanks and KR 🙏 

cc: @mivu-odoo

PR #2542, opened 2022-08-01
[IMP] install: guide the user through the login page

AntoineVDV:
Developers sometimes have a hard time making the difference between the
PostgreSQL user and the internal user. They would try to log into Odoo's
web client with their PostgreSQL credentials rather than with admin. See
https://reddit.com/r/Odoo/comments/w8zubf/logging_in_on_source_install.


AntoineVDV:
@robodoo r+

PR #2543, opened 2022-08-01
[IMP] stripe: clarify the conditions for the support of a payment method

AntoineVDV:
Even though the documentation was correct, it was confusing the reader
to state that local payment methods are considered enabled if their
related payment icon record does not exist in the database. The intended
meaning was that this statement only holds if the local payment method
is also supported, otherwise it cannot be enabled.

Note: the content structure is partly inspired by that in 15.0+ to ease the forward-port.


dama-odoo:
Thanks for this @AntoineVDV. It's definitely clear !

toaa-odoo (approved):

AntoineVDV:
@robodoo r+

PR #2550, opened 2022-08-02
[IMP] backend: clarify and improve tour steps doc

xmo-odoo:
Does not exactly belong with odoo/odoo#96517 but I added a warning about the new requirements associated with that PR so...


Feyensv (approved):

content/developer/reference/backend/testing.rst
```diff
@@ -727,6 +790,14 @@ Set `watch` to `True` when starting a tour to automatically open a new Chrome ta
 
       self.browser_js("/web", code, watch=True)
 
+.. tip::
+
+   - to observe a step, the tour can be interrupted by setting a
+     ``run() {debugger;}`` step, or a **trigger** which never normally
+     appear and a very large ``timeout``
```
Demesmaeker, line 797: Maybe add a reference to the "Running the tour in debug mode" where the debugger was first explained? :thinking: It's clear enough as is though...

Demesmaeker (approved):
So much better... :partying_face:

content/developer/reference/backend/testing.rst
xmo-odoo: It only mentions the `debugger` stanza and I really wanted to note the trigger + timeout trick, as I don't know that everyone would think of it. 

It can be more convenient than `debugger` because under `debugger` chrome locks up the entre page (which makes sense as it can't run javascript while blocked in the javascript).

But maybe I could rename the section to something like "inspecting steps" and mention both tricks in that section? And maybe merge the debug menu thing in the "running tours in the browser" section instead?

xmo-odoo: So we'd have:

- section running tour in the browser
  - explain how to run it directly, via the debug menu or via the console
  - explain how to wartch it run via the tests
- section about introspecting steps
  - `debugger` trick
  - `timeout` trick

Demesmaeker: I think that separating it would be for the best, and far clearer than what was done before, indeed

xmo-odoo: Added a new commit for a version following that principle, feedback?

Demesmaeker: I love the advantages and drawbacks of each method!

I'd put the screenshots/screencast part above the introspecting, since it's sometime easier I think (or add a ref to it in the introspecting part)

Also, the part with the python setup in the test module (the caution point) might still be useful in the `run via browser` and/or the `introspection` part, even if it can be explained better than that. That's often something I've to say when helping people to debug testing tours with a python setup :thinking:

xmo-odoo: > I'd put the screenshots/screencast part above the introspecting, since it's sometime easier I think (or add a ref to it in the introspecting part)

Sure.

> Also, the part with the python setup in the test module (the caution point) might still be useful in the `run via browser` and/or the `introspection` part, even if it can be explained better than that. That's often something I've to say when helping people to debug testing tours with a python setup thinking

I left it as an advantage of the `watch` method and a drawback of running  via the browser, you think it should still have an additional warning box callout in addition?

Demesmaeker: Being able to introspect or to launch the tour with the browser while still having the test setup (in local at least) is a huge advantage and not as obvious as we think, so, keeping the information on how to do that might be useful. But I won't complain if it disappears either. :woman_shrugging:

xmo-odoo: I think I missed something because I'm completely lost.

- The information on how to run tours in the browser is not removed, and I don't intend to remove it.
- The information that it *can not* work when the tour *needs* Python setup is present in both "watch via browser" and "watch via `watch`" sections, it's listed as a drawback of the former and an advantage of the latter.
- The introspection part was expanded and made into its own section.

Demesmaeker: We lost the badly explained part where if we want to launch the tour with the python setup of the test in order to write/debug/whatever it, we can still do it on local by putting a breakpoint in the test, run the test, and then open the browser, where we could then have the tour with the python setup. It's really useful to check what exactly is there and to investigate what trigger should be used with the data that'll be present.
That's something I often have to show devs that first try their hands on tours, that's why I think it should be written somewhere.

xmo-odoo: > We lost the badly explained part where if we want to launch the tour with the python setup of the test in order to write/debug/whatever it, we can still do it on local by putting a breakpoint in the test, run the test, and then open the browser, where we could then have the tour with the python setup.

Oh wow, is that what it meant? I completely failed to understand what it was saying then. So you're saying to run the tour via the browser when it has python setup you put a Python breakpoint *in the python side of the test*, and when the breakpoint is hit you open a browser to the server's URL and run the tour there?

Demesmaeker: Haha, yes, indeed, that's why it was badly explained =D

xmo-odoo: OK I've tried to restore and clarify this in a note.

content/developer/reference/backend/testing.rst
```diff
@@ -681,31 +725,73 @@ To start a tour from a python test, make the class inherit from
...
+**Drawbacks**
+  - harder to use with test tours involving Python setup
+  - may not work multiple times depending on tour side-effects
 
-.. caution::
-   Be aware that when running the tour, any data added to the setup of your python test won't be
-   present in the tour unless you launched the test calling the tour with a breakpoint.
+.. note::
+
+   It's possible to use this method to observer or interact with tours
```
Demesmaeker, line 776:
```suggestion
   It's possible to use this method to observe or interact with tours
```

xmo-odoo: :fearful:

content/developer/reference/backend/testing.rst
Demesmaeker: It's crystal clear now, thank you very much :smile:

content/developer/reference/backend/testing.rst
```diff
@@ -681,31 +725,73 @@ To start a tour from a python test, make the class inherit from
...
-   {
-       trigger: "body",
-       run: () => {debugger}
-   }
+    odoo.startTour(tour_name);
+
+in the javascript console, or by enabling ``?debug=tests``, then
+selecting **Start Tour** in the debug menu and picking a tour:
+
+.. image:: testing/tours.png
```
AntoineVDV, line 763: The image is rather small in full size and is not easy to read. Would you mind taking a higher resolution screenshot and scaling it down if needed? The image modal will allow reading the details better.

content/developer/reference/backend/testing.rst
```diff
@@ -681,31 +725,73 @@ To start a tour from a python test, make the class inherit from
...
+   :align: center
+
+**Advantages**
+  - easier to run
+  - can be used on runbot
```
AntoineVDV, line 768: runbot is Odoo S.A.'s own CI tool, and it's only mentioned if very few places in the documentation. Wouldn't it be better to abstract the name away and say something like "not only in local", "on a CI tool", ...

content/developer/reference/backend/testing.rst
```diff
@@ -681,31 +725,73 @@ To start a tour from a python test, make the class inherit from
...
+  - allows running in "Onboarding" mode (manual steps)
+**Drawbacks**
+  - harder to use with test tours involving Python setup
+  - may not work multiple times depending on tour side-effects
 
-.. caution::
-   Be aware that when running the tour, any data added to the setup of your python test won't be
-   present in the tour unless you launched the test calling the tour with a breakpoint.
+.. note::
```
AntoineVDV, line 774:
```suggestion
.. tip::
```

AntoineVDV (commented):
LGTM, but I added some suggestions for you to consider.

@robodoo delegate+

By the way, have you any idea with the first bulleted list's items have more space between them than before? It seems like added sub-lists adds a bottom margin :thinking:

content/developer/reference/backend/testing.rst
```diff
@@ -639,28 +638,73 @@ Here are some example of steps:
...
-  `tripleclick`, `text Example`, `drag_and_drop selector1 selector2`...
+- **trigger**: selector/element to ``run`` an action on, the tour will
+  wait until the element exists and is visible before ``run``-ing the
+  action *on it*
+- **extra_trigger**: optional secondary condition for the step to
+  ``run``, will be waited like the **trigger** element but the action
+  will not ``run`` on the extra trigger.
+
+  Useful to have a precondition, or two different and unrelated
+  conditions.
```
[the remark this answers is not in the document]
xmo-odoo: I... didn't remove the example? They're at the top of the section, where they've always been.

Or do you mean move the examples to the corresponding step attribute?

content/developer/reference/backend/testing.rst
```diff
@@ -639,28 +638,73 @@ Here are some example of steps:
...
+  By default, tries to set the **trigger**'s content to ``Text`` if
+  it's an ``input``, otherwise ``click`` it.
+
+  The action can also be:
+
+  - a function, synchronous, executed with the trigger's ``Tip`` as
+    context (``this``) and the action helpers as parameter
+  - the name of one of the action helpers, which will be run on the
+    trigger element
+
```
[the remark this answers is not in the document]
xmo-odoo: Why is this not the default style for dls?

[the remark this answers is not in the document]
AntoineVDV: It could be if it has no side effect.

content/developer/reference/backend/testing.rst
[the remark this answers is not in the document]
AntoineVDV: I meant this sentence: "This is especially useful when the tour needs to wait for a wizard to open, a line added to a list view..." but, indeed, the information is already present in the example admonitions above. Let's keep it as it is now then.

xmo-odoo:
@robodoo r+

PR #2231, opened 2022-06-16
[ADD] accounting: electronic invoicing (edi)

Donapi:
[Task 2857860](https://www.odoo.com/web#id=2857860&cids=1&model=project.task&view_type=form)

@AntoineVDV do not forward port it in 15.0 please, changes are needed

PR #2573, opened 2022-08-08
[IMP] accounting : further info added to EDI information

Donapi:
ID: 2857860

Finetuning of #2231 for 15.0+


Feyensv (approved):
@robodoo r+

PR #2635, opened 2022-08-26
[IMP] accounting: stripe page update

toaa-odoo:
Ambiguous note which implied local payment methods were supported even if not listed.

Forward up to V14, create new for V15.

task id: 2942093


jcs-odoo (changes requested):
@toaa-odoo just two suggestions.

cheers :)

toaa-odoo:
> @toaa-odoo just two suggestions.
> 
> cheers :)

Done! Thanks ;)

PR #2645, opened 2022-08-30
[MOV] reference: test form subfields not exposed directly anymore

xmo-odoo:
todo: find a better way to show them off without providing a full path?

There is no reason for users to know their full path since there is no reason for users to instantiate them.


Feyensv:
**todo: find a better way to show them off without providing a full
path?**

Well, either do it or remove the comment :D

xmo-odoo:
> **todo: find a better way to show them off without providing a full path?**
> 
> Well, either do it or remove the comment :D

mr yvele

PR #2392, opened 2022-07-06
[IMP] payment_stripe: add information about express checkout

chevalierv:
One will now be able to pay using third parties express checkout
(e.g. Apple Pay, Google Pay) from the cart. At the moment, only Stripe
will support this flow.

See also:
- https://github.com/odoo/odoo/pull/88374
- https://github.com/odoo/enterprise/pull/29915

R&D: task-2754209

PR #2658, opened 2022-09-01
[IMP] finance/acquirers: alipay and ogone are now deprecated

Feyensv:
See also:
- https://github.com/odoo/odoo/pull/99025
- https://github.com/odoo/upgrade/pull/3848

### DIFF
--- a/content/applications/finance/payment_acquirers/alipay.rst
+++ b/content/applications/finance/payment_acquirers/alipay.rst
@@ -5,6 +5,9 @@ Alipay
 `Alipay <https://www.alipay.com/>`_ is an online payments platform established in China by Alibaba
 Group.
 
+.. warning::
+   The provider Alipay is deprecated. It is recommended to use :doc:`asiapay` instead.
+
 Configuration
 =============
 
--- a/content/applications/finance/payment_acquirers/ogone.rst
+++ b/content/applications/finance/payment_acquirers/ogone.rst
@@ -9,6 +9,9 @@ company that provides the technology involved in secure electronic transactions.
    - :ref:`payment_acquirers/add_new`
    - `Ogone's documentation <https://epayments-support.ingenico.com/get-started/>`_.
 
+.. warning::
+   The provider Ogone is deprecated. It is recommended to use :doc:`stripe` instead.
+
 Settings in Ogone
 =================
 
--- a/content/applications/finance/payment_acquirers.rst
+++ b/content/applications/finance/payment_acquirers.rst
@@ -72,11 +72,6 @@ Online payment acquirers
      - |V|
      - Full only
      - Full and partial
-   * - :doc:`Alipay <payment_acquirers/alipay>`
-     - Payment from the acquirer website
-     -
-     -
-     -
    * - :doc:`Amazon Payment Services <payment_acquirers/amazon_payment_services>`
      - Payment from the acquirer website
      -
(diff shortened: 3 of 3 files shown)


content/applications/finance/payment_acquirers/alipay.rst
```diff
@@ -5,6 +5,10 @@ Alipay
 `Alipay <https://www.alipay.com/>`_ is an online payments platform established in China by Alibaba
 Group.
 
+.. warning::
+
+    Alipay is now deprecated.  As of version 16.0, the module won't be maintained anymore.
```
Feyensv, line 10:
```suggestion
    The provider Alipay is deprecated.  It is recommended to use :doc:`asiapay` instead.
```

jcs-odoo (commented):
@Feyensv 
If they're deprecated for 16.0+, may I suggest we remove these pages, and add redirections for these pages to a note or section in the general provider doc?
What's the point of keeping pages that are not valid anymore?

Feyensv:
We're not removing the feature, only making it not available for new instances anymore.
So the doc is still useful for the clients having the app already installed.
We simply do not want to promote them anymore ;).

jcs-odoo:
> We're not removing the feature, only making it not available for new instances anymore. So the doc is still useful for the clients having the app already installed. We simply do not want to promote them anymore ;).

so these docs would be useful only for databases created prior a **date** and not really a version isn't it?
Does this mean we'd have to keep these docs forever, for all versions, though?

Feyensv:
Nope, we will probably remove the feature fully in 17 (and then from the doc as well).
It's kinda a 'tampon' period where we encourage people to migrate from the old provider(s) to the new one(s), and then we remove the feature, but still not forcing the change directly without prior notice ;).

AntoineVDV (approved):

content/applications/finance/payment_acquirers/alipay.rst
```diff
@@ -5,6 +5,10 @@ Alipay
 `Alipay <https://www.alipay.com/>`_ is an online payments platform established in China by Alibaba
 Group.
 
+.. warning::
+
+    The provider Alipay is deprecated.  It is recommended to use :doc:`asiapay` instead.
```
jcs-odoo, line 10:
```suggestion
.. warning::
   The provider Alipay is deprecated. It is recommended to use :doc:`asiapay` instead.
```

content/applications/finance/payment_acquirers/ogone.rst
```diff
@@ -9,6 +9,10 @@ company that provides the technology involved in secure electronic transactions.
    - :ref:`payment_acquirers/add_new`
    - `Ogone's documentation <https://epayments-support.ingenico.com/get-started/>`_.
 
+.. warning::
+
+    The provider Ogone is deprecated.  It is recommended to use :doc:`stripe` instead.
```
jcs-odoo, line 14:
```suggestion
.. warning::
   The provider Ogone is deprecated. It is recommended to use :doc:`stripe` instead.
```

jcs-odoo (commented):
I would understand if you hated me but I saw two spaces between `deprecated.` and `It is`.
🤣

(--> Je sors)

content/applications/finance/payment_acquirers/alipay.rst
AntoineVDV: That's actually a correct form of writing according to I-don't-remember-which-university-rulebook :p

jcs-odoo (approved):

AntoineVDV:
Closing this one temporarily to allow the merge without conflict with https://github.com/odoo/documentation/pull/2392. It'll need to be reopened and rebased later.

PR #2663, opened 2022-09-02
[IMP] razorpay: warning for void not working

Demesmaeker:
Fix payment flow + add information about the void not supported

### DIFF
--- a/content/applications/finance/payment_providers.rst
+++ b/content/applications/finance/payment_providers.rst
@@ -123,7 +123,7 @@ Online payment providers
      -
      -
    * - :doc:`Razorpay <payment_providers/razorpay>`
-     - Payment from Odoo
+     - Payment from the provider website
      -
      - Full only
      - Full and partial
--- a/content/applications/finance/payment_providers/razorpay.rst
+++ b/content/applications/finance/payment_providers/razorpay.rst
@@ -36,12 +36,11 @@ Configuration on Odoo
 #. Configure the rest of the options to your liking.
 
 .. important::
-  - If you configure Odoo to capture amounts manually, make sure to set the
-    :guilabel:`Manual Capture` to :guilabel:`manual` on Razorpay. Otherwise, the transaction
-    would be blocked in the authorized state in Odoo. To do so, go to your Razorpay Dashboard and
-    then to :menuselection:`Settings --> Configuration`.
-  - After **five days**, if the transaction hasn't been captured yet, the customer has the right
-    to **revoke** it.
+  If you configure Odoo to capture amounts manually:
+
+  - Be aware that the **manual voiding** of a transaction is not supported by Razorpay.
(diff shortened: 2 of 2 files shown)


content/applications/finance/payment_acquirers/razorpay.rst
```diff
@@ -36,10 +36,14 @@ Configuration on Odoo
...
 
 .. important::
-  - If you configure Odoo to capture amounts manually, make sure to set the
-    :guilabel:`Manual Capture` to :guilabel:`manual` on Razorpay. Otherwise, the transaction
-    would be blocked in the authorized state in Odoo. To do so, go to your Razorpay Dashboard and
-    then to :menuselection:`Settings --> Configuration`.
+  If you configure Odoo to capture amounts manually:
+
+  - Make sure to set the :guilabel:`Manual Capture` to :guilabel:`manual` on Razorpay. Otherwise,
+    the transaction would be blocked in the authorized state in Odoo. To do so, go to your Razorpay
```
toaa-odoo, line 42:
```suggestion
    the transaction would be blocked at the authorized state in Odoo. To do so, go to your Razorpay
```
I'm not sure I fully understand the flow just by reading, so this is just a suggestion. If the transaction is stopped at the step of authorizing, then I would use "at". :)

toaa-odoo (changes requested):

content/applications/finance/payment_acquirers/razorpay.rst
Demesmaeker: It's the status/state of the transaction, more than a step, so IMHO, "in" might be more adapted?

toaa-odoo (approved):

content/applications/finance/payment_acquirers/razorpay.rst
Demesmaeker: I mean, the transaction will be processed and done, but on Odoo side, the transaction will be kept in the wrong state (maybe this is clearer? :thinking: )

toaa-odoo: > I mean, the transaction will be processed and done for Razorpay, but on Odoo side, the transaction will be kept in the wrong state (maybe this is clearer? thinking ), so no payment created etc

I think this might be clearer for users, yes :)

toaa-odoo (approved):

content/applications/finance/payment_acquirers/razorpay.rst
Demesmaeker: Checking right now, this was a copy/paste from Adyen and not needed anyway for Razorpay, so I took it off

AntoineVDV (approved):
@robodoo r+

jcs-odoo:
@toaa-odoo don't forget to add the next reviewer when you approve a PR.

PR #2673, opened 2022-09-02
[IMP] account_reports: create abstract model for custom reports

epictete:
See https://github.com/odoo/odoo/pull/98973
See https://github.com/odoo/enterprise/pull/30777

The field `caret_options_initializer` is removed in the aforementioned PRs,
so it should no longer be referenced in the documentation.

task-2954761

### DIFF
--- a/content/developer/reference/standard_modules/account/account_report.rst
+++ b/content/developer/reference/standard_modules/account/account_report.rst
@@ -13,7 +13,6 @@ Report
     .. autofield:: chart_template_id
     .. autofield:: country_id
     .. autofield:: only_tax_exigible
-    .. autofield:: caret_options_initializer
     .. autofield:: availability_condition
     .. autofield:: load_more_limit
     .. autofield:: search_bar


oco-odoo (approved):

oco-odoo (approved):
@robodoo r+

william-andre (commented):
I see a link to a P&L and a Balance Sheet in the documentation, shouldn't we add at least examples to a custom report (and other reports if applicable)?

Should be a pretty easy/fast thing to do here

oco-odoo:
> I see a link to a P&L and a Balance Sheet in the documentation, shouldn't we add at least examples to a custom report (and other reports if applicable)?
> 
> Should be a pretty easy/fast thing to do here

Maybe, maybe not. I'll work on the documentation of the whole Reportalypse thing after the freeze, we'll decide then what is and what is not a "too narrow technical subject" (I'm ok explaining custom reports; but I remember having a talk with the marketing team about the use of documentation, and for them, too many technical explanations are pointless. An to be fair, I'm not sure anyone else than us writes those, unlike xml-defined reports ^^)

qdp-odoo:
@robodoo r+

qdp-odoo:
@william-andre do you have r+ rights here?

william-andre:
Nope, only @odoo/doc-review has it

william-andre (approved):

jcs-odoo (approved):
@robodoo r+

PR #2674, opened 2022-09-05
[IMP] legal: replace txt icon with html icon

jcs-odoo:
The "txt" icon is misleading as it logically implies that clicking on it
would open a .txt file, while it actually redirects to an HTML page.

task-2972827


jcs-odoo:
@AntoineVDV Do we need approval from legal for such changes?

AntoineVDV:
@jcs-odoo I'd say no because you just improve the look of the page. The codeowner is there to prevent people from changing the terms. If it turns out the .txt icon was actually important, we can still revert the change.

odony:
Really? I wonder who, in the audience for our legal documents, cares about the difference, or even knows what "HTML" is supposed to be?

So with "HTML" they should logically expect to download a .html version of the doc?

odony:
I wasn't very clear: feel free to proceed with this change if you think it's important ;-)

Though personally I don't think it fixes the confusion. If visitors are surprised by the format, we should perhaps avoid mentioning any extension? Except for the PDF file of course, as most people look for a "PDF icon" for download.

Here's an example based on the `fa-file-lines` and `fa-file-pdf` fontawesome icons - but any similar icon would work:

Legal-—-Odoo-15-0-documentation

PR #2752, opened 2022-09-22
[FEEDBACK]

ray-odoo:
Can we discuss the Documents App here, it has several Budget Templates that make it much easier to work with the results than our other methods

P.S. This is not a real PR, I have done absolutely none of the work needed

image

image

image


jcs-odoo:
Hi @ray-odoo ,

You can send your feedback to doc@mail.odoo.com (or create a task directly in the project Documentation). 
External users can also provide feedback here on GitHub through the **Issues** tab. But let's leave the **Pull Requests** tab for the pull requests :)

Actually, I think @Donapi wanted to start working on that but we got caught by all the OXP stuff to produce.
[Here is the task](https://www.odoo.com/web#id=2899532&menu_id=4720&cids=1&action=333&active_id=3835&model=project.task&view_type=form).

I added your feedback there.

Have a good day :)

PR #2782, opened 2022-09-30
[ADD] reference/standard_modules: API reference for the payment engine

AntoineVDV:
task-2804999

See also:
- https://github.com/odoo/odoo/pull/101756


AntoineVDV:
@robodoo r+

PR #2785, opened 2022-10-04
[FIX] name field of ir.ui.view is optional

legalsylvain:
The documentation says that the field  ``name`` is required. 
However, since (at least) version 8, the field is optional, and will be autogenerated if not defined.

### DIFF
--- a/content/developer/reference/backend/views.rst
+++ b/content/developer/reference/backend/views.rst
@@ -37,9 +37,10 @@ Fields
 
 View objects expose a number of fields. They are optional unless specified otherwise.
 
-* ``name`` (mandatory) :class:`~odoo.fields.Char`
+* ``name``  :class:`~odoo.fields.Char`
 
   Only useful as a mnemonic/description of the view when looking for one in a list of some sort.
+  If the field is not defined, it will be generated automatically like this "MODEL TYPE".
 
 * ``model`` :class:`~odoo.fields.Char`
 


AntoineVDV:
Toggling the base branch to force it on runbot.

legalsylvain:
Hi  @AntoineVDV. Thanks for taking into account my PR. 
I don't understand why CI is broken. 
- ci/codeowner  : ``` Some pr have an invalid target: 22935,8305 ```
- ci/documentation  : ``` No "build succeeded." found in logs. ````
do you know why ? 
thanks !

AntoineVDV:
Hi @legalsylvain, it's because runbot tries to bundle this PR with https://github.com/odoo/odoo/pull/22935 as both have the same `patch-1` branch. I don't believe that there is a way to unbundle them from runbot so the best fix would be to close this PR and open a new one with a different branch. Pick a name that will not collide with another of your odoo/* branches; something like `15.0-view-optional`.

legalsylvain:
> Hi @legalsylvain, it's because runbot tries to bundle this PR with https://github.com/odoo/odoo/pull/22935 as both have the same patch-1 branch. I don't believe that there is a way to unbundle them from runbot so the best fix would be to close this PR and open a new one with a different branch. Pick a name that will not collide with another of your odoo/* branches; something like 15.0-view-optional.

thanks ! done here https://github.com/odoo/documentation/pull/3295

PR #2800, opened 2022-10-10
[IMP] orm/changelog: store translations in jsonb

HydrionBurst:


Feyensv:
Should target branch 16.0 instead.

PR #2840, opened 2022-10-21
[FIX] install: update installation from the nightly debian repo

d-fence:
As the `apt-key` command will is deprecated and will be removed in the next Debian release, it's time to update the documentation.


jcs-odoo (approved):

AntoineVDV (approved):
@robodoo r+

d-fence:
Thank you

PR #2848, opened 2022-10-25
[FIX] install: update 15.0 references to 16.0

d-fence:


content/administration/install/install.rst
```diff
@@ -183,7 +183,7 @@ Distribution package
       <download_>`_.
 
       .. note::
-         Odoo 15.0 'deb' package currently supports `Debian 11 (Bullseye)`_, `Ubuntu 20.04 (Focal)`_
+         Odoo 16.0 'deb' package currently supports `Debian 11 (Bullseye)`_, `Ubuntu 22.04 (Jammy)`_
```
Feyensv, line 186: `Ubuntu 20.04 (Focal)`_ was a link reference, which is why the doc build fails when trying to find `Ubuntu 22.04 (Jammy)`_ reference.

You should replace the unused Focal link and replace it with a new Jammy link ;).

jcs-odoo (approved):
@d-fence I fixed the commit as suggested by @Feyensv

Feyensv (approved):
@robodoo r+

PR #2882, opened 2022-10-28
14.0 managevendorbills toaa

toaa-odoo:


toaa-odoo:
Mistakenly built on 16.0

PR #2910, opened 2022-11-04
[IMP] developer/api/external_api: update for /RPC2

Julien00859:
All the code snippets have been updated to use the new XML-RPC API at the /RPC2 endpoint.

Co-authored-by: Julien Castiaux <juc@odoo.com>


AntoineVDV:
@Julien00859 The CI is red. It looks like you used a 15.0 or 16.0 base.

Also, is it possible to target a lower version than master? As much as possible, you should target the first relevant version of the documentation so we don't have to maintain dozens of variants of a given paragraph.

PR #2927, opened 2022-11-09
[FIX] deploy: wsgi application entrypoint moved

Julien00859:
The wsgi application entrypoint moved during the httpocalypse. Some clients don't use the odoo builtin wsgi server and have troubles upgrading from 15.0 to 16.0 because the `odoo.service.wsgi_server` module doesn't exist anymore.

https://github.com/odoo/odoo/pull/105434

### DIFF
--- a/content/administration/install/deploy.rst
+++ b/content/administration/install/deploy.rst
@@ -352,6 +352,11 @@ However the WSGI server will only expose the main HTTP endpoint for the web
 client, website and webservice API. Because Odoo does not control the creation
 of workers anymore it can not setup cron or livechat workers
 
+.. versionchanged:: 16.0
+
+   The application entry-point has been moved from
+   ``odoo.service.wsgi_server.application`` to ``odoo.http.root``.
+
 Cron Workers
 ------------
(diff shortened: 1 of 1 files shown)


Gorash (approved):

Julien00859:
@robodoo r+

Feyensv (approved):
@robodoo r+

AntoineVDV:
@robodoo r-

AntoineVDV (changes requested):
Hello, I really don't think this is the place to mention a path change that is not referenced in the paragraphs above. I feel like this is a bandage for a missing deprecation warning, an upgrade changelog, or similar.

The user documentation should neither be formatted as a changelog nor dispense upgrade advises. Instead, we should rather keep it factual and up-to-date: on a given version (here 15.3), this deployment guide should give the required steps, no more no less.

Julien00859:
We restored the `wsgi_server` module with its `application` function with a deprecation warning. The deprecation warning replaces the `versionchanged` of this PR

PR #2977, opened 2022-11-17
[REF] dev_backend_reports: assets and less

switch87:
Before this commit the section "Custom fonts" was outdated with adding assets in a view and use of less

After this commit the assets are added in the manifest file and the stylesheet is of filetype scss


switch87:
@toaa-odoo @mart-e

kmagusiak:
Could you rebase and solve conflicts?
Otherwise the PR looks good.

PR #3008, opened 2022-11-22
[IMP] cli: suggest auto complete script

william-andre:
Add a hint to how to build autocompletion values for odoo-bin

### DIFF
--- a/content/developer/cli.rst
+++ b/content/developer/cli.rst
@@ -51,6 +51,14 @@ Help & version
 
    shows Odoo version e.g. "Odoo Server 13.0"
 
+.. tip:: You can enable auto-completion in your shell by running
+
+  .. code-block:: bash
+
+    echo "complete -W '`./odoo-bin --help | \
+      sed -e 's/[^a-z_-]\(-\+[a-z0-9_-]\+\)/\n\1\n/' | \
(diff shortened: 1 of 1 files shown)


william-andre:
See discussion on https://github.com/odoo/odoo/pull/64541

content/developer/cli.rst
```diff
@@ -51,6 +51,14 @@ Help & version
 
    shows Odoo version e.g. "Odoo Server 13.0"
 
+.. note:: You can enable auto-completion in your shell by running
```
Feyensv, line 54:
```suggestion
.. tip:: You can enable auto-completion in your shell by running
```

?

Feyensv (commented):
Interesting and useful information :+1:
Maybe worth adding in 14 + as well ?

AntoineVDV:
@william-andre That actually depends on the page. For this one, 16.0 is good enough because it's very unlikely that we update the "Help" section in an earlier version. For other pages, like Installing Odoo, it's not a good idea to target 16.0 because any (required) change in 14.0/15.0 will conflict at the forward-port.

AntoineVDV (approved):
@robodoo r+

Feyensv (approved):
@robodoo r+

PR #3017, opened 2022-11-23
[ADD] Legal: Translation in Simplified Chinese of Subscription Agreement

jopa-odoo:


AntoineVDV:
@robodoo delegate=odony

AntoineVDV:
@jopa-odoo The `ci_documentation_guidelines` check is red because the length of the headings doesn't match that of the headings' underlines.

Ideally, it should look like this:
```rst
Some heading of 29 characters
-----------------------------
```

Now, it's not a big deal. So if Olivier reviews before you have a chance to fix the headings, the PR can be merged ignore the check.

jopa-odoo:
Hi @AntoineVDV, I've talked about this with JCS who told me to ignore these warnings for these languages... 
But if you want me to fix the headings, no problem :)

AntoineVDV:
It's indeed not important for legal pages because we never update those. Now, if Olivier asks you for other changes, you could fix the check at the same time. Otherwise, it'd fine as it is.

PR #3036, opened 2022-11-25
[IMP] contributing/documentation: suggest to clone with ssh

AntoineVDV:
GitHub requires users to "use token-based authentication for all authenticated operations", for example `git push`. See https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/

This commit suggests new contributors to clone the repository with SSH so that the remote URL is automatically configured with the SSH protocol.

task-3036845


AntoineVDV:
@robodoo r+

PR #3044, opened 2022-11-28
[IMP] web: adds the header to the list view

FrancoisGe:
This commit adds the documentation related to the <header> tag to the list view.

### DIFF
--- a/content/developer/reference/backend/views.rst
+++ b/content/developer/reference/backend/views.rst
@@ -1918,6 +1918,8 @@ Possible children elements of the list view are:
             matching the current search)
     ``name``
         see ``type``
+    ``groups``
+        lists the groups which should be able to see the button
     ``args``
         see ``type``
     ``attrs``
@@ -2048,6 +2050,31 @@ Possible children elements of the list view are:
     </groupby>
 
   A special button (`type="edit"`) can be defined to open the many2one form view.
+
+``header``
+  defines custom :ref:`buttons <reference/views/list/button>` in the control panel that perform an action/call a model's method.
+
+  .. code-block:: xml
+
+      <header>
+          <button name=toDo" type="object" string="Always Display" display="always"/>
+          <button name="toDo" type="object" string="Selection Display"/>
+      </header>
+
+  Does not support any attribute but can have a child:
+
+  .. rst-class:: o-definition-list
+
+  ``button``
+    as a :ref:`button <reference/views/list/button>` which accepts an extra attribute when placed in a `header`:
+
+    .. rst-class:: o-definition-list
+
+    ``display``
+      By default, those buttons are only displayed when some records are
+      selected, and they apply on the selection. When the attribute ``display``
+      is set to ``always``, the button is available all the time, even if there's
+      no selection.
 
 ``control``
   defines custom controls for the current view.


content/developer/reference/backend/views.rst
```diff
@@ -1863,6 +1863,81 @@ Possible children elements of the list view are:
...
+            (``active_ids``, may be just a subset of the database records
+            matching the current search)
+
+    ``display``
+      attribute allowing to know when the button should be displayed.
+      The default value is  ``selection``.
+        ``always``
+          the button is always displayed.
+        ``selection``
+          the button is only displayed if at least one record is selected.
```
aab-odoo, line 1910: By default, those buttons are only displayed when some records are selected, and they apply on the selection. When the attribute `display` is set to `always`, the button is available all the time, even if there's no selection.

aab-odoo (approved):

aab-odoo (approved):
robodoo r+

content/developer/reference/backend/views.rst
```diff
@@ -1931,6 +1933,25 @@ Possible children elements of the list view are:
...
+      </header>
+
+  ``button``
+    as :ref:`button <reference/views/list/button>` with:
+
+    ``display``
+      By default, those buttons are only displayed when some records are
+      selected, and they apply on the selection. When the attribute ``display``
+      is set to ``always``, the button is available all the time, even if there's
+      no selection.
```
AntoineVDV, line 1953: RST is very sensitive to indentation. Even 1 extra space creates a description list in HTML, which is probably not intended here.

AntoineVDV (commented):
The Views page is currently constructed to list allowed attributes rather than inner elements. I feel that listing `button > display` would create confusion.

Instead, what do you think of listing `display` as a conditionally allowed attribute to the definition of `button` in `reference/views/list/button`, and mentioning that extra attribute in `header`?

Something like this:
```rst
.. _reference/views/list/button:

.. rst-class:: o-definition-list

``button``
    displays a button in a list cell

   .. rst-class:: o-definition-list

    ``icon``
        icon to use to display the button

    [...]

    ``display``
        By default, buttons are only displayed...
        
[...]

``header``
  defines custom :ref:`buttons <reference/views/list/button>` in the control panel that perform an action/call a model's method.
  When used within a header, buttons accept the extra attribute :ref:`display <...>`.

  .. code-block:: xml
      <header>
          <button name=toDo" type="object" string="Always Display" display="always"/>
          <button name="toDo" type="object" string="Selection Display"/>
      </header>
```

content/developer/reference/backend/views.rst
FrancoisGe: We discussed it with aab and we prefer our solution. We find it more clear.

FrancoisGe: > RST is very sensitive to indentation. Even 1 extra space creates a description list in HTML, which is probably not intended here.

Are you sure for this fix ?

AntoineVDV: I'm not sure to understand the question. Am I sure that this is the RST syntax for a description list? Then yes^^

What are you trying to achieve here with that extra space? I can't tell if it's a mistake or an attempt at reproducing a structure you saw somewhere else?

FrancoisGe: I do the same as in the rest of the code. I'm not sure I understand.
```
``header``
  defines custom :ref:`buttons <reference/views/list/button>` in the control panel that perform an action/call a model's method.
  When used within a header, buttons accept the extra attribute :ref:`display <...>`.
```

AntoineVDV: Ok, I see! Then you must insert `.. rst-class:: o-definition-list` before the code block on line 1939 and before \``display`` on line 1949. This applies the appropriate styling on description lists, which are used as a hack to list attributes.

AntoineVDV, line 1953: I find it confusing and disruptive (compared to the rest of the page) to duplicate *part of* the reference of `button` inside that of `header` *without* mentioning the existence of the `display` attribute in the original definition of `button`, which is where readers will look for an explanation on `display`.

Now, if you're convinced that it is clearer, please consider adding more text to make it a bit easier to understand:
```suggestion
``header``
  defines custom :ref:`buttons <reference/views/list/button>` in the control panel that perform an action/call a model's method.

  .. code-block:: xml

      <header>
          <button name=toDo" type="object" string="Always Display" display="always"/>
          <button name="toDo" type="object" string="Selection Display"/>
      </header>
   
  Does not support any attribute but can have a child:             

  .. rst-class:: o-definition-list

  ``button``
    as a :ref:`button <reference/views/list/button>` which accepts an extra attribute when placed in a `header`:
    
    .. rst-class:: o-definition-list

    ``display``
      By default, those buttons are only displayed when some records are
      selected, and they apply on the selection. When the attribute ``display``
      is set to ``always``, the button is available all the time, even if there's
      no selection.
```

AntoineVDV (commented):
@robodoo delegate+

AntoineVDV:
@FrancoisGe In case you didn't see it, you can r+ this PR when you feel it's ready ;)

FrancoisGe:
> @FrancoisGe In case you didn't see it, you can r+ this PR when you feel it's ready ;)

Ok thx :)

aab-odoo:
robodoo r+

FrancoisGe:
@AntoineVDV Hi, I haven't changed anything since your last validation and the runbot is red. Do you know what I should do?

AntoineVDV:
New tests have been added to runbot in the meantime. It should go through now.

@robodoo r+

PR #3059, opened 2022-12-01
[IMP] pos: update tip info, remove and update screenshots

LoredanaLrpz:
TaskID: 3005150


content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -2,30 +2,35 @@
...
+of Sale`, select the POS, scroll down to the :guilabel:`Payment` section and enable
+:guilabel:`Tips`. Once enabled, add a :guilabel:`Tip Product` in the corresponding field, and save.
+The referred product will be used as a reference on customer receipts.
+
+.. image:: tips/tips-setup.png
+   :align: center
+   :alt: enable tips in a POS
+
+.. note::
+   - You can create a product on the spot. To do so, enter your product's name and click
```
toaa-odoo, line 21:
```suggestion
   - You can create a product on the spot. To do so, enter your product's name and click
```
I guess this is because you have to create a product "tip" to use in the Tip Product field, but this not explained anywhere, nor is it explained how to configure the product in the product template. IMO, this should either be explained or :ref:/:doc: if it is somewhere in the documentation. Also, it should be in the main text and not as a note (perhaps in a H3 heading, as suggested above). If it does come as part of the POS module (as it is available in a runbot under [TIPS] Tips), then it should be documented people can search for that pre-made product (and that it comes by default).

Also, I can see from the screenshot: two options Add tip after payment (NA only), and add tip through payment terminal (Adyen). Neither are explained either, and are only notes (which IMO should be in the main text as well). Also, why only Adyen ? I tried in a runbot and only have Adyen too, if it's only doable in Adyen, it should be documented as it doesn't say otherwise, nor says how to do it otherwise (if possible).

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -2,30 +2,35 @@
...
+     :guilabel:`Create`. The product is automatically configured, but you can click
+     :guilabel:`Create & edit...` to open the product creation form and set it up manually.
+   - If you are using a payment terminal and wish to enable **tips** using the terminal, check
+     :guilabel:`Add tip through payment terminal`.
+
+Add tips
+========
+
+Once an order is done, click :guilabel:`Payment` to access the payment screen. Click
```
toaa-odoo, line 30:
```suggestion
Upon payment of the order/When registering the payment in your POS, click :guilabel:`Payment` to access the payment screen. Click
```
Once an order is done (to me) makes it sound as if it is *after* the payment is cashed-in

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -2,30 +2,35 @@
...
+:guilabel:`♥ Tip`, enter the amount and click :guilabel:`Confirm` to validate. Then, process to the
+payment.
```
toaa-odoo, line 32:
```suggestion
payment. Tips can be found on the receipt at the :guilabel:`Tips` line.
```
loredana reveiw

toaa-odoo: Picture only for your illustration, I'm not suggesting you should add it to the doc

toaa-odoo (changes requested):
Hi @LoredanaLrpz, I've made (rather) substantial suggestions. Since I've reviewed 2-3 PoS PRs recently, I would suggest keeping in mind when writing documentation that it should always be assumed the user doesn't know anything, and therefore everything should be explained to them. From the step where they open the app to the feature they want to enable, each and every single step should be explained. Also, other options related to what is documented are not documented (in this case, the north America)

content/applications/sales/point_of_sale/restaurant/tips.rst
LoredanaLrpz: You don't have to "create" a product, but if you do, as written, you can just "type the product's name and click create," which creates a product and set it up completely.
However, since it appears that part was unclear, I've added the h3 title about the product creation :)

Only Adyen because it only works for Adyen 
I don't think it should be in the main text; these are just options to activate; there's not enough content for a whole part in the main text; I've let them in the note
However, I added some info about the tip after the payment

content/applications/sales/point_of_sale/restaurant/tips.rst
LoredanaLrpz: it was already written at line 14: The referred product will be used as a reference on customer receipts.
I ignored that :)

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,75 @@
...
+:guilabel:`Consumable` to avoid unnecessary inventory movements.
 
-.. image:: tips/tips03.png
-    :align: center
+.. future: remove 1st sentence and add
+   :ref:`create a product <product_creation>`
+
+You can also create a product on-the-spot. To do so, enter a product's name in the
+:ref:`Tip Product <configuration>` field and click :guilabel:`Create`. The product is automatically
+configured to be used as a tip, but you can click :guilabel:`Create & edit...` to open the product
```
toaa-odoo, line 48:
```suggestion
configured to be used as a tip, but you can click :guilabel:`Create and edit...` to open the product
```

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,75 @@
...
+     :guilabel:`Save & Close`.
+   - You can only select one tip product per POS, but you can choose a different tip product for
+     each POS.
+
+Add tips
+========
+
+To add a tip to your order, you can add the **tip product** to the cart or click :guilabel:`Payment`
+to access the payment screen. Then, click :guilabel:`♥ Tip`, enter the amount and click
+:guilabel:`Confirm` to validate. Then, process to the payment.
```
toaa-odoo, line 66:
```suggestion
:guilabel:`Confirm` to validate. Then, process to the payment.
```
"Then" repetition

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,75 @@
...
+You can also create a product on-the-spot. To do so, enter a product's name in the
+:ref:`Tip Product <configuration>` field and click :guilabel:`Create`. The product is automatically
+configured to be used as a tip, but you can click :guilabel:`Create & edit...` to open the product
+creation form and set it up manually.
+
+.. note::
+   - If you use an :doc:`Adyen <../payment/adyen>` payment terminal and wish to enable **tips**
+     using the terminal, check :guilabel:`Add tip through payment terminal (Adyen)`.
+   - If you create a tip product on-the-spot, the **Available in POS** feature is not activated.
+     Thus, the product won't appear in a POS session. If you want to select the product in POS,
```
toaa-odoo, line 55:
```suggestion
     Thus, the product will not appear in a POS session. If you want to select the product in POS,
```
or "cannot"

toaa-odoo (changes requested):

jcs-odoo (changes requested):
review done IRL

LoredanaLrpz:
Hello @jcs-odoo , I've done many changes, tell me what you think about it now :)

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,88 @@
...
+**Tip products** can be created on the spot. To do so, enter a product's name in the :ref:`Tip
+Product <configuration>` field and click :guilabel:`Create` or press **enter**. The product is
+automatically configured to be used as a tip at the payment screen.
 
-Add the tip your customer wants to leave and process to the payment.
+However, if you wish to be able to select the tip product in a POS session, you must activate the
+**Available in POS** setting. To do so, click :guilabel:`Create and edit...` instead of
+:guilabel:`Create` after you enter the product's name to open the product configuration form. Then,
+tick the :guilabel:`Available in POS` checkbox in the :guilabel:`Sales` tab and click
+:guilabel:`Save & Close`.
```
jcs-odoo, line 36: I suggest you directly explain where to go, rather than saying "click create and edit instead of create after you...."

--> To do so, go to the Sales tab of the tip product and tick....

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,88 @@
...
+
+If you use a POS system in a bar or a restaurant, you can enable :guilabel:`Add tip after payment
+(North America specific)`. Doing so generates a bill to print and complete manually by the customer
+and the waiter. That bill indicates the tip value the customer chooses to give after the payment.
+
+At checkout, select the payment method and click :guilabel:`Close Tab`. If no tip was added before
+the payment, the bills get generated. Print and complete them.
+
+.. important::
+   To use this feature, the selected payment method must have a bank journal attributed.
```
jcs-odoo, line 80: Don't all payment methods require a bank journal anyway?

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,88 @@
...
+
+Alternatively, you can select the :ref:`tip product <tip-product>` on the POS interface to add it to
+the cart. When selected, the product is automatically set as a tip, and its default value equals its
+**Sales Price**.
+
+Tip using an Adyen terminal
+---------------------------
+
+If you use an :doc:`Adyen <../payment/adyen>` payment terminal and wish to enable **tips**
+using the terminal, check :guilabel:`Add tip through payment terminal (Adyen)` when activating the
```
jcs-odoo, line 62: nitpicking, but it can happen at different times and not only when the admins activates the tip setting.

content/applications/sales/point_of_sale/restaurant/tips.rst
```diff
@@ -1,31 +1,88 @@
...
+Configuration
+=============
 
-Under the Bills & Receipts category, you will find *Tips*. Select it
-and create a *Tip Product* such as *Tips* in this case.
+To allow tipping in your POS, activate the :guilabel:`Tips` feature in :menuselection:`Point of Sale
+--> Configuration --> Settings`. At the top of the page, select the POS in which you wish to allow
+**tipping**, scroll down to the :guilabel:`Payment` section and check :guilabel:`Tips`. Once
+enabled, add a :guilabel:`Tip Product` in the corresponding field, and save. The designated product
+will be used as a reference on customers' receipts.
```
jcs-odoo, line 18: Suggestion:
Describe here that they can also enable adyen and tipping after payment, with custom links.

This way you leave the configuration in this section, and can explain only how it works in the other section.

jcs-odoo (changes requested):
Looking good @LoredanaLrpz 
Last suggestions in my comments.

jcs-odoo (approved):

AntoineVDV (approved):
@robodoo r+

PR #3061, opened 2022-12-01
[I18N] application: update 16.0 source terms

rockan007:
update the pot and zh_CN po files


mart-e:
Hello,

Thank you for your PR but could you
- not add new modules, we do not translate some parts of the doc (e.g. developer doc) on purpose
- make the translations on [Transifex](https://transifex.com/odoo/public), not Github.

Best,

rockan007:
OK

PR #3074, opened 2022-12-02
[IMP] paypal: typo update

toaa-odoo:
V14 has a dedicated "Fees" tab for PayPal, current documentation does not reflect that.

taskid: 3089273


jcs-odoo (approved):
@robodoo r+

PR #3172, opened 2022-12-09
[FIX] rdtraining: remove outdated line about `journal_id` in chapter 14

hote-odoo:
In chapter 14 there was a line pointing to how to deduce a default journal for an `account.move` record. This is no longer needed as the journal is automatically deduced using the `move_type`. Also, the link point to an outdated line of code that causes an error.


AntoineVDV (approved):
Retarget to 14.0.

@robodoo delegate+

hote-odoo:
@robodoo r+

AntoineVDV:
@robodoo retry

hote-odoo:
@robodoo retry

PR #3197, opened 2022-12-15
[FIX] core: latex build

Feyensv:
* latest droid package does not allow \usepackage{droid} anymore, replace it with the right package

See https://github.com/MiKTeX/miktex-packaging/issues/136

* remove reference to package texlive-generic-recommended since it's not available nor necessary since ubuntu 20.04.


AntoineVDV (approved):

AntoineVDV (commented):
@robodoo r+

PR #3199, opened 2022-12-15
15.0 security pitfall flvr

flvr-odoo:
Adding 3 sub-sections to the security pitfalls

-Markup
-Http requests good practice
-SQL injection

### DIFF
--- a/content/developer/reference/backend/security.rst
+++ b/content/developer/reference/backend/security.rst
@@ -418,6 +418,46 @@ likely it is to break things.
     >> html_sanitize(code, strip_classes=True)
     '<p>Important Information</p>'
 
+**Markup**
+
+The library Markup is used in Odoo to escape and sanitize the content of html and xml documents.
+When using Markup, the developer must ensure that the user content is passed correctly.
+
+The inside of the constructor of Markup is used for trusted content, you should never pass user-modifiable
+the content inside of the constructor of markup.
+
+.. code-block:: python
+
+    >>text = Markup('<strong> Your username is %s </strong>') % self.env.user.name
+    # We put our HTML inside of the constructor because it is trusted and fully under the control of the developer
+    # Since the user's name is passed by the operator '%', it will be escaped and any HTML code present will not be interpreted by the browser.
+
+Markup can also handle being combined with another Markup object without escaping them:
+
+.. code-block:: python
+
+    >>text = Markup('<i>Your username is : %s</i>') % self.env.user.name
+    # Now we want to make it <strong>
+    >>code = Markup('<strong> %s </strong>') % text
+    # Markup is smart enough not to escape the text a second time. Since this object is already of type Markup, it will be trusted.
+
+Markup can also be used with newer string formating options of python:
+
+.. code-block:: python
+
+    >>text = Markup('Your username is : {name}').format('name': self.env.user.name)
+
+Here is a bad example:
+
+.. code-block:: python
+
+    >>text = f'<strong> Your username is : {self.env.user.name}'
+    >>html = Markup(text)
+    # This example is wrong, we use the data in the control of the user (the username) inside of an f-string, with is directly passed to the constructor of Markup.
+    # Since we always trust the content of the constructor, it will be trusted and interpreted as HTML.
+
+
+
 Evaluating content
 ------------------
 Some may want to ``eval`` to parse user provided content. Using ``eval`` should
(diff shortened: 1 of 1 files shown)


content/developer/reference/backend/security.rst
```diff
@@ -481,3 +515,78 @@ field value can be easily achieved safely:
...
+If the domain can be changed by the user, a malicious user could use the server running the code to send multiple requests to a chosen target.
+This is known as a reflective-DDOS attack, where one can overload a target without sending the request himself.
+
+**URL path/query Injectable**
+
+In the same way as for the domain, care must be taken when manipulating the string of the URL. It should not allow for any user input to be added to the URL. If a user were to modify the URL, he could make the request do other actions than the expected one.
+
+**Timeout**
+
+By default, requests never timeout. If the destination server is not reachable, the running working will halt forever.
```
Julien00859, line 537:
```suggestion
By default, requests never timeout. If the destination server is not reachable, the running worker will halt forever.
```

content/developer/reference/backend/security.rst
```diff
@@ -481,3 +515,78 @@ field value can be easily achieved safely:
...
+For this reason, it is crucial to set a timeout for the request.
+It is best practice to set the timeout slightly higher than a multiple of 3, which is the default retransmission window of TCP.
```
Julien00859, line 539: Of 3 whats? hours? :p

content/developer/reference/backend/security.rst
```diff
@@ -481,3 +515,78 @@ field value can be easily achieved safely:
...
+
+**Error Handling**
+
+Errors should always be handled explicitly, an error response code can happen depending on the state of the other host. In order to avoid a simple crash with traceback presented to the user, you should verify
+response.ok is set to True.
```
Julien00859, line 544: `response.raise_for_status()` is another good overall solution, it raises for any non 100/200/300 http status

content/developer/reference/backend/security.rst
```diff
@@ -418,6 +418,40 @@ likely it is to break things.
     >> html_sanitize(code, strip_classes=True)
     '<p>Important Information</p>'
 
+**Markup**
+
+The library Markup is used in Odoo to escape and sanitize the content.
```
Julien00859, line 423:
```suggestion
The library Markup is used in Odoo to escape and sanitize the content of html and xml documents.
```

content/developer/reference/backend/security.rst
```diff
@@ -418,6 +418,40 @@ likely it is to break things.
...
+When using Markup, the developer must ensure that the user content is passed correctly.
+
+The inside of the constructor of Markup is used for trusted content, you should never pass user-modifiable
+the content inside of the constructor of markup.
+
+.. code-block:: python
+
+    >>text = Markup('<strong> Your username is %s </strong>') % self.env.user.name
+    # We put our HTML inside of the constructor because it is trusted and fully under the control of the developer
+    # Since the user's name is passed by the operator '%', it will be escaped and any HTML code present will not be interpreted by the browser.
```
Julien00859, line 433: Question, does `Markup` supports new-format-style too? In which case I would rather read a `.format` instead of a `%` but that's kinda nitpicking and personal preferences

content/developer/reference/backend/security.rst
```diff
@@ -481,3 +515,78 @@ field value can be easily achieved safely:
...
+
+.. code-block:: python
+
+    # this is not allowed
+    def get_table(self, arg1):
+        self.env.cr.execute("SELECT * FROM table WHERE %s" % arg1)
+
+    # this is allowed
+    def get_table(self, arg1):
+        self.env.cr.execute("""SELECT * FROM table WHERE %s""", (arg1,))
```
Julien00859, line 568: Our own coding guideline at the ORM is to pass the arguments in a list, we find it more readable. It is a slight adaptation of python's DB2 API (which takes a tuple, not a list) but much more readable for 1 argument 

```suggestion
        self.env.cr.execute("""SELECT * FROM table WHERE %s""", [arg1])
```

content/developer/reference/backend/security.rst
```diff
@@ -481,3 +515,78 @@ field value can be easily achieved safely:
...
+
+However, sometimes one need to build a query dynamically,
```
Julien00859, line 570: I think we have a query builder for such cases, @rco-odoo will know more :)

PR #3225, opened 2022-12-21
[MERGE] odoo_theme: SCSS code clean up

edi-odoo:
In order to reduce the size of our css files:
Some unused or repeated code was found and removed. 
Variables have been cleaned up, the way the admonitions (alerts) were created has been modified,

Expanding the use of BS variables in order to reduce repeated code and to stay consistent in the general design.
Sphinx was updated which made some changes in the DOM, breaking some of the CSS: 
the literals for example, these have been fixed in this PR.
The font sizes were breaking the typescale, this has been fixed by modifying the typescale
rather than hardcoding font sizes.


extensions/odoo_theme/translator.py
```diff
@@ -10,15 +10,12 @@
 #         └── Odoo Translator
 
 ADMONITION_MAPPING = {
-    # ???: 'alert-success',
-
-    'note': 'alert-note',
-
-    'hint': 'alert-info',
+# The alert classes have been replaced by default BS classes to reduce number of scss lines.
```
AntoineVDV, line 13: Don't write changelogs in the code. It won't mean anything to new readers :wink: 
```suggestion
```

extensions/odoo_theme/translator.py
```diff
@@ -28,8 +25,8 @@
     'danger': 'alert-danger',
     'error': 'alert-danger',
 
-    'example': 'alert-example',
-    'exercise': 'alert-exercise',
+    'example': 'alert-success',
+    'exercise': 'alert-dark',
 }
```
AntoineVDV, line 30:
```suggestion
}  # Map admonitions to Bootstrap alert classes.
```

extensions/odoo_theme/static/js/layout.js
```diff
@@ -15,9 +15,11 @@
                 modalContainer.innerHTML = `<div class="modal fade" id="modal-${imageModalId}">
                        <div class="modal-dialog modal-dialog-centered">
                          <div class="modal-content">
+                           <div class="modal-header">
+                           <button type="button" class="btn-close btn-close-white" data-bs-dismiss="modal" aria-label="Close"></button>
```
AntoineVDV, line 19:
```suggestion
                               <button type="button" class="btn-close btn-close-white" data-bs-dismiss="modal" aria-label="Close"></button>
```

extensions/odoo_theme/static/style.scss
```diff
@@ -1153,25 +1150,6 @@ header {
     }
 }
 
-//------------------------------------------------------------------------------
-// Image modals
-//------------------------------------------------------------------------------
-
-.o_image_modal {
```
AntoineVDV, line 1222: Shouldn't the class be removed too?

AntoineVDV (commented):
Looking good! :+1: 

I have a few remarks in addition to the review comments, but some are just for double-checking if a change is intended:
- The bar and texts were darker. The footer too.
  image
- Generally speaking, it's good that (almost) all admonitions follow the same style now, but some (e.g., exercise) are maybe too dark and their content hard to read.
  image
- Literals are in bold now. I don't think they should. This is a bit weird on technical pages with a lot of literals.
  image
- I like that h3/h4 are now more easily distinguishable from regular text, but I feel like h4 look even bolder than h3. `<h3>Bulleted list</h3><h4>Bulleted list</h4>` :point_down: 
  image
- Open question, should spoilers have the same style as exercises since they are used together?
  image
- There should be no highlight on spoilers' buttons.
  image
- The close button on modals moved to the left.
  image
- We lost the indentation on /developer/api/extract_api.html#request-body
  image
- In /developer/cli.html, other generic literals are now grey instead of white with a yellow border, but `odoo` still has the yellow border.
  image

extensions/odoo_theme/static/style.scss
```diff
@@ -948,140 +902,95 @@ header {
...
-
-                > .alert-title, > h3 {
-                    color: shade-color($brand-warning, 10%);
-                    &:before {
-                        content: '#{$i-warning}';
-                    }
+            // Note
+            &.alert-primary {
+                > .alert-title:before, > h3:before {
+                    content: '#{$i-note}';
```
stefanorigano, line 924: you can define css variables in the main selector and then apply the customization by class (better inside a loop)... something like:

```scss
.alert {
    > .alert-title:before, > h3:before {
        content: var(--o-alert-icon, #{$i-default-icon});
    }
    
    // [.. smthg similar for borders]
    
    &.alert-success {
        --o-alert-icon: #{$i-whatever};
        --o-alert-border-color: #{$o-color-xhatever}
    }
}



```

extensions/odoo_theme/static/style.scss
```diff
@@ -432,39 +428,80 @@ header {
...
+            padding-left: $padding-s;
+            padding-top: $padding-xs;
+        }
+        &.class em.property {
+            font-size: $font-size-sm;
+            &:first-child {
+                border-bottom: 3px solid $o-literals-border;
+            }
+            &:not(:first-child) {
+                @include o_code;
```
stefanorigano, line 477: typo?
 
```suggestion
                @include o-code;
```

stefanorigano (changes requested):
Thanks @edi-odoo , some notes in the code. Some suggestions:
- you can remove legacy browser's prefixes in _mixins.scss and probably the mixins themselves since rules like `transform-origin` are widespread across all browsers nowadays 
- looks no all `fw_*` class is used except `fw_medium`

extensions/odoo_theme/static/scss/bootstrap_overridden.scss
```diff
@@ -3,22 +3,59 @@
...
+$primary:                                   $o-teal;
+$secondary:                                 $o-violet;
+$light:                                     $o-light;
+$dark:                                      $o-dark;
+$success:                                   $green;
+$info:                                      $blue;
+$warning:                                   $orange;
+$danger:                                    $red;
+
+$theme-colors: (
```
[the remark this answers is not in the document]
edi-odoo: Do you mean $theme-colors?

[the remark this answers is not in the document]
edi-odoo: I've changed all the colors of theme-colors and added new ones.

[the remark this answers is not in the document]
edi-odoo: Indeed I just realised I can inject the `"tip":        $primary,` instead. Duh!

AntoineVDV:
@robodoo r+ merge

PR #3229, opened 2022-12-26
[ADD] eCommerce: shipping page

toaa-odoo:
Added a shipping page to the eCommerce section. Forward to master.

taskid: 3111638


jcs-odoo (changes requested):
Nice. It's much better to only add eCommerce-related info and links to inventory than duplicate the explanations here as well :)

- make sure that there is only one type of change per commit.
- the removal of some pages that won't be updated should be in a separate commit.
- Also, each commit should be auto-sufficient. For example, if you delete a page in a commit, all its references on other pages must be deleted in that same commit.

content/applications/websites/ecommerce/checkout_payment_shipping.rst
```diff
@@ -0,0 +1,10 @@
+:nosearch:
```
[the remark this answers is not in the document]
toaa-odoo: I just copy/pasted an existing :toctree: and changed the files contained. No idea what :nosearch: does, but since it works like that for other :toctree:'s ... :)

Removed it.

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,60 @@
+========
+Shipping
+========
+
+Depending on your shipping strategy, you have the choice to either use your own shipping methods, or
+use an integration with an existing shipping provider.
```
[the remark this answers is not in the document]
toaa-odoo: Good idea

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,60 @@
...
+You can create your own custom shipping methods and define rules to compute shipping costs. To do
+so, go to :menuselection:`Website app --> Configuration --> Shipping Methods`, and either select an
+**existing** shipping method, or :guilabel:`Create` one.
+
+You can choose between :guilabel:`Fixed Price`, :guilabel:`Based on Rules`, and :guilabel:`Pickup in
+store`. :guilabel:`Pickup in store` must first be **enabled** in the settings
+(:menuselection:`Website app --> Configuration --> Settings --> Shipping section)` by checking
+:guilabel:`On Site Payments & Picking`. Once enabled, you can select and :guilabel:`Customize Pickup
+Sites`. The :guilabel:`Picking sites` are **website-specific**, although the same :guilabel:`picking
+site` can be enabled for **multiple** websites.
```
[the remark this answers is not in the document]
toaa-odoo: Added :ref: for the first two options (since they are explained in the ref'd doc), and added an H3 for in-store pickup

toaa-odoo:
> Nice. It's much better to only add eCommerce-related info and links to inventory than duplicate the explanations here as well :)
> 
> * make sure that there is only one type of change per commit.
> * the removal of some pages that won't be updated should be in a separate commit.
> * Also, each commit should be auto-sufficient. For example, if you delete a page in a commit, all its references on other pages must be deleted in that same commit.

I might just use this branch when I'm done to upload all the new doc at once

toaa-odoo:
I'm aware of the conflicts, they will be fixed before merging (other PRs are being merged right now and it will just create further conflicts anyways)

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,71 @@
...
+   - :doc:`/applications/inventory_and_mrp/inventory/shipping/operation/cancel`
+
+Shipping providers
+==================
+
+.. _ecommerce-shipping-providers:
+
+Another solution is to use one of the integration with an existing shipping provider. The advantage
+of using an integration is that delivery cost are automatically computed based on each order as well
+as generating shipping labels.
```
jcs-odoo, line 44:
```suggestion
Another solution is to use one of the integrations with an existing shipping provider. The advantage
of using an integration is that delivery costs are automatically computed based on each order as well
as generating shipping labels.
```

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,71 @@
...
+You can create your own custom shipping methods and define rules to compute shipping costs. To do
+so, go to :menuselection:`Website --> Configuration --> Shipping Methods`, and either select an
+**existing** shipping method, or :guilabel:`Create` one. When creating a shipping method, you can
+choose between
+:doc:`Fixed Price </applications/inventory_and_mrp/inventory/shipping/setup/delivery_method>`,
+:doc:`Based on Rules </applications/inventory_and_mrp/inventory/shipping/setup/delivery_method>`,
+and :guilabel:`Pickup in store`.
+
+Pickup in store
+---------------
```
jcs-odoo, line 23: Do you explain this here only because it's missing from the other page (which seems quite deprecated)?

If it's specific to eCommerce, then it's fine like this. If not, I advise you either add the missing info to the other inventory page, even if the other page is deprecated, or contact the content responsible of this other page to see what can be done.

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,71 @@
...
+
+:guilabel:`Pickup in store` must first be **enabled** in the settings (:menuselection:`Website -->
+Configuration --> Settings --> Shipping section)` by checking :guilabel:`On Site Payments &
+Picking`. Once enabled, you can select and :guilabel:`Customize Pickup Sites`. The
+:guilabel:`Picking sites` are **website-specific**, although the same :guilabel:`picking site` can
+be enabled for **multiple** websites.
```
jcs-odoo, line 29: Actually, they _can_ be website-specific but they don't have to be.

I guess that explaining that shipping methods can be (multi-)website-specific is the important piece of info here, especially if you move the "pickup in store" part to its inventory page where it belongs. But then you already have a dedicated section for this.

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,71 @@
...
+========
+
+Depending on your shipping strategy, you have the choice to either use your :ref:`own shipping
+methods <ecommerce-own-shipping>`, or use an integration with an :ref:`existing shipping provider
+<ecommerce-shipping-providers>`.
+
+Own shipping methods
+====================
+
+.. _ecommerce-own-shipping:
```
jcs-odoo, line 12: put the anchors before the headings, not after. Same below.

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
```diff
@@ -0,0 +1,71 @@
+========
+Shipping
```
jcs-odoo, line 2: "Shipping methods" maybe?

content/applications/websites/ecommerce/checkout_payment_shipping/shipping.rst
toaa-odoo: This is website-specific (This only talks about how to enable it within and configure the basics through the eCommerce app), it doesn't go into any details about configuration of "On Site Pick-up" nor its settings.

jcs-odoo (approved):

AntoineVDV (approved):
@robodoo delegate+

toaa-odoo:
@robodoo r+

PR #2925, opened 2022-11-08
[IMP] Calendar: Outlook sync warning

mivu-odoo:
If committed, this will add a more elaborate warning on what users must do before syncing Odoo Calendar and Outlook Calendar. It also removes mentions of the import/export process and archiving events, as these methods will not prevent email spam from Outlook.

Closes Odoo Task [2937318](https://www.odoo.com/web#id=2937318&cids=3&model=project.task&view_type=form)

PR #3243, opened 2022-12-29
[IMP] Calendar: Outlook sync warning

mivu-odoo:
Need to apply changes on #2925 to `14.0` as well, instead of just `15.0` and `16.0`

> If committed, this will add a more elaborate warning on what users must do before syncing Odoo Calendar and Outlook 
> Calendar. It also removes mentions of the import/export process and archiving events, as these methods will not prevent email 
> spam from Outlook.
>
> closes odoo/documentation#2925
>
> Signed-off-by: Antoine Vandevenne (anv) <anv@odoo.com>


AntoineVDV (approved):
@robodoo r+
@fw-bot ignore

PR #3251, opened 2023-01-02
[IMP] odoo_theme: scroll the active menu entry into view

AntoineVDV:
In large and deeply nested TOCs like finance/accounting/fiscal_localizations/localizations/...
the user had to constantly scroll the menu to the menu entry they just
clicked because the expanded menu would not fit entirely on the page.
With this commit, the active menu entry is scrolled into view each time
the page is reloaded.

The previous behavior was also the reason for expanding the top menu
entries on only the homepage. Thanks to the scroll, they now always
expand.


edi-odoo (approved):
I approve this a lot

AntoineVDV:
@robodoo rebase-ff r+

PR #3269, opened 2023-01-03
[FIX] maintain: correct cname definition

samueljlieber:
This PR fixes the definition of CNAME on line 129 of `maintain/domain_names.rst`.

From:
> A CNAME record is a type of DNS record that points to the IP address of another website rather than to directly to an IP address.

To:

> A CNAME record is a type of DNS record that points to the domain of another website rather than directly to an IP address.

### DIFF
--- a/content/administration/maintain/domain_names.rst
+++ b/content/administration/maintain/domain_names.rst
@@ -126,8 +126,8 @@ that you proceed with the following actions in this order:
 Add a CNAME record
 ------------------
 
-A **CNAME record** is a type of DNS record that points to the IP address of another website rather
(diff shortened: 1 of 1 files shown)


jcs-odoo (approved):
Hi @samueljlieber 

Thanks for the fixes (CNAME and missing bracket).
I'll revert the other changes (going to the line), as these lines are already within 100 characters in length.

jcs-odoo:
@robodoo r+

jcs-odoo:
@robodoo retry

jcs-odoo:
@robodoo r+

PR #3295, opened 2023-01-04
[FIX] name field of ir.ui.view is optional

legalsylvain:
Supersed : https://github.com/odoo/documentation/pull/2785

CC : @AntoineVDV

### DIFF
--- a/content/developer/reference/backend/views.rst
+++ b/content/developer/reference/backend/views.rst
@@ -37,9 +37,10 @@ Fields
 
 View objects expose a number of fields. They are optional unless specified otherwise.
 
-* ``name`` (mandatory) :class:`~odoo.fields.Char`
+* ``name``  :class:`~odoo.fields.Char`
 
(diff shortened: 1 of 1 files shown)


AntoineVDV:
@legalsylvain Now that the CI checks are fixed, the responsible persons are assigned for review. I'll let them have a look at your PR ;)

content/developer/reference/backend/views.rst
```diff
@@ -37,9 +37,10 @@ Fields
 
 View objects expose a number of fields. They are optional unless specified otherwise.
 
-* ``name`` (mandatory) :class:`~odoo.fields.Char`
+* ``name``  :class:`~odoo.fields.Char`
```
VincentSchippefilt, line 40: I think this is a bit misleading: functionally and database-wise, the name _is_ required. However if you do not provide one, it will default to `model_name view_type`

PR #3301, opened 2023-01-06
[MOV] purchase: moving accounting doc to purchase section

Donapi:
Moving the From Vendor Bills to Payment document page from the accounting section to the Purchase section, following discussion with @BRSE. This doc is more relevant to the Purchase app.

taskid: 2687652


Donapi:
Hi @brse-odoo and @larm-odoo, as per our discussion, I moved this page under the Purchase section :) Thanks!

PR #3322, opened 2023-01-12
[IMP] web: list: document arch attribute

fdardenne:
The `open_on_click` arch attribute is introduced in odoo/odoo#108903. This PR document it.


fdardenne:
spec changed

PR #3332, opened 2023-01-13
Inform new users how long before a new dev branch is ready

msafi:


jcs-odoo:
Thank you for your [contribution to the Odoo documentation](https://www.odoo.com/documentation/master/contributing/documentation.html), @msafi !

I'll apply a small change.

jcs-odoo:
> @jcs-odoo This change should not be merged because:
> 
> 1. the time depends on the dependencies of the module itself (1 hour is when you test website or another huge module), so basically saying that it takes 1 hour is false
> 2. we are always trying to reduce the testing time (because Odoo.sh is not runbot) and we are trying to avoid executing standard tests, if we succeed, huge modules will not take the same time as now.

That is fine by me, thanks for the explanation @amigrave

PR #3376, opened 2023-01-20
[FIX] developer/guidelines: odoo addons import order

nle-odoo:
The example for import order is a little confusing since it is wrong.


AntoineVDV (approved):
@robodoo r+

PR #3410, opened 2023-01-26
[IMP] install/maintain: update content of upgrade SLA

avg-odoo:
As per the head of the upgrade dept, the Service Level Agreement has to be updated. The maintenance fees cover the maintenance of the custom code (whoever did it) and are not specifically dedicated to custom code written by Odoo development teams

### DIFF
--- a/content/administration/upgrade.rst
+++ b/content/administration/upgrade.rst
@@ -225,15 +225,14 @@ Service Level Agreement
 What is covered by the Enterprise Licence?
 ------------------------------------------
 
-Databases hosted on Odoo’s Cloud platforms (Odoo Online and Odoo.sh) or On-Premise (Self-Hosting)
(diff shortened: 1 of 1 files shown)


content/administration/upgrade.rst
```diff
@@ -232,7 +232,7 @@ The upgrade of:
 
 * standard applications
 * Studio customization (as long as the :guilabel:`Studio` app is still active)
-* customizations done by our consulting and developer services *if* they are covered by a
+* customizations and developer services *if* they are covered by a
   ‘Maintenance of Customisations’ subscription
```
augustoPerez, line 236: IINM, the `developer services` reffers also to who did the customizations
IMO, we should remove all the `done by our consulting and developer services` text and just leave:

```suggestion
* customizations *if* they are covered by a ‘Maintenance of Customisations’ subscription
```

jcs-odoo (approved):

jcs-odoo:
@robodoo r+

PR #3445, opened 2023-01-31
[IMP] install/maintain: clarification in the upgrade SLA

thkl-odoo:
The Service Level Agreement has to be updated. Any additional line of code, without using studio, in standard modules is not supposed to be covered by the upgrade team. It is slightly different than the point telling "development of custom modules"


jcs-odoo (approved):

jcs-odoo:
@robodoo r+

PR #3471, opened 2023-02-02
Update web.rst

mamume:


AntoineVDV:
Hello, thank you for your suggestion :)

I prefer not to merge this because "javascript" is acceptable, and it's not worth going through the whole merge and forward-port (with potential conflicts) process just for this. We will fix this typo later when we re-write the page.

PR #3504, opened 2023-02-08
[FIX] web: fix wrong ActionSwiper documentation

Iucapad:
Since 16.0, the documentation for this component was no longer correct, since the component was moved to web during the refactor of the webclient views. This commit also adds documentation of props that were introduced during the refactor, already present in 16.0.


fdardenne (approved):

AntoineVDV (approved):
@robodoo r+

PR #3523, opened 2023-02-09
[IMP] reference: rewrite and move views documentation to user_interface

Gorash:
The documentation of the views is redone in order to be up to date and
has allowed a better visualization of the task on the modifiers.

https://github.com/odoo/odoo/pull/104741
https://github.com/odoo/enterprise/pull/35255
https://github.com/odoo/upgrade/pull/4884

task-2495504


content/developer/reference/user_interface/views.rst
```diff
@@ -0,0 +1,109 @@
...
+.. note::
+    In the views architecture, some XML node attributes canuse `python expression`_,
+    in this case this expression is evaluated with the field values of the current view
+    and this other values:
+
+    * ``context``: dict_ with the values of the current view context;
+    * ``uid``: integer_ the id of the current user;
+    * ``today``: string_ the current local date as a string of the form ``YYYY-MM-DD``;
+    * ``now``: string_ same as ``today`` with the addition of the current time. This value is formatted as ``YYYY-MM-DD hh:mm:ss``;
+    * ``parent``: The properties of this "object" refere to the container view values. This magic object is only available for the subviews of :ref:`relational fields <studio/fields/relational-fields>`.
```
VincentSchippefilt, line 88: are not all the fields of the current record available ?

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+  <form>
+    ...
+  </form>
+
+Optional attributes_ are added on root element ``<form>`` to customize the view.
+
+:string:
+  string_
+
+  View name
```
VincentSchippefilt, line 21: used where and for what ?

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+:delete:
+  boolean_
+
+  Disable/enable record deletion on the view through the **Action** dropdown.
+
+:js_class:
+  string_
+
+  Name of the javascript class to use when instantiating the view.
+
```
VincentSchippefilt, line 47:
```suggestion
:js_class:
  string_

  Name of the javascript component the webclient will instantiating instead of the form the view.

```

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+:disable_autofocus:
+  boolean_
+
+  Disable automatic focus of the first field in the view.
```
VincentSchippefilt, line 51:
```suggestion
:disable_autofocus:
  boolean_

  Disable automatic focussing of the first field in the view.
```

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+
+.. _reference/user_interface/views/form/semantic:
+
+Semantic components
+-------------------
+
+Semantic components tie into and allow interaction with the Odoo system.
+Placeholders are denoted in all caps.
```
VincentSchippefilt, line 59:
```suggestion
Semantic components tie into the Odoo system and allow interaction with it.
remark: (with the remark style)
Placeholders are denoted in all caps.
```

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+
+.. code-block:: xml
+
+  <form>
+    <field name="FIELD_NAME"/>
+  </form>
+
+renders (and allow editing of, possibly) a single field of the current
+record. Using several times a field in a form view is supported and the fields
+can receive different values for modifiers 'invisible' and 'readonly'. However,
```
VincentSchippefilt, line 74: shouldn't fields receive different values for 'invisible' so only one is visible at a time ?

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+  (including default widgets)
+
+  .. code-block:: xml
+
+    <field name="tag_ids" widget="many2many_tags" options="{'color_field': 'FIELD_NAME', 'no_quick_create': True}"/>
+
+:groups:
+  `Comma-separated values`_ whose choices are the :class:`~odoo.addons.base.models.res_users.Groups` reference
+
+  only displays the field for specific users
```
VincentSchippefilt, line 115: could you add/link to the syntax of the groups: what does ! does, how to find the actual string that represent the group (AFAIR it's not the name but the xmlid) etc. pls :)

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+
+  .. code-block:: xml
+
+    <field name="fname" context="{
+        'TYPE_view_ref': 'ADDON.MODEL_view_TYPE',
+        'group_by': 'FIELD_NAME',
+        'default_FIELD_NAME': ANY,
+        'search_default_FIELD_NAME': True,
+        'OTHER_BUSINESS_KEY': ANY,
+      }"/>
```
VincentSchippefilt, line 144: is there some place where we explain the *special* behavior of certain context keys, like `default_* ?
`

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+:invisible:
+  :ref:`python expression <user_interface/views/python_expression>` that defines a boolean_
+
+  standard dynamic attributes based on record values. Hide the field
+  if trully or if the evaluate expression is trully.
+
+  .. code-block:: xml
+
+    <field name="fname_a" invisible="True"/>
+    <field name="fname_b" invisible="fname_c != 3 and fname_a == parent.fname_d"/>
```
VincentSchippefilt, line 179: could you explain why it is useful to have invisible fields on a view, pls

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+
+:nolabel:
+  boolean_
+
+  don't automatically display the field's label, only makes sense if the
```
VincentSchippefilt, line 184:
```suggestion
  if True, do not automatically display the field's label, only makes sense if the
```

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+:kanban_view_ref:
+  string_ defined by the pattern: ``%(ADDON.MODEL_view_TYPE)s`` (target the :ref:`view reference <reference/view_record>`)
+
+  for opening specific kanban view when selecting records from m2o/m2m in mobile
+  environment
+
+:default_focus:
+  boolean_
+
+  defines the field on which the focus will be made when displaying the form view
```
VincentSchippefilt, line 243:
```suggestion
  If True, defines that this field is the fields that will be focussed when the view opens. Cannot be present on more than one field of a view.
```

content/developer/reference/user_interface/views/include/form.rst
```diff
@@ -0,0 +1,778 @@
...
+  .. image:: views/form_group.svg
+    :class: col-xxl-6
+
+.. _reference/user_interface/views/form/sheet:
+
+<sheet>: Responsive layout
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``<sheet>`` can be used as a direct child to ``<form>`` for a narrower and more responsive
+form layout. Usually it contains :ref:`<group> <reference/user_interface/views/form/group>`.
```
VincentSchippefilt, line 503: can you explain what sheet does to make the form more responsive ?

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+  <reference/user_interface/views/form>`'s fields and buttons are thus accepted by list
+  views although they may not have any meaning if the list view is non-editable.
+
+  If the ``edit`` attribute is set to ``false``, the ``editable`` option **will be ignored**.
+
+:multi_edit:
+  ``1``
+
+  editable or not editable list can activate the multi-editing feature by defining the
+  `multi_edit="1"`
```
VincentSchippefilt, line 74:
```suggestion
  editable or not editable list can activate the multi-editing feature that allows to change the same field to the same value for multiple records in a single operation by defining the
  `multi_edit="1"`
```

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+
+
+  whether the list view should be grouped if no grouping is specified via the action or
+  the current search. Should be the name of the field to group by when no grouping is
+  otherwise specified
+
+:default_order:
+  `Comma-separated values`_
+
+  overrides the ordering of the view, replacing the model's order (:attr:`-odoo.models.BaseModel._order` model attribute).
```
VincentSchippefilt, line 87:
```suggestion
  overrides the ordering of the model, replacing the model's order (:attr:`-odoo.models.BaseModel._order` model attribute).
```

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+  The value is a comma-separated list of :ref:`fields <reference/orm/fields>`, postfixed by ``desc`` to
+  sort in reverse order:
+
+  .. code-block:: xml
+
+    <tree default_order="sequence,name desc">
+      ...
+    </tree>
+
+:decoration-{$name}:
```
VincentSchippefilt, line 97:
```suggestion

:decoration-bf:
:decoration-it:
:decoration-danger:
:decoration-warning:
:decoration-info:
:decoration-muted:
:decoration-primary:
:decoration-success:
```

in this case I would specify them all (so it's easier to find) and explain just once what this does with the variations

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+  .. code-block:: xml
+
+    <tree decoration-danger="field_qty > field_limit">
+      ...
+    </tree>
+
+:limit:
+  integer_
+
+  the default size of a page. It must be a positive integer
```
VincentSchippefilt, line 121: add the default: 
80 for list view, 
40 for x2many list in form view (to be verified)

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+
+:groups_limit:
+  integer_
+
+  when the list view is grouped, the default number of groups of a page. It must be a
+  position integer
+
```
VincentSchippefilt, line 128: add the default pls

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+:expand:
+  boolean_
+
+  when the list view is grouped, automatically open the first level of groups if set
+  to true (default: false)
```
VincentSchippefilt, line 133: maybe add a warning that this ca be very slow (today)

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+  string_
+
+  the title of the field's column (by default, uses the ``string`` of
+  the model's field)
+
+:optional:
+  string_ chooses from ``hide`` or ``show``
+
+  if set, the visibility of the field is optional. The display can be chosen
+  via a button in the form view. By default fields will be visible if the value
```
VincentSchippefilt, line 168:
```suggestion
  via a button in the list view. By default fields will be visible if the value
```

content/developer/reference/user_interface/views/include/list.rst
```diff
@@ -0,0 +1,569 @@
...
+  is `show` or not visible if `hide`.
+
+  .. code-block:: xml
+
+    <field name="fname_a" optional="hide"/>
+
+:readonly:
+  :ref:`python expression <user_interface/views/python_expression>` that defines a boolean_
+
+  standard dynamic attributes based on record values. If the value is trully
```
VincentSchippefilt, line 178: replace all the trully with truthy :)

content/developer/reference/user_interface/views/include/search.rst
```diff
@@ -0,0 +1,339 @@
+.. _reference/user_interface/views/search:
+
+Search
+======
+
+Search views are a break from previous view types in that they don't display
+*content*: although they apply to a specific model, they are used to filter
+other view's content (generally aggregated views
```
VincentSchippefilt, line 8:
```suggestion
Search views are different from other view types: they don't display
*content*. Although they apply to a specific model, they are used to filter
another view's content (generally aggregated views
```

content/developer/reference/user_interface/views/include/search.rst
```diff
@@ -0,0 +1,339 @@
...
+e.g. :ref:`reference/user_interface/views/list` or :ref:`reference/user_interface/views/graph`). Beyond that
+difference in use case, they are defined the same way.
```
VincentSchippefilt, line 10:
```suggestion
e.g. :ref:`reference/user_interface/views/list` or :ref:`reference/user_interface/views/graph`).
```

content/developer/reference/user_interface/views/include/search.rst
```diff
@@ -0,0 +1,339 @@
...
+.. @string is not displayed anywhere, should be removed
+
+Possible children elements of the search view are:
+
+.. rst-class:: o-definition-list
+
+``field``
+    fields define domains or contexts with user-provided values. When search
+    domains are generated, field domains are composed with one another and
+    with filters using **AND**.
```
VincentSchippefilt, line 23:
```suggestion
    fields define domains or contexts with user-provided values. When search
    domains are generated, field domains are joined with each other and
    with filters using the **AND** operator.
```

content/developer/reference/user_interface/views/include/search.rst
```diff
@@ -0,0 +1,339 @@
...
+    ``name``
+        the name of the field to filter on
+    ``string``
+        the field's label
+    ``operator``
+        by default, fields generate domains of the form :samp:`[({name},
+        {operator}, {provided_value})]` where ``name`` is the field's name and
+        ``provided_value`` is the value provided by the user, possibly
+        filtered or transformed (e.g. a user is expected to provide the
+        *label* of a selection field's value, not the value itself).
```
VincentSchippefilt, line 38: I don't understand this part.
I think this needs to be reorganized. I would define the property operator as something like 
"Changes the default operator of a filter. The default operator depends on the field type (e.g. = for float and ilike char). "
Also, do you think of a good use case to use as an example ?

Gorash: I did not improve the doc for this view there...

conf.py
```diff
@@ -56,6 +56,7 @@
 exclude_patterns = [
     'locale',
     'README.*',
+    '**/include',
```
AntoineVDV, line 59: I'm really not a fan of this hack. It looks like you factor content out of `views.rst` only for aesthetics, and it backfires in multiple ways: content built twice, duplicated refs, headings defined in the included file...

Instead, you could insert the RST directly where it's included, and it would be perfectly fine.

Alternatively, you could also create one page per view type and neatly nest them under `views`.

content/developer/reference/user_interface.rst
```diff
@@ -7,5 +7,7 @@ User interface
 .. toctree::
     :titlesonly:
 
+    user_interface/view_record
```
AntoineVDV, line 10: What is the purpose of this page?

IMHO, it should be merged into `views.rst`, which would be easily restructured if each view type is moved as a separate sub-page.

redirects/13.0.txt
```diff
@@ -167,7 +167,8 @@ developer/reference/orm.rst                               developer/reference/ad
 developer/reference/reports.rst                           developer/reference/addons/reports.rst
 developer/reference/security.rst                          developer/reference/addons/security.rst
 developer/reference/testing.rst                           developer/reference/addons/testing.rst
-developer/reference/views.rst                             developer/reference/addons/views.rst
```
AntoineVDV, line 170: The page is moved in master; therefore you should add the redirect rules in a new `redirects/saas-16.4.txt` file.

AntoineVDV (commented):
Hello, kudos for taking the initiative to revamp this doc :+1: 

I posted a few preliminary comments below to discuss the structure before I dive deeper and review the content itself.

conf.py
Gorash: It seems to me that there is an option to not build the includes but just the pages that use them. I was counting on this to simplify centralizing the doc makes it possible to update all the places at the same time and not to have an outdated piece. Code of doc aesthetics = easier to maintain. According to your comment I was wrong in the scope of the rst options.

The include is indeed used to not have a gigantic atrocious page to work in. I could split the pages My goal was to have the h1 for the views and h2 for each of the views. Either way, it doesn't change much. This will add titles to the left reference menu.

content/developer/reference/user_interface.rst
Gorash: View and record views are two separate things.
It is not wise to mix the semantics of the views with the record that contains them. That was the error before. Record options are not equal to view options. Putting them together is just confusing.

A record is any object and in this case it is a view. We could make a csv that would not change the view. We can go to the client database we see views with the arch, we don't see <record>....

content/developer/reference/user_interface/views.rst
Gorash: It depends on the place. For example a column of a tree in an x2m does not have the values for the records.

I will add this to the doc.

conf.py
AntoineVDV: Then maybe go with "Alternatively, you could also create one page per view type and neatly nest them under views." for a start since it requires less work, and we'll see how it looks? FYI, there are UI improvements coming in to better render sub-pages :)

If the proposed structure poses a problem, I think it'd be better to stick with the giant RST file then.

content/developer/reference/user_interface.rst
AntoineVDV: Okay, makes sense. It wasn't clear to me when scanning through the content that the distinction what about that, but again I didn't read everything yet.

content/developer/reference/user_interface/view_architecture.rst
```diff
@@ -0,0 +1,3774 @@
...
+'readonly'. This fields have the same values but can be display
+differentaly. However, the behavior is not guaranteed when several
+fields exist with different values for modifier 'required'. ``<field>``
+can have the following attributes:
+
+:name:
+  string_ (mandatory) :ref:`model <reference/orm/model>` field name
+
+  the name of the field to render
+
```
AntoineVDV, line 204: Missing `:string:` attribute.

AntoineVDV (commented):
Hello, while reviewing this PR, I started making heavy changes to how elements and attributes are presented to follow how it's done in other reference pages and make everything more resilient to theming revisions. Since it will take longer than expected, I split my commits in two and kept only those that can be easily squashed in this branch, while the others have been moved to #5237, which will be merged when ready.

It's good work overall; this doc deserved a refresh :)

I added a few comments below when I was unsure, please have a look.

content/developer/reference/user_interface/view_architecture.rst
```diff
@@ -0,0 +1,3774 @@
+=================
+View architecture
+=================
+
+Generic view
+============
+
+.. _user_interface/view_architecture/options:
+
+Options / Attributes
```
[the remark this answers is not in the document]
AntoineVDV: Actually, it's already done in #5237, so it can stay as-is in this PR.

content/developer/reference/user_interface/view_architecture.rst
AntoineVDV: Done in #5237, it can stay as-is in this PR.

content/developer/reference/user_interface/view_architecture.rst
```diff
@@ -0,0 +1,3774 @@
...
+  :ref:`controller route url<reference/controllers>` will be fetched and
+  displayed above the view. The json response from the controller should
+  contain an "html" key.
+
+  If the html contains a stylesheet <link> tag, it will be
+  removed and appended to <head>.
+
+  To interact with the backend you can use <a type="action"> tags. Please take
+  a look at the documentation of the _onActionClicked method of
+  AbstractController (*addons/web/static/src/js/views/abstract_controller.js*)
```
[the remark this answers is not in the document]
Gorash: Indeed. They updated the js framework and I did not adapt.

content/developer/reference/user_interface/view_architecture.rst
```diff
@@ -0,0 +1,3774 @@
...
+
+  The styling use the Bootstrap_ framework and :doc:`UI icons <icons>`.
+
+  Below are the common Odoo_ classes:
+
+  ``oe_inline``: prevent the usual line break following fields and limit their span.
+
+  ``oe_left``, ``oe_right``: floats_ the field to the corresponding direction
+
+  ``oe_read_only``, ``oe_edit_only``: only displays the field in the corresponding form mode
```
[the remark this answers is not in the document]
Gorash: yes
```scss
    .o_form_editable {
        .oe_read_only {
            display: none!important;
        }
```

content/developer/reference/user_interface/view_architecture.rst
```diff
@@ -0,0 +1,3774 @@
...
+direct children of groups take a single column.
+:ref:`<field> <reference/user_interface/view_architecture/form/field>` direct children of groups
+display a ``label`` by default, and the label and the field itself have a colspan of 1
+each.
+
+Children are laid out horizontally (tries to fill the next column before changing row).
+
+``<group>`` can have the following attributes:
+
+:col:
```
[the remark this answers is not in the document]
Gorash: nop
```js
QUnit.test("form rendering with groups with col/colspan", async function (assert) {
```

AntoineVDV (approved):
@robodoo delegate+

Julien00859:
can someone give me the delegation? chm is off today and we're taking responsibility to merge his work with @odoo/rd-framework-editor

AntoineVDV:
@robodoo delegate=Julien00859

Julien00859 (approved):
@robodoo r+

PR #3535, opened 2023-02-10
[MOV] finance: move fiscal localisation pages and merge /overview/

jcs-odoo:
The content written in the fiscal localisation pages are not only for Accounting, but also for other apps such as Payroll and Point of Sales. This commit moves the content to the "Finance" level, to better reflect this reality, and merges the pages in the /overview/ folder in the main category page.


jcs-odoo:
@robodoo r+

PR #3536, opened 2023-02-10
[FIX] *: first small clean of the code

jcs-odoo:
Friday afternoon cleanup. This is required so we can use "make test", as there are currently hundreds of errors. For now, it is unusable because of the oldest code in this repo.


jcs-odoo:
@robodoo r+

PR #3548, opened 2023-02-13
[FIX] legal: RST cleanup (headings delimiters characters)

jcs-odoo:
RST cleanup to comply with the RST guidelines. This is required so we can use "make test", as there are currently hundreds of errors. For now, it is unusable because of the oldest code in this repo.

### DIFF
--- a/content/legal/terms/enterprise.rst
+++ b/content/legal/terms/enterprise.rst
@@ -153,7 +153,7 @@ or in Covered Extra Modules.
 .. _secu_self_hosting:
 
 Self-Hosting
-++++++++++++
+~~~~~~~~~~~~
 
 For the duration of this Agreement, Odoo SA commits to sending a "Security Advisory" to the Customer
 for any security Bug that is discovered in the Covered Versions of the Software (this excludes Extra
@@ -169,7 +169,7 @@ the public disclosure.
 .. _secu_cloud_platform:
 
 Cloud Platform
-++++++++++++++
+~~~~~~~~~~~~~~
 
 Odoo SA commits to apply the security remedies for any security Bug discovered in a version of
 the Software hosted on the Cloud Platform, on all systems under its control, as soon as
@@ -184,7 +184,7 @@ the remedy is available, without requiring any manual action of the Customer.
 .. _upgrade_odoo:
 
 Upgrade Service for the Software
-++++++++++++++++++++++++++++++++
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 For the duration of this Agreement, the Customer can submit upgrade requests through the appropriate
 channel (typically Odoo SA's upgrade service website), in order to convert a database of the Software
@@ -235,7 +235,7 @@ https://www.odoo.com/cloud-sla.
 --------------------
 
 Scope
-+++++
+~~~~~
 
 For the duration of this Agreement, the Customer may open an unlimited number of support tickets
 free of charge, exclusively for questions regarding Bugs (see :ref:`bugfix`) or guidance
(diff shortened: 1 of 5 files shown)


content/legal/terms/i18n/partnership_fr.rst
```diff
@@ -331,10 +337,11 @@ Partie impliqué dans l’exécution ou l’utilisation des Services repris dans
 pour toute la durée de l’accord et une période de 12 mois suivant la date de résiliation ou
 d’expiration de ce Contrat. En cas de non-respect des conditions de cette section qui mène à la
 résiliation dudit travailleur à cet effet, la Partie fautive accepte de payer à l’autre Partie
-la somme de 30 000,00 (trente mille) euros (€).
+la somme de 30 000,00 (trente mille) euros (€).
```
jcs-odoo, line 340: J'ai remplacé plusieurs espaces insécables par des espaces normaux dans ce fichier, en partant du principe qu'il s'agissait d'un résidu d'un copier-coller. En tout cas, ces espaces insécables ne sont pas présents dans les autres documents.

jcs-odoo:
@odony 
This PR doesn't change anything to the content. It's only an RST cleanup so we can use an automated check while writing.

jcs-odoo:
Hello @odony and @AntoineVDV 

I merely applied [our RST guidelines](https://www.odoo.com/documentation/master/contributing/documentation/rst_guidelines.html), so we can use the lint (and new custom tests) locally. 

There is nothing about minimum line length, though. But going to the line in RST may mislead the writer into thinking that the text is easier to read than how it will be once built in HTML.

But I see your point, and that would be a discussion between the two of you, then :)

AntoineVDV:
It's not planned to enforce a minimum line length. The general recommendation is only to *not* break the line because of the punctuation, which might lead the reader in believing that the text will be rendered onto multiple lines in HTML (it will not).

For example, this :point_down: 
```rst
Odoo is a suite of open source business apps:
CRM, eCommerce, accounting, inventory, point of sale, project management, etc.
```
is not recommended.

jcs-odoo:
@Feyensv told me that the line length won't be included in the local tests. I thought it would be.

@odony 
let me know if you want me to revert the changes made to the lengths of the lines, then?
Or is it okay to merge this PR as is?

AntoineVDV:
@odony There was a misunderstanding between me and @jcs-odoo regarding what we wanted to change in existing files regarding our guidelines.

For the context, we are working on adding custom (guidelines-related) linting checks that will apply to new PR's diff, same as for the R&D's `ci/style` check. There are currently no other checks in place other than sphinxlint's defaults. 

There are, however, *a few* custom checks that I'd want to use to fix existing files. For example, absolute references to the documentation (`https://...` instead of `:doc:`), valid but error-inducing RST (e.g., heading separators using random characters whereas 99% of other files use the same, making it it hard for writers to figure headings order out), unused files, etc.

I created a stupid command-line tool for @jcs-odoo that outputs every warning instead of only those we're interested in, hence this PR's diff. After update, this PR will only target problematic RST and fix visual inconsistencies.

jcs-odoo:
Indeed, apologies for the misunderstanding. @odony @AntoineVDV 

There are two commits:
- the first one fixes error-inducing RST, as explained by Antoine. In this case, it adapts the characters used as delimiters and adds missing blank lines when needed.
- the second one is questionable and can be dropped if it's inappropriate: it adapts the length of seven headings flagged as errors by the command-line tool that we would use when we work on the documentation repo.

I also have a third commit on another local branch, but I wanted to leave it out as it adds a lot of noise (especially in the Spanish and French translations).
It primarily replaces all apostrophes with the character `'` and quotation marks with the character `"` for consistency. Currently, the RST code is a mix of text copy-pasted from external text editors and edits made directly in RST. As a result, there is a mix of symbols, sometimes in the same sentence.
Here is an example:
image
It's important to note that the character `"` is displayed differently in HTML according to the lang attribute in the HTML head.
Mixed characters sometimes give this kind of results
image

AntoineVDV:
Let's go ahead with this PR since it now addresses only problematic RST. It can still be reverted if it poses a problem.

@robodoo r+

PR #3573, opened 2023-02-14
[ADD] web: add SelectMenu documentation

Iucapad:
This commit adds the documentation for the SelectMenu component, introduced in saas-16.1. This documentation gives some example of its usage.

An image is also added, to better illustrate the grouping functionality.


AntoineVDV (approved):
:ok_hand:

@robodoo r+

PR #3579, opened 2023-02-15
[FIX] 02_setup: remove gramatical error

Melidee:
in line 53 of the 02_setup portion of the getting started docs, there is an unnecessary doubled word "based", one of them should be removed


AntoineVDV (approved):
Hello, thank you for the fix :)

@robodoo r+

Issue #3480, opened 2023-02-03
Pitfall in the official odoo 16 Document - proxy_set_header, https, reverse proxy, nginx, location, redirection


skygunner:
What are the steps to reproduce your issue?
There's a pitfall in the odoo 16.0 document , system configration part
https://www.odoo.com/documentation/16.0/administration/install/deploy.html#https

while transiting from odoo 15 to 16, the
`
location /longpolling
` 
part has been changed to
**location /websocket**
and the developer put in some extra in the
location /websocket {} block:
```
location /websocket {
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection $connection_upgrade;
  proxy_set_header X-Forwarded-Host $host;
  proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  proxy_set_header X-Forwarded-Proto $scheme;
  proxy_set_header X-Real-IP $remote_addr;
}
```
**without remembering that if proxy_set_header was set in the Nginx's location match
level, then those set in the previous configration level like server block will no longer be effective.
Ref:
These directives are inherited from the previous configuration level if and only if there are no proxy_set_header directives defined on the current level.**

What is the current behavior that you observe?

So this change breaks the four lines above in the server block:
```
# Add Headers for odoo proxy mode
proxy_set_header X-Forwarded-Host $host;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_set_header X-Forwarded-Proto $scheme;
proxy_set_header X-Real-IP $remote_addr;
```
This made the backend can not get the $host value, while the user setup the dbfilter with anything but  **.*** , they will be granted with a new database creation page due to the backend has no corrrect $host value.
And this made the backend lost the $scheme, if the user is using https
by default, the returned http header's location part would become http://domainname/web instead of https://domainname/web

Though later connection will be redirected to https after the browser learned the server has that, but the first connection will always be
**'The connection to backend is not secure'** 
greeting page to visitor.

Combined with proxy_redirect off; the backend might just keep redirecting to the upstream server's name and can not continue to work.
Lost $proxy_add_x_forwarded_for and $remote_addr might also affect other things.

If the user tried to search the internet to find a solution for the above issues, they may fall into use
#proxy_set_header Host $host; to make the dbfilter work (but does not fix the location http:// problem in the response header). And this may lead to other problem in odoo's working as it changes many behavior on the resource/endpoints calls.

What would be your expected behavior in this case?

[...]

PR #3628, opened 2023-02-20
[FIX] deploy: add set_headers, fixing wrong $host value

tsm-odoo:
This PR fixes a pitfall in the documentation. The `location /websocket` block uses several `proxy_set_header`. Those `proxy_set_header` disable the other ones at the same level (e.g. in the `location /` block).

The solution is to use `proxy_set_header` in the other blocks that need it on the same level.

closes: @odoo/documentation#3480


AntoineVDV (approved):
@robodoo r+

PR #3639, opened 2023-02-20
Update sales_order.rst

SampoVirmasalo:
Added crucial note about service products. This feature caused a lot of headache to us while figuring out why draft orders (quotations) we tried to settle in POS unset the service product quantities.


LoredanaLrpz:
Thanks for your input @SampoVirmasalo, I'll take a look into that to make this info a part of the documentation
Have a great day,

PR #3647, opened 2023-02-22
Make absence of rate limits explicit

monty241:
Many cloud platforms use rate limits in various ways (per minute, per hour, per day, per company, per database, per table, whatever) to limit loads. Odoo online currently does not seem to apply those. This change makes that explicit.

### DIFF
--- a/content/developer/api/external_api.rst
+++ b/content/developer/api/external_api.rst
@@ -118,6 +118,11 @@ Once you have keys configured on your account, they will appear above the
 **A deleted API key can not be undeleted or re-set**. You will have to generate
 a new key and update all the places where you used the old one.
 
+Rate limits
+~~~~~~~~~~~
+
+The Odoo API has no rate limits.
+
 Test database
 ~~~~~~~~~~~~~
 


monty241:
Thanks for your feedback!

We are creating a SQL-driver that real-time exchanges messages with 1..1000 Odoo databases of a tenant. Currently it is configured to just go ahead. We focus on Odoo.com and .sh online, since other users have direct access to PostgreSQL and upload features are typically only used by 10% of the user population.

Is there any way to quantify the acceptable use? For instance, at most 100 calls per minute for a tenant, or 100 calls per minute per database?

Just to avoid downtime or perceived sluggishness by other users when a large user fires in highly parallel mode an extreme number of API requests (we have had some large accounting firm bring down one of your competitors with a single query).

monty241:
Thank you for the extensive reply. We will limit it to 1 call/sec per database since 1 call/sec per tenant (which I assume is the wide definition of the counter party paying the subscription) has a very low throughput when there are hundreds of databases to query. This is generally throttled to a configurable number between 8 and 32 parallel queries across all databases.

The SQL-engine definitely applies a large number of optimiazations, including what we call "server-side filtering", also known as "query-folding" on Power BI, similar to a transitive relationship. So in general filters are forwarded to reduce the amount of data processed as much as possible. Not all filters can be forwarded, such as for instance an UPPER SQL-function in an equality. Database identifiers ("partitions") are filtered early on to reduce number of databases queries.

Caching is done in various ways, but I am sorry to tell that aggregations are not yet supported (select NAME, count(*) from MODEL where NAME like 'BE%' group by NAME retrieves all 'BE%' names first).

The SQL-engine is not meant to compete with Odoo own's tools but to supplement them, for instance to allow the data integration (synchronize Odoo and some other systems like WMS), or a Windows Explorer plugin or Excel plugin to exchange data seamlessly.

Thanks for the tips, I highly appreciate it!

monty241:
In between time I have understood that on Odoo, it seems that "tenant" is defined as "database" which somewhat deviates from other packages, and that for accountants there is a separate pricing model which eases many databases. We will limit it at 1 call / second per database.

PR #3706, opened 2023-03-01
[FIX] website: syntax error (fix for translators)

toaa-odoo:
A translator notified me that one of the sentences in English made no sense, so I'm fixing it to 1) make sense now 2) make it easier for the translator.

Small fix, this page will be depecrated when I have time to make a new one.

Forward to master.

taskid: 3211784


jcs-odoo (approved):
@robodoo r+

PR #3720, opened 2023-03-02
[FIX] contributing/coding_guidelines: remove mention to assets.xml

poma-odoo:
We no longer load assets by xml files, but we use __manifest__.py for that.


AntoineVDV (approved):
@robodoo r+ squash

Note that the guidelines will undergo a full revamp soon but it doesn't hurt fixing the mistakes until then ;)

PR #3749, opened 2023-03-07
[IMP] core: introduce search_fetch() and fetch()

rco-odoo:
Companion of https://github.com/odoo/odoo/pull/114551.


AntoineVDV (approved):
Would it make sense to add a quick `seealso` to these methods in the "Record cache and prefetching" section?

@robodoo delegate+

rco-odoo:
@robodoo r+

Issue #3675, opened 2023-02-26
[16.0] typo in deploy, section 'Serving static files' could be misleading


skygunner:
In the Serving static files section, it reads

The two git repositories for Community and Enterprise have been cloned in /opt/odoo/community and /opt/odoo/enterprise respectively and the [--addons-path](https://www.odoo.com/documentation/16.0/developer/reference/cli.html#cmdoption-odoo-bin-addons-path) is '**/opt/odoo/community/odoo/addons,/opt/community/addons,/opt/odoo/enterprise**'.
The above bold part should be 
/opt/odoo/community/odoo/addons,/opt/**odoo**/community/addons,/opt/odoo/enterprise

Because the given nginx conf example is :

location \@odoo {
    # copy-paste the content of the / location block
}

\# Serve static files right away
location ~ ^/[^/]+/static/.+$ {
    **root /opt/odoo;**
    try_files /community/odoo/addons$uri /community/addons$uri /enterprise$uri @odoo;
    expires 24h;
}

They should be match up otherwise nginx wont gonna able to find that 2nd path files.

skygunner:
Also some improvements can be done to the part 'Serving static files'

The nginx example conf given under serving static files section is:

location @odoo {
    \# copy-paste the content of the / location block
}

\# Serve static files right away
location ~ ^/[^/]+/static/.+$ {
    \# Add Headers for odoo proxy mode
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_set_header X-Real-IP $remote_addr;
    root /usr/lib/python3/dist-packages/odoo/addons;
    try_files $uri @odoo;
    expires 24h;
}

Here I believe the 4 proxy_set_header is not necessary because while try_files $uri find the files, the files would be servered by 
Nginx directly without going to the proxy backend. 
When any file has not been found in the try_files $uri section, the request will go to the backend @odoo where there would be 
necessary proxy_set_header in that section.

Julien00859:
Indeed you are right, I'm on holiday until next week, I'll apply those changes when I'll be back :) Thank you for reporting it!

Julien00859:
[...]

PR #3763, opened 2023-03-08
[FIX] deploy: duplicated proxy_set_headers in serve static block

tsm-odoo:
This commit removes unnecessary `proxy_set_headers` in the serve
static config block.

Indeed, the request will either be served by Nginx in which case
the request won't go to the proxy backend or by the `@odoo` backend
which have its own `proxy_set_headers` set as the user is supposed
to copy/paste the contennt of the `location /` block.

Closes: #3675


jcs-odoo (approved):
@robodoo r+

PR #3779, opened 2023-03-09
[I18N] theme: remove reference to translation

mart-e:
It does not exists and was never present, making the synchronisation crash


mart-e:
robodoo r+

PR #3800, opened 2023-03-10
Add section on rate limits

monty241:
The acceptable use of the external API is not clearly documented.

Based upon a previous PR, the text has been adapted.

Measurements on performance of the external API have been documented on https://forums.invantive.com/t/odoo-sql-driver-performance-improvements-for-very-large-odoo-environments/3232


AntoineVDV:
@robodoo delegate=odony

PR #3846, opened 2023-03-20
[FIX] gmail plugin: on-premise instructions didn't work

yajo:
Instructions for deploying the Gmail Odoo plugin don't work. The instructions found in the upstream project README are for managing only Odoo's own SaaS implementation.

Besides, they are outdated. So it's better to add more serious docs here.

@moduon MT-2612


yajo:
Setting as draft because even with these instructions it doesn't work. Yikes, the docs are deeply broken in this regard. 😞

PR #3866, opened 2023-03-23
[ADD] ecommerce: performance

toaa-odoo:
Adding a doc page on performance monitoring and analysis for eCommerce
websites

taskid: 3246984


chiaraprattico (changes requested):

chiaraprattico (approved):

toaa-odoo:
@robodoo r+

xpl-odoo (commented):
You changed the `ecommerce_management.rst` (containing a toctree only) to` store_management.rst`, whitout renaming its folder. Is there a reason? (to avoid creating redirects, right? ;))

We usually avoid renaming files unless necessary (as far as I remember). WHat's your take on this @AntoineVDV ?

AntoineVDV:
It's okay to rename page files when the old name vs the new content is just wrong. In this case I let you judge, but take for example "CLI" which was renamed to "Command-line interface"; the file can definitely stay `cli.rst`.

However, when a page file is renamed, the related resource directory must be renamed too. This guarantees clarity and easier codeowner rules definition. Renamed pages should also be accompanied by a redirect rule if it was published more than a few days ago.

toaa-odoo:
@xpl-odoo @AntoineVDV I just felt the name was maybe more appropriate than eCommerce. It doesn't matter, though. It's too much trouble for such a simple subjective change. I reverted it to eCommerce

PR #3875, opened 2023-03-24
[IMP] Support - Update What Can I Expect Page

tiku-odoo:


content/services/support/what_can_i_expect.rst
```diff
@@ -33,6 +33,10 @@ What kind of support is included?
 *Odoo Support does not make changes to your production database without your agreement and gives you
 the material and knowledge to do it yourself!*
 
+.. seealso::
+   - :ref:`upgrade/sla`
+   - :doc:`../../administration/maintain/supported_versions`
```
StraubCreative, line 38: For future reference, we can just write the whole path from the parent directory instead of from wherever the file is locally located.
```suggestion
   - :doc:`/administration/maintain/supported_versions`
```

content/services/support/what_can_i_expect.rst
```diff
@@ -54,3 +58,6 @@ You can get this type of support with a `Success Pack <https://www.odoo.com/pric
 With a pack, one of our consultants will analyze the way your business runs and
 tell you how you can get the most out of your Odoo Database. We will handle all
 configurations and coach you on how to use Odoo.
+
+.. seealso::
+   :doc:`../../administration/upgrade`
```
StraubCreative, line 63: Same here.
`doc` tags are ok, however `refs` are better since they're more resilient against broken links 😉 
```suggestion
   :doc:`/administration/upgrade`
```

StraubCreative (approved):
Hi @tiku-odoo 
Thanks for the quick changes. 
Approving, with some comments below regarding `doc` tag paths.

StraubCreative:
@robodoo r+

PR #3876, opened 2023-03-24
[IMP] pos: update fiscal positions page

LoredanaLrpz:
Task ID: 2862506


content/applications/sales/point_of_sale/restaurant/fiscal_position.rst
```diff
@@ -1,43 +1,51 @@
...
-    :align: center
+Configuration
+=============
 
-.. note::
-   You need to create your fiscal positions before using this feature.
+To enable the feature, go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll
+down to the :guilabel:`Accounting` section, and enable :guilabel:`Flexible Taxes`.
+
+Then, set a default fiscal position that should be applied to all sales in POS in the
```
jcs-odoo, line 21: "to all sales in POS" sounds like this setting is applied to the whole app (which would be "Point of Sale" instead, then)

It isn't clear here that this action is applied to the current pos. I know it's the case for all settings, but it wouldn't hurt to add "select the pos" at the beginning.

content/applications/sales/point_of_sale/restaurant/fiscal_position.rst
```diff
@@ -1,43 +1,51 @@
...
+:guilabel:`Allowed` field.
 
-Using fiscal positions
-======================
+.. image:: fiscal_position/flexible-taxes-setting.png
+   :align: center
 
-Once on your *PoS* interface, click on the *Tax* button.
-Now, choose the fiscal position you need for the current order.
+According to the fiscal localization package activated, several fiscal positions are preconfigured
```
jcs-odoo, line 28: I suggest adding a link to "fiscal localization package"

content/applications/sales/point_of_sale/restaurant/fiscal_position.rst
```diff
@@ -1,43 +1,51 @@
...
+and can be set and used in POS. However, to create new ones, :ref:`follow the related documentation
+<fiscal_positions/mapping>`.
```
jcs-odoo, line 30: Avoid "click here" links and try to put meaningful text on it. For example:
```suggestion
and can be set and used in POS. You can also :ref:`create new fiscal positions
<fiscal_positions/mapping>`.
```

content/applications/sales/point_of_sale/restaurant/fiscal_position.rst
```diff
@@ -1,43 +1,51 @@
...
 
-.. image:: fiscal_position/fiscal_position_02.png
-    :align: center
+.. note::
+   If you do not set a fiscal position, the products are sold based on the customer taxes assessed
+   on the product form.
```
jcs-odoo, line 34:
```suggestion
   If you do not set a fiscal position, the tax remains as defined
   on the product form.
```

content/applications/sales/point_of_sale/restaurant/fiscal_position.rst
```diff
@@ -1,43 +1,51 @@
...
-Set up a default fiscal position
-================================
+Use fiscal positions
+====================
 
-If you want to use a default fiscal position, meaning that a preexisting value is always
-automatically assigned, go to :menuselection:`Point of Sale --> Configuration
---> Point of Sale` and enable *Fiscal Position*. Now, choose one to set as the default one.
+Open a :ref:`POS session <pos/start-session>` to use one of the allowed fiscal positions. Then,
+click the :guilabel:`Tax` button next to the **book-shaped** icon and select from the list of
```
jcs-odoo, line 40: You need to select something (to put before "from the list")

content/applications/sales/point_of_sale/restaurant/fiscal_position.rst
jcs-odoo: or "set" instead of "defined" since you seem to use that wording in line 48.

jcs-odoo (changes requested):
Hi @LoredanaLrpz 
Thanks for the PR, this review was made with @zulu-odoo and @auva-odoo who will now come next to your desk and explain everything at length :D 

Cheers!

PR #3887, opened 2023-03-27
[FIX] inventory: fix BoM typo

Felicious:
Replace "baterials" with "materials."


StraubCreative (approved):
Good catch @Felicious 
Approving 🚀

StraubCreative:
@robodoo r+

PR #3898, opened 2023-03-27
[IMP] coding_guidelines: fix python code indents

vin-odoo:
A few python code blocks on the coding guidelines are indented twice (8 spaces instead of 4), which is not correct.


Feyensv:
@robodoo r+

Feyensv (approved):

PR #3900, opened 2023-03-27
[IMP] supported_version: release saas-16.2

AntoineVDV:
X-original-commit: 697cdad097617f20dfa0a3131b0679de7ae5e31e

Follow-up of #3894 that GitHub would not update after force-push.


AntoineVDV:
@robodoo r+

PR #3919, opened 2023-03-28
[IMP] Adyen: additional minimum requirements for users

toaa-odoo:
taskid: 3159712

Adding info on minimum requirements to use Adyen. 
Forward to master.


AntoineVDV (commented):
Although this is true for Adyen as an online payment provider, the support ticket guy complained about Adyen's PoS integration. The note should be added in applications/sales/point_of_sale/payment/adyen.rst too. cc @LoredanaLrpz 

@robodoo delegate+

toaa-odoo:
@robodoo r+

PR #3950, opened 2023-03-29
[IMP] web_editor: powerbox search from description

odoostan:
Adapts documentation to a change in the Powerbox search filters. Since that change, command descriptions are also matched (partially but not fuzzily) to return search results.

See https://github.com/odoo/odoo/pull/115917

task-3188758


content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -124,8 +124,7 @@ There are three ways to filter commands:
    :ref:`command <reference/frontend/odoo_editor/powerbox/command>`'s
    `isDisabled` entry.
 #. The user can filter commands by simply typing text after opening the
-   Powerbox. It will fuzzy-match that text with the names of the categories and
-   commands.
+   Powerbox. It will fuzzy-match that text with the names of the categories and commands, or match that text with the description of the commands.
```
Zinston, line 127: "description" -> "descriptions"
Also, reading this I understand that the matching with descriptions is not fuzzy, is that correct? If so, maybe specify to remove ambiguity (something like "or exact-match"). If not, maybe say "of the categories, commands, and their descriptions".

Zinston: Then I'd say go ahead and call it fuzzy ^^'

odoostan: @Zinston I went ahead and did that.
I've slightly changed the structure of the sentence as well, it seems clearer like this to me. What do you think ?

Zinston: Looks perfect to me, nice job :+1:

@odoo/javascript-doc-review Can you please validate this small PR?

content/developer/reference/frontend/odoo_editor.rst
```diff
@@ -124,8 +124,8 @@ There are three ways to filter commands:
    :ref:`command <reference/frontend/odoo_editor/powerbox/command>`'s
    `isDisabled` entry.
 #. The user can filter commands by simply typing text after opening the
-   Powerbox. It will fuzzy-match that text with the names of the categories and
-   commands.
+   Powerbox. It will fuzzy-match that text with the names, categories and
+   descriptions of the commands.
```
dmo-odoo, line 128: Technically the sentence is wrong because it does _not_ fuzzy match in the description.

odoostan: See discussion above.

content/developer/reference/frontend/odoo_editor.rst
odoostan: @dmo-odoo then do you prefer I call it "exact" match even though it's not an exact match either ?

content/developer/reference/frontend/odoo_editor.rst
dmo-odoo: Oh ok I didn't see that you originally did exactly what I thought you should do but that @Zinston had you change it... ^^' I guess it was quite verbose and fuzzy is not technically as incorrect as I thought since, like you say, we still do some conversion work rather than doing a pure exact match anyway, so I guess I overreacted on this one. It's fine by me.

dmo-odoo:
@robodoo r+

dmo-odoo:
@odoo/javascript-doc-review Can we have an r+ here ?

odoostan:
> I'm sorry, @dmo-odoo. I'm afraid I can't do that.

I so wish he'd called you Dave :eye: 

@dmo-odoo maybe we should wait for https://github.com/odoo/odoo/pull/118322 to pass ?

dmo-odoo:
@odoostan I don't think so. It's just that the documentation repository is a different one and I don't have r+ rights here.

PR #3964, opened 2023-03-29
[REF] web: remove allow_group_range_value

FrancoisGe:
The allow_group_range_value option is used in one arch of a kanban in CRM. We therefore believe that this is not standard behaviour. So we decided to move this logic to the custom view that uses it.


fdardenne (approved):

ged-odoo:
@odoo/be-doc-review looks good to me

PR #3970, opened 2023-03-30
[IMP] tutorials: add warning about mutable global variables

bso-odoo:
The whole concept of multi-tenancy is not really approached within the tutorial.

This commit adds a warning about never using mutable global variables within odoo to seed the idea in the reader's mind.

task-3059110


AntoineVDV (approved):
@robodoo r+

PR #3994, opened 2023-04-03
[IMP] tutorials/js_framework: reword, clarify and improve instructions

AntoineVDV:
See commits for details.

I also took some notes regarding the exercises themselves:

### TUTORIAL DISCOVER JS FRAMEWORK

- all solutions: import from "odoo/owl" everywhere for all exercises
- all solutions: (master only) static properties should be defined in the static block.
- Might be good to require basic knowledge of the prototype chain and the class system.
- 1.1.3: shouldn't we recommend `t-out` rather than `t-esc`?
- 1.3.1: could not guess that the props were not just a string but an object.
    -> maybe first suggest to use <Todo todo="'3. buy milk'"/> then the object.
    -> add a caution for the inner quotes.
- 1.3.2: it's not clear why the object needs to be defined in setup rather than directly in the template
    -> setup is never properly introduced
- 1.4: https://github.com/odoo/owl/blob/master/doc/reference/app.md#dev-mode is not in the ToC of the page.
- 1.6.1: not important but we could recommend classes for the <input/> so that readers don't spend time on that.
- 1.6.4: the id in the solution should start at 1.
- 2.1: need a hint to lead the reader to the recommended solution because `<h1>Dashboard</h1><p>Hello owl</p>` as default slot works just fine.
- 2.2: Let us now use the `action` service <- missing reference documentation
- 2.4.1.solution: The service is not named "awesome_tshirt.statistics"
- 2.4.1: Add a tip to use the `async` key of the Service API
- 2.5.2: the solution wraps loadJS in an array, while the example does not
- 2.5.2: (master only) solution should import from "@web/core/colors/colors" and not "@web/views/graph/colors"

### TUTORIAL MASTER ODOO WEB FRAMEWORK

- all solutions: import from "odoo/owl" everywhere for all exercises
- all solutions: (master only) static properties should be defined in the static block.
- 1.1: (master only) solution should be `<CharField t-if="!props.readonly" t-props="props"/>`
- 1.2: (master only) the exercise could be a bulleted point of 1.1.
- 1.2: (master only) solution should be `<CharField t-if="!props.readonly" t-props="props"/><t t-set="url" t-value="props.record.data[props.name]"/><p class="text-danger" t-else="">MISSING TSHIRT DESIGN</p>`
- 1.3: (master only) solution should be  `<xpath expr="//CheckBox" position="after"><span t-if="props.record.data[props.name]" class="text-danger">Late!</span></xpath>`
- 1.4: the solution defines `fieldDependencies`; what is it for? I didn't need it. Also, it's not documented anywhere (reference, tutorial, ...)
- 1.4: remove the two t-set in the solution
- 1.6: in the solution, the exported view id should be "awesome_tshirt.order_form_view" and not just "order_form_view". adapt the js_class= too.
- 1.6: solution should be  `<xpath expr="//button[hasclass('o_form_button_create')]" position="after"><button t-if="model.root.resId" t-on-click="printLabel" class="btn" t-att-class="isPrintBtnPrimary ? 'btn-primary' : 'btn-secondary'">Print Label</button></xpath>`
- 2.2: solution should be `useService("tshirtService");`
- 2.2: in the solution, StatSystray and systrayItem don't need to be exported.
- 2.2: solution should be   `<div class="dropdown o-dropdown o-dropdown--no-caret"><button class="dropdown-toggle" t-on-click="openNewOrders"><i class="fa fa-shopping-basket"/><span class="badge" t-esc="statistics.nb_new_orders"/></button></div>`
- 2.2: solution should be` res = super().session_info()`
- 2.4: missing reference doc for the command service
- 2.4: (master only) solution should be  `window.open(this.props.record.data[props.name], "_blank");`
- 2.7: as discussed, it's not clear when to use `useCommand`
- 2.7: typo in the solution: "Actvate"
- 2.8: could not guess that extra bootstrap assets might be needed in some situation

### DIFF
--- a/content/developer/tutorials/discover_js_framework/01_owl_components.rst
+++ b/content/developer/tutorials/discover_js_framework/01_owl_components.rst
@@ -251,7 +251,7 @@ way to do this is by using a `callback prop
       be checked if the state `done` is true.
 
       .. tip::
-         QWeb does not create attributes computed with the `t-att` directive it it evaluates to a
+         QWeb does not create attributes computed with the `t-att` directive if it evaluates to a
          falsy value.
 
    #. Add a callback props `toggleState`.
@@ -309,12 +309,12 @@ components. This is useful to factorize the common layout between different part
          <div class="card" style="width: 18rem;">
              <img src="..." class="card-img-top" alt="..." />
              <div class="card-body">
-             <h5 class="card-title">Card title</h5>
-             <p class="card-text">
-                 Some quick example text to build on the card title and make up the bulk
-                 of the card's content.
-             </p>
-             <a href="#" class="btn btn-primary">Go somewhere</a>
+                 <h5 class="card-title">Card title</h5>
+                 <p class="card-text">
+                     Some quick example text to build on the card title and make up the bulk
+                     of the card's content.
+                 </p>
+                 <a href="#" class="btn btn-primary">Go somewhere</a>
              </div>
          </div>
 
--- a/content/developer/reference/frontend/services.rst
+++ b/content/developer/reference/frontend/services.rst
@@ -639,8 +639,8 @@ API
 .. js:function:: rpc(route, params, settings)
 
     :param string route: route targeted by the request
-    :param Object params: parameters sent to the server
-    :param Object settings (optional): request settings (see below)
+    :param Object params: (optional) parameters sent to the server
+    :param Object settings: (optional) request settings (see below)
 
     The ``settings`` object can contain:
 


content/developer/tutorials/discover_js_framework/01_owl_components.rst
```diff
@@ -247,6 +249,11 @@ way to do this is by using a `callback prop
 
    #. Add an input with the attribute :code:`type="checkbox"` before the id of the task, which must
       be checked if the state `done` is true.
+
+      .. tip::
+         QWeb does not create attributes computed with the `t-att` directive it it evaluates to a
+         falsy value.
```
fdardenne, line 255: if* it evaluates ?

content/developer/tutorials/discover_js_framework/01_owl_components.rst
```diff
@@ -284,15 +291,18 @@ The final touch is to let the user delete a todo.
...
 Owl has a powerful `slot <{OWL_PATH}/doc/reference/slots.md>`_ system to allow you to write generic
 components. This is useful to factorize the common layout between different parts of the interface.
 
 .. exercise::
 
-   #. Write a `Card` component using the following Bootstrap HTML structure:
+   #. Insert a new `Card` component between the `Counter` and `Todolist` components. Use the
+      following Bootstrap HTML structure for the card:
 
       .. code-block:: html
```
fdardenne, line 308: I just noticed that there is no indentation in the `card-body` div in the html below. Maybe we can fix it in this PR

fdardenne (approved):
Looks good to me aside two little things 👍
I will put your feedback to exercise solutions in our pipe

PR #4168, opened 2023-04-14
[IMP] odoo_theme: fix menu links

edi-odoo:
When searching for a word, if that word happens to be in the menu, the `display-flex` on the `a` would remove the space around the word because a `span class=highlight` gets added around it.
Fixed by changing `display-flex` on the `a` to `display-block` (but keeping it for `.o_toc_entry_wrapper`)


AntoineVDV (approved):
@robodoo r+

AntoineVDV:
@robodoo retry

PR #4169, opened 2023-04-14
[MOV] field service: improve the structure and merge docs

zulu-odoo:
Currently, there are two pages about onsite interventions in two separate categories. This task aims to merge both pages into a single one and remove the useless categories.

task-3275322


jcs-odoo (approved):
@robodoo r+

PR #4193, opened 2023-04-17
Update 01_architecture.rst

alosiuse:
Translation into Ukrainian


Feyensv:
There is no plan to translate (nor maintain translations) for the developer section of the documentation right now.
And if there ever is, it will be managed on Transifex (as it is already the case for the functional part of the documentation).

PR #4211, opened 2023-04-19
[FIX] website: missing redirection for plausible.rst

jcs-odoo:
The redirection rule for plausible.rst was missing from the forward-port in PR #4107. Because of that, a link in the app leads to a 404 page instead of the appropriate doc page. This commit fixes it.


jcs-odoo:
@robodoo r+

PR #4217, opened 2023-04-20
[IMP] install/maintain: update the wording

avg-odoo:
This PR updates the wording 'upgrade' to 'change'.
An upgrade is switching to a newer version of Odoo, while a migration or change reflects the change of editions or change of hosting type.


xpl-odoo:
@avg-odoo I changed the base branch to `14.0` and put the PR in draft.
However, this page is outdated and I would prefer to update the whole page at once, rather than give the impression it's somehow still correct by updating it now.
cc @d-fence

PR #4221, opened 2023-04-21
[IMP] web: add quick_create_form_view_id attribute

leonardopavanrocha:
In https://github.com/odoo/odoo/pull/114827, a new attribute was added in calendar views to allow passing the id of a form view that is used internally in the calendar module to show a different form view when quick_creating an event.

task-3129030


ged-odoo:
@odoo/doc-review

content/developer/reference/backend/views.rst
```diff
@@ -462,6 +462,10 @@ calendar view are:
     (the field to which this values is saved can be controlled through
     ``rec_name``) and tries to create a new event with just that and the clicked
     event time. Falls back to a full form dialog if the quick creation fails
+``quick_create_form_view_id``
+    allows passing the id of a form view to the calendar view. Internally, this is used
+    in the calendar module to show a custom form view when trying to quick create a
+    calendar event
```
Feyensv, line 468:
```suggestion
    view to open when the user tries to quick create a record.
```
Not sure a reference to an example in the code really adds something here (and those kind of examples do not always age well :D)

Feyensv (approved):
@robodoo r+

AntoineVDV (approved):
@robodoo r+

AntoineVDV:
So close.
image

PR #4226, opened 2023-04-21
[MOV] pos: restructuration of pos toctree

jcs-odoo:
task-3284514


jcs-odoo:
@robodoo r+

jcs-odoo:
@robodoo r-

PR #4228, opened 2023-04-21
Master visual improvments ged

ged-odoo:


ged-odoo:
@odoo/doc-review would you mind having a look at this? Thank you

AntoineVDV (commented):
Hello there, thanks for taking the time to propose improvements to the documentation :100:

I took the time to gather feedback from Design, Content Marketing, and developers about the various changes suggested in your PR:

- Regarding the UI/UX changes, the consensus is to prefer the current design, both from a look and feel perspective and because the changes introduce several regressions and inconsistencies.

  Here are some of the remarks regarding the regressions: removing the menu icons leads to a weird look and behavior for the top-level menu items; the menu button is no longer accessible on mobile; the spacing of the dropdown button in the header is not good; the "Edit on Github" button is not aligned anymore; double-clicking makes the menu disappear on some page (search.html...); various responsiveness issues (hard to describe)...
  
  The "merged" header is, however, an interesting take; if the menu button was moved underneath the logo and [the alignment issues fixed](https://media.discordapp.net/attachments/689044806181453866/1102946008423481364/Screenshot_2023-05-02_at_15.10.16.png), it would allow removing the grey bar on desktop. :+1:
- Regarding the top-level page structure, there surely is room for improvement. The final structure is still being debated within the Content team but the "Docs" from "User Docs" could already be removed to avoid suffixing "Documentation" everywhere and be consistent with "Developer". Even though I personally like "Deployment", I was told it's not entirely coherent with the actual content of that directory. I assume "Install and Maintain" will be split into two directories later on.
- About the purely technical changes addressing the flickering issues, I still have some flickering on my end, but maybe less (?) If so, it'd be good to merge the related commit(s). How is it done in Odoo to prevent that?

PR #4304, opened 2023-04-29
[REV] accounting: add Hong Kong localisations

tong-odoo:
This reverts commit 29b3313d85c5be33a2b804bd6eab477ae58f4c40. The reason is this commit should deploy together with odoo/odoo#120050.


AntoineVDV (approved):
@robodoo r+

AntoineVDV:
@tong-odoo I didn't request @fw-bot to ignore `master` because I don't know if you the module will be merged before the `saas-16.3` freeze. I suggest merging the reverts in 16.1 and 16.2, and keeping the master PR open until you see in which version your module lands.

tong-odoo:
@AntoineVDV I think the previous commit haven't deployed to 16.1 and the other newer version yet. So only 16.0 should be enough.

PR #4391, opened 2023-05-10
[FIX] requirements: bump libsass to match odoo

d-fence:
In Odoo, the libsass version was bumped to 0.20.1 in order to match the Jammy version. As a consequence, in the Docker images on the runbot this strange behavior was observed during the image build:

* the debian package is installed -> 0.20.1
* the documentation requirements are installed -> 0.18.0
* the odoo requirements are installed -> 0.20.1

This is useless and moreover, the final version used is the one from pip instead of the Debian package.


AntoineVDV (approved):
@robodoo r+

PR #4428, opened 2023-05-15
[REM] accounting: remove page about purchase receipts

jcs-odoo:
Sale and purchase receipts are deprecated and aren't really useful anymore? However, they are left in the app for corner cases and to avoid upgrade issues. Since this page is already deprecated, it is best to remove it.


jcs-odoo:
@robodoo r+

PR #4441, opened 2023-05-16
Update serial_numbers.rst

StephanieSpeiserSAMSA-IT:
added more details about connection PoS & inventory stock
example stock and pos


LoredanaLrpz:
Hello @StephanieSpeiserSAMSA-IT, thank you for reaching out and for your work. Unfortunately, we won't promote the use of a third-party Odoo app to update inventory level since it's already done by default using only Odoo Point of Sale.

Have a great day, and thanks again

PR #4446, opened 2023-05-17
[FIX] accounting: mark salt edge bank sync provider available worldwide

jcs-odoo:
task-3329454
fw-port of #4444


jcs-odoo:
@robodoo r+

PR #4537, opened 2023-05-25
[IMP] accounting: reports: add missing subformulas for aggregation engine

oco-odoo:
https://github.com/odoo/enterprise/commit/f39862263c7e2332468a69736589a35d66d87b16 introduced two new subformulas for aggregations, but they were still missing in the doc.


content/applications/finance/accounting/reporting/customize.rst
```diff
@@ -160,6 +160,16 @@ expression's **label** (ex. **code.label**).
    The value of the arithmetic expression will be returned only if it is strictly between the
    provided bounds. Otherwise, it will be brought back to the closest bound.
 
+`if_other_expr_above(LINE_CODE.EXPRESSION_LABEL, CUR(amount))`
+   The value of the arithmetic expression will be returned only if the value of the expression denoted
```
toaa-odoo, line 164:
```suggestion
   The value of the arithmetic expression will be returned only if the value of the expression denoted
```
It seems "denoted" is two characters past the character limit. Your technical knowledge is much better than mine, so I'm not sure if there's a specific reason or if it's just a mistake :)

I'll approve it, but if it is indeed a mistake, could you please fix it?

toaa-odoo (approved):

jcs-odoo (approved):
@robodoo r+

PR #4603, opened 2023-06-02
[IMP] accounting: OSS note removal

toaa-odoo:
Removal of a note in 16 that only applies to versions below it.

task-3305037


jcs-odoo (approved):
@robodoo r+

PR #4605, opened 2023-06-02
[FIX] install: minimal python version is 3.8

Julien00859:
See odoo/odoo@44d60e3


xpl-odoo (approved):
@robodoo delegate+

Julien00859:
@robodoo r+

rdeodoo:
> See https://github.com/odoo/odoo/commit/44d60e3

It's a dead commit, do you know which commit were you referring to?

PR #4640, opened 2023-06-07
odoo.conf is located in /etc/odoo.conf

mbendel-andagon:
Describe path of odoo.conf as /etc/odoo/odoo.conf instead of /etc/odoo.conf


d-fence:
Thanks for you contribution, the issue that you highlighted will be handled in #4785 as the config file location depends on how Odoo is started.

PR #4674, opened 2023-06-09
[IMP] odoo_theme: toctrees-l1,l2 design improvement

mano-odoo:
Issue 1: 
-------
The toctrees in the documentation are visually inconsistent because of
the content. Sometimes we have titles that are also links and visually
collide with the rest of the toc making it hard to read. This applies
to toctree-l1 and toctree-l2.

Fix 1:
-------
When we have a link and title toctree-l1 or toctree-l2 in a list 
containing other nested toc, we apply another styling displaying it 
with the same color as a title, but with an icon and hover behavior 
indicating that it is a link. (This commit also changes the direction of
the i-link icon to make it standard). 

Issue 2:
-------
When we only have toctree-l1 links without nested toc the toctree
is uselessly taking up a lot of space. 

Fix: 2
-------

In this scenario we replace the toctree-l1 class with toctree-l2 to
make them appear as a discreet list.

task-3138525
task-3138563

part of 

task-3059178


edi-odoo (approved):
LGTM! @AntoineVDV could you check the JS please ?

AntoineVDV:
@robodoo retry

Issue #4767, opened 2023-06-16
[16] Theme prefix


imanie383:
By default, the prefix for the themes is "theme_" but in the odoo tutorial 16 it says that the prefix is "website_"

https://www.odoo.com/documentation/16.0/developer/howtos/website_themes/theming.html#global
![image](https://github.com/odoo/documentation/assets/35231827/4ceab097-edf8-45dd-a677-d38a6dfbcb4a)

but in the code the prefix evaluated in odoo 16 is `theme_`
https://github.com/odoo/odoo/blob/a65dda25b95fa904879ca8e7b3309f00ed21fbbf/addons/website/models/ir_module_module.py#L83

And the default theme is named theme_default
https://github.com/odoo/odoo/tree/16.0/addons/theme_default
![image](https://github.com/odoo/documentation/assets/35231827/57331747-f619-4dc9-b2ca-785d0e60921d)

xpl-odoo:
Thanks for the feedback @imanie383 !
@CVDE-odoo could you have a look?

CVDE-odoo:
Thank you for the feedback @imanie383 @xpl-odoo 

This is because we are creating a module, not an actual theme.
We prefix the module with `website_` because it's linked to the website app.

CVDE-odoo:
I closed this issue. Don't hesitate to reopen it if needed.

CVDE-odoo:
@imanie383 Yes, the themes are modules too. In this case, we can't use "themes" modules as they contain python and won't be usable on Saas. So we have to create "classic" modules to make it work, and we can't use the "theme_" prefix.

imanie383:
FYI: the theme can have Python by example the post_coy

PR #4768, opened 2023-06-16
[FIX] developer/howto: correct theme prefix

niyasraphy:
before this commit, in the theme doc, expected
prefix for the theme module is written as website_ , but it is theme_

after this commit, the website_ is changed to theme_

issue: https://github.com/odoo/documentation/issues/4767

### DIFF
--- a/content/developer/howtos/website_themes/theming.rst
+++ b/content/developer/howtos/website_themes/theming.rst
@@ -29,7 +29,7 @@ The first step is to create a new directory.
    website_airproof
 
 .. note::
-   Prefix it with `website_` and use only lowercase ASCII alphanumeric characters and underscores.
+   Prefix it with `theme_` and use only lowercase ASCII alphanumeric characters and underscores.
 
    In this documentation, we will use **Airproof** (a fictional project) as an example.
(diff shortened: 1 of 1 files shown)


AntoineVDV:
This sounds about right, but I'm surprised that the example below uses the module name `website_airproof`. Shouldn't that be corrected too @CVDE-odoo?

CVDE-odoo:
Thank you for the feedback @AntoineVDV @niyasraphy 

This is because we are creating a module, not an actual theme.
We prefix the module with `website_` because it's linked to the website app.

AntoineVDV:
@CVDE-odoo I'm not sure to understand the nuance. In both case you need to create a module, no?

CVDE-odoo:
@AntoineVDV You are right, themes are modules too but they are more restricted and we can't work with them on saas at the moment. This is the reason why we create a "classic" module. But this might change in the future.

AntoineVDV:
@niyasraphy As per @CVDE-odoo's comment, the doc should still recommend using the prefix `website_`, but it could be clearer why is that.

PR #4823, opened 2023-06-21
[IMP] developer/views: add optional attribute for fields of list views

bso-odoo:
The attributes of fields of list views do not mention the `optional` attribute while it is used by chapter 12 of the R&D Training.

This commit describes this `optional` attribute.

task-3380953


AntoineVDV (approved):
@robodoo delegate+

PR #4876, opened 2023-06-26
[IMP] accounting: AU rebrands Keypay to Employment Hero

tong-odoo:
Impacted Versions:

- 16.0 and above

task-3250954


toaa-odoo (approved):

toaa-odoo:
> Impacted Versions:
> 
> * 16.0 or above
> 
> [task-3250954](https://www.odoo.com/web#view_type=form&model=project.task&id=3250954)

All good to me, and I guess you meant 16.0 *and* above ;)

jcs-odoo (changes requested):
Hi @tong-odoo and @toaa-odoo 

Content-wise, I think the text needs to be adapted as well since we have to retrieve the codes from the employmenthero portal, and not from keypay anymore. Since it's still named keypay in Odoo's UI, we could also add an explication note for the readers, to avoid confusion.

Also, the images haven't been compressed with pngquant yet.

Thanks :)

jcs-odoo (approved):
@robodoo r+

PR #4878, opened 2023-06-26
[IMP] Inventory: updated images and removed edit steps

larm-odoo:
Making targeted edits for 15 - needed to update some screen shots and eliminate some "edit" and "save" steps.


hojo-odoo (approved):
Hey @larm-odoo,

These changes look good for the most part. The only thing I noticed is that you need to have a second route selected in addition to MTO on the acoustic-block-screen-replenish.png image. 

Also, just so you know, I'm in the middle of writing a series of docs that will replace this replenishment strategies doc. I think it's fine to push these changes, but this doc will be deleted entirely after I'm finished with those other docs.

Thanks,
hojo

larm-odoo:
Thanks @hojo-odoo ! I also realized the image in question- but I had only the MTO selected for 14, and I wanted to keep them consistent (but I see your point, and I'll fix that now). Also, I totally understand- I started this doc way back when I was in Inventory, and they took a while to get reviewed and posted, so I must have started all this before the big re-organization you started.

larm-odoo:
Hi @samueljlieber - I tagged you since I am supposed to have you review this before Zac, but if I don' need to please let me know. I only changed/removed 2 sentences, and updated/changed a few images- very minimal edits.

StraubCreative (changes requested):
Hi @larm-odoo
The one image `replenishment-report-dashboard.png` seems to not be in HD (the dashboard and it's column seem too wide). 

Can you resize the browser like in [Step 2 of the instructional GIF](content/applications/inventory_and_mrp/inventory/management/products/strategies/replenishment-report-dashboard.png) please?

After that we should be good for merge, thanks!

larm-odoo:
Hi @StraubCreative - I checked with Sam I was doing the right steps again, so I tried to eliminate the unnecessary columns, make them as close as I could, zoomed in, and then did the normal capture/compress. It's now clearer that it was previously- LMK if this is good!

StraubCreative (approved):
@fw-bot ignore
@robodoo r+

PR #4898, opened 2023-06-27
[FIX] howto/rdtraining: ch.13 adapt reference to salesperson field

bso-odoo:
In chapter 13 the `property_ids` field is defined as the inverse of a `user_id` field. But in chapter 8 when the salesperson field is added, no such name is given to it.

This commit adapts the description to reference the general concept of the field used to link to the salesperson instead of an actual field identifier.

task-3381757


AntoineVDV (approved):
@robodoo r+

PR #4899, opened 2023-06-27
useAssets has been deprecated in 16.0

poma-odoo:
useAssets was introduced in 15.0 but removed in 16.0 in https://github.com/odoo/odoo/commit/8603d13d1961e0146df19aad65936c84eb6071ac 

Now there is an easier way to do it


sdegueldre:
That's not the case, Widget is not a class that cannot be extended in this fashion, it's a legacy feature that's extended using `Widget.extend`. The proper way to load assets in 16.0 for components is to use `onWillStart` in combination with `loadJS/loadCSS/loadBundle`

PR #5037, opened 2023-07-06
[FIX] odoo_theme/switchers: fallback to the original target URL

AntoineVDV:
Attempting to switch from one version to another when navigating <old_version>/applications/hr/employees.html led to the user being redirected to /applications/documentation/<new_version>/hr/employees.html because `hr` was interpreted as a version string, and thus the generated target URLs of the version switcher were all wrong.

This commit ensures the redirection now falls back onto the original target URL when all the generated ones point to a 404.

### DIFF
--- a/extensions/odoo_theme/static/js/switchers.js
+++ b/extensions/odoo_theme/static/js/switchers.js
@@ -19,7 +19,7 @@
                     if (!targetUrl.startsWith('/')) {  // Don't test for valid URLs if in localhost.
                         event.preventDefault();
                         const fallbackUrls = await _generateFallbackUrls(targetUrl);
-                        const fallbackUrl = await _getFirstValidUrl(fallbackUrls);
+                        const fallbackUrl = await _getFirstValidUrl(fallbackUrls) ?? targetUrl;
                         window.location.href = fallbackUrl;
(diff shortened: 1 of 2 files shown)


mart-e (commented):
That works but it's fixing the symptoms. We keep making 4 useless pings for a page that does exists. How about
```diff
diff --git a/extensions/odoo_theme/static/js/utils.js b/extensions/odoo_theme/static/js/utils.js
index d27facd55..cf6d613ce 100644
--- a/extensions/odoo_theme/static/js/utils.js
+++ b/extensions/odoo_theme/static/js/utils.js
@@ -134,7 +134,7 @@ const _generateFallbackUrls = async (targetUrl) => {
     const [urlBase, version, language, originalPathParts] = _deconstructUrl(targetUrlObject);
 
     // Generate the fallback URLs.
-    const fallbackUrls = [];
+    const fallbackUrls = [targetUrl];
     for (let i = originalPathParts.length; i >= 0; i--) {
         const fallbackPathParts = originalPathParts.slice(0, i);
 ```

mart-e (approved):

AntoineVDV:
@robodoo r+

PR #5078, opened 2023-07-11
"Domains" should not be under "View inheritance"

SirTakobi:
Why is "Domain" under "View inheritance"?
Could not find an answer, so I proposed this change.

Let me know if there is some reason that I'm not seeing why "Domains" should be under "View inheritance".

Thanks!

### DIFF
--- a/content/developer/tutorials/backend.rst
+++ b/content/developer/tutorials/backend.rst
@@ -614,7 +614,7 @@ instead of a single view its ``arch`` field is composed of any number of
    * Using view inheritance, display this fields in the partner form view
 
 Domains
-~~~~~~~
+=======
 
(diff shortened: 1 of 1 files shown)


AntoineVDV:
Hello @SirTakobi, thanks for taking the time to suggest an improvement. As you may know already, this tutorial is deprecated and kept only for archival reasons. Therefore, we don't merge any new PR targeting this particular file ;)

SirTakobi:
> Hello @SirTakobi, thanks for taking the time to suggest an improvement. As you may know already, this tutorial is deprecated and kept only for archival reasons. Therefore, we don't merge any new PR targeting this particular file ;)

Oh sorry then, just now I noticed the red box on top of https://www.odoo.com/documentation/16.0/developer/tutorials/backend.html

PR #5081, opened 2023-07-11
[IMP] payments: add section about ACH payments

xpl-odoo:
task-3303591


toaa-odoo (approved):
Thanks @xpl-odoo for your work! I allowed myself to add "Edit" and "Save" instructions, since we're still in 15.0

xpl-odoo:
@robodoo r+
ACD checked the content.

PR #3242, opened 2022-12-29
[IMP] inventory: update automated inventory valuation

brse-odoo:
Issue #2842 

Addressed concerns listed in Issue 2842.

- Fixed 100th character line breaks
- Fixed indentations
- Fixed typos and improved grammar
- Added abbreviation tags to POs

PR #5088, opened 2023-07-12
[IMP] inventory: add tables to inv val doc

brse-odoo:
This PR is an [IMP] targeting the `inventory_valuation_config` doc in Odoo's Inventory documentation.

To be fwd-ported from 14.0 to 15.0 and 16.0

Per inquiries from multiple people:
- Added accounting tables to two sections to replace JS tables removed in earlier [PR #3242](https://github.com/odoo/documentation/pull/3242)
- Made a few other small fixes to address accuracy and .rst


StraubCreative:
@brse-odoo please see Fel's comments above, address as needed, and move to next stage via @odoo/crm-doc-review if the content here is still relevant. Thanks!

PR #5094, opened 2023-07-12
[FIX] Remove reference to assets_qweb in assets.rst

itdurpro:
Reflects the fact that assets_qweb bundle is no longer used in 16.0


sdegueldre (approved):

AntoineVDV (approved):
@robodoo r+

PR #5131, opened 2023-07-15
[IMP] tutorials/getting_started: disambiguate example model name

chyde-clearwater:
I removed the dot notation in the  "_name" variable for the example model. AKA, "test.model" is now named "test_model". 

Dot notation implies a data structure relationship between a <model> and it's parent (or class etc) <test> (like test.model() or Test.model).  There is no relationship like this between a <test> and <model> - it's just one thing: a <test model>.

The seeming "benefit" of the ORM translating dots to underscores doesn't justify this naming convention.  Yes, I know it's a string, but still, this seems confusing for newbies and adds unnecessary complexity (why name the same thing differently in different places?)

### DIFF
--- a/content/developer/tutorials/getting_started/04_basicmodel.rst
+++ b/content/developer/tutorials/getting_started/04_basicmodel.rst
@@ -57,12 +57,11 @@ model::
     from odoo import models
 
     class TestModel(models.Model):
-        _name = "test.model"
+        _name = "test_model"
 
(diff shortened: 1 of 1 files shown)


AntoineVDV (approved):
Hello @chyde-clearwater, this is indeed clearer for new Odoo developers. They'll figure out the dot notation later. Tanks for the contribution :)

@robodoo r+

PR #5132, opened 2023-07-17
[IMP] core: improve manifest's version documentation

aj-fuentes:
Odoo will strip the currently running version from the module version in
the manifest. This behaviour is undocumented and may cause issues for
upgrades.


Julien00859:
When we *do* update it, it is to implement hacks: odoo/odoo#86217

aj-fuentes:
@Julien00859 I've updated the text. I kept a _tangential_ reference to semver because it's the simplest way to make it clear what a version is used for in Odoo.

aj-fuentes:
Does it look good to merge?

PR #5234, opened 2023-07-31
[IMP] developer: document new parameter

mart-e:
Doc of odoo/odoo#129977

### DIFF
--- a/content/developer/reference/backend/mixins.rst
+++ b/content/developer/reference/backend/mixins.rst
@@ -88,6 +88,7 @@ to manage followers on your record:
         parent partners to the message in case of private discussion
     :param list(tuple(str,str)) attachments: list of attachment tuples in the form
         ``(name,content)``, where content is NOT base64 encoded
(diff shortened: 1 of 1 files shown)


content/developer/reference/backend/mixins.rst
```diff
@@ -88,6 +88,7 @@ to manage followers on your record:
         parent partners to the message in case of private discussion
     :param list(tuple(str,str)) attachments: list of attachment tuples in the form
         ``(name,content)``, where content is NOT base64 encoded
+    :param body_is_html: indicates `body` should be threated as HTML even if `str`
```
AntoineVDV, line 91:
```suggestion
    :param bool body_is_html: indicates whether `body` should be treated as HTML, even if `str`.
```

AntoineVDV (commented):
@robodoo delegate+

AntoineVDV:
This method could be `autodoc`umented to avoid such PRs and eventual discrepancies.

mart-e:
robodoo r+

PR #5237, opened 2023-08-01
[IMP] reference/user_interface: reformat and clarify views reference

AntoineVDV:
In particular, the following changes are made:
- Use the `class` and `attribute` admonitions along with custom attributes to define classes, views' root attributes, views' components, and attribute values. This allows re-using the responsive design that was made for reference lists, and getting rid of the previous implementation that relied on class attributes, which were not intended for this usage and reduce readability while hindering further contributions due to a lack of flexibility (no admonitions, sub-attributes...)  
  
  image
- Use definition lists to define view types to allow for clearer and longer descriptions.
  
  image
- Rewrite and restructure the explanations when there is a lack clarity.
- Extract duplicated content to included RST files.
- Display SVG images into dedicated admonitions.
- Fix RST and English mistakes.
- Rename `view_architecture` to `view_architectures`, as it lists all existing architectures and doesn't describe the "architecture of a view".
- Replace underscores with hyphens in image file names to improve SEO.

task-3458320


content/developer/reference/user_interface/view_architectures.rst
```diff
@@ -0,0 +1,3710 @@
...
+   .. list-table::
+      :class: o-showcase-table
+
+      * - .. image:: view_architectures/search_filter.svg
+             :align: center
+
+      * - .. code-block:: xml
+
+             <search>
+                 <filter string="My Custom Name" domain="[('name', 'ilike', 'AAA')]"/>
```
Feyensv, line 2007: Doesn't strictly match the image above

content/developer/reference/user_interface/view_architectures.rst
```diff
@@ -0,0 +1,3710 @@
...
+
+   The current context propagated from either the action that opens the kanban view, or the one2many
+   or many2many field that embeds the kanban view in a form view.
+
+   :type: str
+
+.. attribute:: user_context
+   :noindex:
+
+   :type: str
```
Feyensv, line 2633: I'd avoid mentioning that one, it's not used anywhere (and maybe included in the standard context)

Feyensv (commented):
Very nice work already :+1:

content/developer/reference/user_interface/view_architectures.rst
AntoineVDV: :shrug:

AntoineVDV:
@robodoo r+

PR #5274, opened 2023-08-04
[IMP] test/rst_style, Makefile: add optional review checkers

samueljlieber:
This PR improves the functionality of the Makefile by adding the `review` target as well as adding an optional rst_style check for detecting early line breaks. 

The `make review` target enables all the optional checks to be ran, prompting the user for a path and maximum line length.

task-2801043

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,5 +84,13 @@ static: $(HTML_BUILD_DIR)/_static/style.css
 	cp -r extensions/odoo_theme/static/* $(HTML_BUILD_DIR)/_static/
 	cp -r static/* $(HTML_BUILD_DIR)/_static/
 
+# Called by runbot for the ci/documentation_guideline check.
 test:
 	@python tests/main.py $(SOURCE_DIR)/administration $(SOURCE_DIR)/applications $(SOURCE_DIR)/contributing $(SOURCE_DIR)/developer $(SOURCE_DIR)/services redirects
+
+# Similar as `test`, but called only manually by content reviewers to trigger extra checks.
+review:
+	@read -p "Enter content path: " path; read -p "Enter max line length (default: 100): " line_length; \
+	if [ -z "$$path" ]; then echo "Error: Path cannot be empty"; exit 1; fi; \
+	if [ -z "$$line_length" ]; then line_length=100; fi; \
+	python tests/main.py -e line-too-long -e early-line-breaks --max-line-length=$$line_length $(SOURCE_DIR)/$$path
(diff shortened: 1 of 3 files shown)


StraubCreative (approved):
Approving after discussion and seeing a demo.
I think this is a really great add to our test kit.

Passing to @doc-review for a more thorough look.
Thanks in advance @AntoineVDV @Feyensv 🙏

AntoineVDV (approved):
Hello, I pushed a review commit that you can squash if you're okay with it.

Some parts of the logic for "early breaks" are far-fetched, IMHO, so I didn't bother rewriting them and only reformatted+refactored what I could:
- added `100` as the default value for the line length;
- (re-)wrote some comments to explain better what `make review` is about;
- renamed a few variables and functions;
- simplified `is_valid_line`;
- fixed the `yield` suggesting the wrong target line.

 Otherwise, it seems to work overall, even though some false positives remain.

@robodoo delegate+

samueljlieber:
Hi @AntoineVDV thank you for the improvements, I made two small corrections in https://github.com/odoo/documentation/pull/5274/commits/455e77029ea4ab75feb3c41449746099dd569b02:
1. the `$` added at the end of the REGEX was causing many of the false positives, removing this character stopped the false positives from my testing.
2. I believe the `yield` suggesting the target line number was correct: 
Screenshot 2023-08-07 at 4 11 32 PM
Screenshot 2023-08-07 at 4 11 59 PM

I can squash and merge with your 👍 – thank you! 🙏

AntoineVDV:
> Hi @AntoineVDV thank you for the improvements, I made two small corrections in [455e770](https://github.com/odoo/documentation/commit/455e77029ea4ab75feb3c41449746099dd569b02):
> 
> 1. the `$` added at the end of the REGEX was causing many of the false positives, removing this character stopped the false positives from my testing.
> 2. I believe the `yield` suggesting the target line number was correct:
>    Screenshot 2023-08-07 at 4 11 32 PM
>    Screenshot 2023-08-07 at 4 11 59 PM
> 
> I can squash and merge with your +1 – thank you! pray

Oh yes, indeed. I updated the `lno` after adding the `$` (because it's usually good practice to accommodate the code to a regex ending with `$`, and not the other way around, but it's fine here) to the regex, which spans several lines, hence why it looked wrong to me.

I will squash and r+ now so we can rebase https://github.com/odoo/documentation/pull/4870 today :)

AntoineVDV:
@robodoo r+

PR #5304, opened 2023-08-07
[IMP] supported_versions: release saas-16.4

AntoineVDV:
This commit also moves all versions from 8.0 to 12.0 from the table to
the "older versions" section.


AntoineVDV:
@robodoo r+

PR #5341, opened 2023-08-07
[MOV] Payroll: moved configuration doc to main payroll so there is le…

larm-odoo:
Just moving the config doc so that it is one level closer - it is in the main payroll doc now.

PR #5347, opened 2023-08-07
[IMP] payroll: updated new work entry type image and info for 15

larm-odoo:
Updated payroll doc for v 15- new work entry type image is different and has more items to enter.

### This pr should not be FWP past `saas-15.4`


larm-odoo:
This is not to be completed until PR #5341 is done!

content/applications/hr/payroll/configuration.rst
```diff
@@ -67,17 +67,9 @@ form:
...
-- Check boxes: If any of the items in the list applies to the work entry, check off the box by
-  clicking it. If :guilabel:`Time Off` is checked off, a :guilabel:`Time Off Type` field appears.
-  This field has a drop-down menu to select the specific type of time off, or a new type of time off
-  can be entered.
-
-.. image:: configuration/new-work-entry.png
-   :align: center
-   :alt: New work entry type form.
-
+- :guilabel:`DMFA Code`: This code is used to identify DMFA entries on a corresponding DMFA report,
```
jero-odoo, line 70: I am assuming that if a company is using this code they would have a better understanding of what DMFA is/refers to, so it probably doesn't need to be defined further, right?

jero-odoo (changes requested):

content/applications/hr/payroll/configuration.rst
larm-odoo: Correct- it's Belgian-only (i put that in the doc) so I don't think so. It sounds close to a social security number- we don't have to define that for us US people, so I figured the same for the Belgians =). Thanks for the review BTW!

content/applications/hr/payroll.rst
```diff
@@ -98,6 +87,17 @@ form:
    entry remains 5.5 hours. If :guilabel:`Rounding` is set to :guilabel:`Half Day`, the entry is
    changed to 4 hours. If it is set to :guilabel:`Day`, it is changed to 8 hours.
 
+- :guilabel:`Unpaid in Structures Types`: If the work entry is for unpaid work, specify which pay
+  structure the work entry can apply to from the drop-down menu.
+- Check boxes: If any of the items in the list applies to the work entry, check off the box by
```
jero-odoo, line 92:
```suggestion
- Check boxes: if any of the items in the list apply to the work entry, check off the box by
```

tiku-odoo (commented):
@larm-odoo 

There are some basic capitalization mistakes that need to be corrected. 

Also, I would like to see some highlights on the images you replaced for the fields you mentioned. There are a lot of broad db screenshots in this doc and I think it will help narrow focus if possible.

Tag me again and I will approve --> Looking forward to seeing this merged

Tiku

content/applications/hr/payroll.rst
```diff
@@ -74,17 +71,9 @@ form:
...
-  This field has a drop-down menu to select the specific type of time off, or a new type of time off
-  can be entered.
-
-.. image:: payroll/new-work-entry.png
-   :align: center
-   :alt: New work entry type form.
-
+- :guilabel:`DMFA Code`: this code is used to identify DMFA entries on a corresponding DMFA report,
+  and is for Belgian-based companies only.
+- :guilabel:`Color`: select a color for the particular work entry type.
```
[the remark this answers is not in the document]
larm-odoo: Hi Tim- Jess caught these because per a review from ZST in another doc (one of hers), he prefers lower-case letters after a colon when making lists like these. I thought we caught all of them, but I'll check to be sure they have all been changed. But that is the reason behind this formatting change. I don't think you could easily see it, but in the thread above you can see Jess made the suggestions for those changes. Since this is a personal preference, and it could be capitalized or be lower-case, since we saw that they had been corrected in other docs to be lower-case, that's why we made these changes.

larm-odoo:
Hi @tiku-odoo - thanks for the review. I updated all the generic images with data filled images from Runbot , so those should be more informative now. Also, I made sure the cases were all correct- I wrote a comment on your review to explain further, but I still didn't catch them all originally. This should be good to go now!

tiku-odoo (approved):
@larm-odoo 

Ahh okay- that makes perfect sense then, I think you're correct in keeping things consistent (all upper case like before and moving forward--> all lower). I think consistency is most important, it can look sloppy otherwise imo. You can double-check with zst/meng but I would leave a capital, could be a wormhole. Sorry if the comment(s) seemed harsh; I didn't realize the back story.... 

Nice work!

samueljlieber (approved):
Hi @larm-odoo this PR looks good to me! Nice job 🙂

@StraubCreative this PR is good to go, and should **NOT** be FWP, `15.0` only.

StraubCreative:
@fw-bot up to saas-15.4

StraubCreative:
@fw-bot up to saas-15.3

StraubCreative:
@fw-bot up to saas-15.2

StraubCreative:
@robodoo retry

Issue #5359, opened 2023-08-08
Microsoft Azure sign-in authentication - Access Denied


hedshefer:
Hi,

We have followed the Odoo documentation:
https://www.odoo.com/documentation/16.0/applications/general/auth/azure.html#microsoft-azure-sign-in-authentication

Then, we got 'access denied' error and the following error in the log:

 odoo.addons.auth_oauth.controllers.main: OAuth2: invalid_request
Traceback (most recent call last):
File "/home/odoo/src/odoo/addons/auth_oauth/controllers/main.py", line 134, in signin
db, login, key = env['res.users'].sudo().auth_oauth(provider, kw)
File "/home/odoo/src/odoo/addons/auth_oauth/models/res_users.py", line 124, in auth_oauth
validation = self._auth_oauth_validate(provider, access_token)
File "/home/odoo/src/odoo/addons/auth_oauth/models/res_users.py", line 49, in _auth_oauth_validate
raise Exception(validation['error'])
Exception: invalid_request

Odoo Ticket #3458265
@tiku-odoo @StraubCreative

StraubCreative:
Hi @hedshefer 
We're looking into it, thanks 👍

tiku-odoo:
@hedshefer 

Thanks for reaching out. I also see you have a support ticket open. 

Without testing on your database, I cannot see what you have configured. Just a few things to consider while the support team and I work on your issue: 

**For the Supported account types this selection may vary based on your end use and Microsoft account type. We will make a warning note in the doc clarifying this (for both internal users and portal users). What is your login end goal (organizational users or portal customers)? What type of Microsoft account do you have?**

**Choose "Personal Microsoft accounts only" if the target audience is meant for portal users. Choose "Accounts in this organizational directory only (Default Directory only - Single tenant)" if the target audience is company users.**

Have you input the system parameter in Odoo yet? This is often the case for an access denied error. 

> Odoo System Parameter
> 
> First activate the developer mode, and then go to Settings ‣ Technical ‣ System Parameters.
> 
> Click Create and on the new/blank form that appears, add the following system parameter auth_oauth.authorization_header to the Key field, and set the Value to 1. Then click Save to finish.


Your question in Odoo Ticket #3458265

> 
> In addition - I will be happy to understand the following instruction regarding Odoo.sh hosting:
> 
> Warning
> 
[...]

PR #5421, opened 2023-08-08
[IMP] Misc Oauth Azure Clarification

tiku-odoo:
Clarification on supported account types resolves https://github.com/odoo/documentation/issues/5359


StraubCreative (approved):
Looks good, thanks for the quick fix @tiku-odoo 

@robodoo r+

PR #5460, opened 2023-08-11
[IMP] remove references to owl template tagging

sdegueldre:
In odoo/odoo#130467 the necessity to tag owl templates with an owl="1" attribute has been removed, update the docs to reflect that.


Feyensv (approved):
@robodoo r+

PR #5485, opened 2023-08-14
[FIX] deploy: set Content-Security-Policy on static

Julien00859:
The Content-Security-Policy[^1] http header was only set on the response generated by controllers but it was missing from the `/<module>/static/` route.

It is not strictly necessary to set that header on the responses comming from that routes as it is not possible to add new static files or edit existing ones via the interface (not even as admin). Only the developers and system administrator can access those files.

It is also worth mentionning that using the Odoo internal web server to deliver static files is suboptimal. Outside of a dev environment, those files will typically be delivered via a web server[^2] and sysadmins should configure their web server to set the CSP header on static images.

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
[^2]: https://www.odoo.com/documentation/master/administration/install/deploy.html#serving-static-files-and-attachments

### DIFF
--- a/content/administration/install/deploy.rst
+++ b/content/administration/install/deploy.rst
@@ -439,51 +439,71 @@ Odoo static files are located in each module's :file:`static/` folder, so static
 by intercepting all requests to :samp:`/{MODULE}/static/{FILE}`, and looking up the right module
 (and file) in the various addons paths.
 
+It is recommended to set the ``Content-Security-Policy: default-src 'none'`` header on all images
+delivered by the web server. It is not strictly necessary as users cannot modify/inject content
+inside of modules' :file:`static/` folder and existing images are final (they do not fetch new
+resources by themselves). However, it is good practice.
+
+Using the above NGINX (https) configuration, the following ``map`` and ``location`` blocks should be
+added to serve static files via NGINX.
+
+.. code-block:: nginx
+
+    map $sent_http_content_type $content_type_csp {
+        default "";
+        ~image/ "default-src 'none'";
+    }
+
+    server {
+        # the rest of the configuration
+
+        location @odoo {
+            # copy-paste the content of the / location block
+        }
+
+        # Serve static files right away
+        location ~ ^/[^/]+/static/.+$ {
+            # root and try_files both depend on your addons paths
+            root ...;
+            try_files ... @odoo;
+            expires 24h;
+            add_header Content-Security-Policy $content_type_csp;
+        }
+    }
+
+The actual ``root`` and ``try_files`` directives are dependant on your installation, specifically on
+your :option:`--addons-path <odoo-bin --addons-path>`.
+
 .. example::
-   Say Odoo has been installed via the **debian packages** for Community and Enterprise and the
-   :option:`--addons-path <odoo-bin --addons-path>` is ``'/usr/lib/python3/dist-packages/odoo/addons'``.
 
-   Using the above NGINX (https) configuration, the following location block should be added to
-   serve static files via NGINX.
+   .. tabs::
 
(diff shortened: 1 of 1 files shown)


content/administration/install/deploy.rst
```diff
@@ -439,51 +439,70 @@ Odoo static files are located in each module's :file:`static/` folder, so static
...
-   serve static files via NGINX.
+   .. tabs::
 
-   .. code-block:: nginx
+      .. group-tab:: Debian package
 
-       location @odoo {
-           # copy-paste the content of the / location block
-       }
+         Say Odoo has been installed via the **debian packages** for Community and Enterprise and
```
joda-odoo, line 482: > Say Odoo has been installed

`Say Odoo was installed`

Same thing with the part about git

content/administration/install/deploy.rst
```diff
@@ -439,51 +439,70 @@ Odoo static files are located in each module's :file:`static/` folder, so static
...
-   above snippets only highlight two possible configurations and may not be used as-is.
+         .. code-block:: nginx
+
+            root /usr/lib/python3/dist-packages/odoo/addons;
+            try_files $uri @odoo;
+
+      .. group-tab:: Git sources
+
+         Say Odoo has been installed via the **sources**. The two git repositories for Community and
+         Enterprise have been cloned in :file:`/opt/odoo/community` and :file:`/opt/odoo/enterprise`
```
joda-odoo, line 496: Maybe both parts should be in the same sentence.

```
Say Odoo was installed via the **source**, and both the Community and Enterprise repositories were cloned in ...
```

content/administration/install/deploy.rst
```diff
@@ -439,51 +439,70 @@ Odoo static files are located in each module's :file:`static/` folder, so static
...
 
-.. example::
-   Say Odoo has been installed via the **debian packages** for Community and Enterprise and the
-   :option:`--addons-path <odoo-bin --addons-path>` is ``'/usr/lib/python3/dist-packages/odoo/addons'``.
+It is recommended to set the ``Content-Security-Policy: default-src 'none'`` header on all images
+delivered by the web server. It is not stricly necessary as users cannot modify/inject content
+inside of modules' :file:`static/` folder and that existing images are final: they don't fetch new
+resources by themselves. Still, it is good hygiene.
+
+Using the above NGINX (https) configuration, the following ``map`` and ``location`` blocks should be
```
joda-odoo, line 447: > Using the above NGINX (https) configuration,

In this case the word bellow (ci-dessous) would be more appropriate

```
Using the NGINX (https) configuration provided bellow
```

content/administration/install/deploy.rst
joda-odoo, line 496:
```suggestion
         Say Odoo has been installed via the **sources**, and both the Community and Enterprise repositories were cloned in :file:`/opt/odoo/community` and :file:`/opt/odoo/enterprise`
```

content/administration/install/deploy.rst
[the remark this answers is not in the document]
Julien00859: I'm actually referencing the NGINX configuration shown inside the Https section, which is found earlier (above) in the current document :p The snippet found below is specific to serving static files but work in joint with the https configuration.

content/administration/install/deploy.rst
Julien00859: I reformulated the bit even further using this oxford coma in several other places for consistency

toaa-odoo (approved):
Cleared up a few typos and phrasing in stead of @odoo/be-doc-review

Feyensv (approved):
@robodoo delegate+

PR #5493, opened 2023-08-14
[REF] install: move intro to main install page and split by install type

xpl-odoo:
This PR moves the introduction about the different installation methods and editions found on the `install/install.rst` page directly under the main `install/` page.
It creates three pages, one per installation method documented on the `install/install.rst` page (Docker being documented on docker.com and maintained by us), and moves the related content there.
In addition, it fixes various typos/grammar issues and improves the content according to the documentation guidelines.
Various references and links across the documentation had to be fixed accordingly.

task-3459070


AntoineVDV (approved):
Good initiative :100:

I didn't go through `packages.rst` and `source.rst` because I assume it's more or less the same content.

Did you consider basing your work on 14.0 or, at least, 15.0? I believe the impacted refs and the global `install/install.rst` page structure are very similar, so you could do it with a cherry-pick plus some tweaks. It would allow for maintaining the install doc in those versions without being hit by the forward-port to 16.0 each time.

xpl-odoo:
> Did you consider basing your work on 14.0 or, at least, 15.0? I believe the impacted refs and the global `install/install.rst` page structure are very similar, so you could do it with a cherry-pick plus some tweaks. It would allow for maintaining the install doc in those versions without being hit by the forward-port to 16.0 each time.

At first, I targeted 14.0, thinking the page was the same for all versions. However, I quickly noticed `install/install.rst` on 16.0 differs from previous versions (introducing OS tabs notably). I asked @d-fence his opinion on which version I should target - given he is one of the most active maintainers of the whole section. He told me to target 16.0, but I can re-target the PR if you believe it's worth it.

AntoineVDV:
The content is indeed very different from 14.0 to 15.0, but the global structure is the same, so it'd save contributors some time if the page is split in the earliest version possible ;)

d-fence (approved):
LGTM

xpl-odoo:
@robodoo r+

xpl-odoo:
@robodoo r+

PR #5500, opened 2023-08-16
[IMP] email com: update daily limit

jcs-odoo:
The current default daily limit for one-app free is 20 mails only. This commit updates the value.


jcs-odoo:
@robodoo r+

PR #5541, opened 2023-08-18
[IMP] orm: changelog about group/aggregate/order by no-store related fields

ryv-odoo:
https://github.com/odoo/odoo/pull/127353
https://github.com/odoo/enterprise/pull/46795


AntoineVDV (approved):
@robodoo delegate+

PR #5556, opened 2023-08-21
Adding instructions to patch the database after updating code

ezoer:
Dear Odoo-er,

After reading the instructions on updating the database, I noticed that there were no instructions included to update the database from source. Perhaps this was on purpose in which case it might be good to explain in the document why the step I've added is unnecessary.

Best regards,

Erwin Zoer


xpl-odoo:
Hello @ezoer, thank you for your contribution!
I'll check with my colleague @d-fence to see if these instructions should be added.

PR #5653, opened 2023-08-25
[IMP] howtos/translations: text & tips

odoostan:
Remove parenthesis around a term that's fairly straightforward to understand and later used as a title
Add a much-needed point (comma would be fine as well)
Correct an improper plural form

### DIFF
--- a/content/developer/howtos/translations.rst
+++ b/content/developer/howtos/translations.rst
@@ -12,8 +12,8 @@ This section explains how to provide translation abilities to your module.
 Exporting translatable term
 ===========================
 
-A number of terms in your modules are "implicitly translatable" as a result,
-even if you haven't done any specific work towards translation you can export
+A number of terms in your modules are implicitly translatable. As a result,
+even if you haven't done any specific work towards translation, you can export
 your module's translatable terms and may find content to work with.
 
 .. todo:: needs technical features


odoostan:
I would like to have methods `get_field_string` and `get_field_help` mentioned somewhere in the docs, what are your thoughts on using this page ?

odoostan:
I don't see it anywhere that it's possible to translate hardcoded strings from the code. I think it would be a great addition to this page as well.
@brinkflew 
image

content/developer/howtos/translations.rst
```diff
@@ -156,14 +156,14 @@ Plural
...
-    msg = _("You have %(count)s invoice", count=invoice_count)
+    msg = _("You have %(count) invoice", count=invoice_count)
     if invoice_count > 1:
       msg += _("s")
 
 **Do** keep in mind every language has different plural forms::
 
     if invoice_count > 1:
-      msg = _("You have %(count)s invoices", count=invoice_count)
+      msg = _("You have %(count) invoices", count=invoice_count)
```
AntoineVDV, line 166: The format for labeled placeholders is `%(placeholder)s`. This change is not correct.

AntoineVDV (changes requested):

odoostan:
@AntoineVDV sorry for the wait, I forgot about this PR.
Adapted following your suggestion and corrections.
I would like to see some of the stuff I mentioned (and much more) in the docs, right now there's no docs at all on the "ir.translation" model.
However it seems that's very common, only [two standard modules](https://www.odoo.com/documentation/16.0/developer/reference/standard_modules.html) have their documentation in the reference.
Any plans to expand on that ?

AntoineVDV (approved):
@robodoo r+

AntoineVDV:
>I would like to see some of the stuff I mentioned (and much more) in the docs, right now there's no docs at all on the "ir.translation" model.
However it seems that's very common, only [two standard modules](https://www.odoo.com/documentation/16.0/developer/reference/standard_modules.html) have their documentation in the reference.
Any plans to expand on that ?

The developer documentation is a collaborative effort and only a few teams and developers invest time in maintaining that doc, so the best way to see something being added to the documentation is to create the PR ;)

PR #5702, opened 2023-09-01
[FIX] developer/tutorials/discover_js_framework: fix broken "domain.js" link

Armin-FalDiS:
The link pointing to domain.js file was broken.
The xml id on the line before that is also broken but it will be fixed with PR#5699.

### DIFF
--- a/content/developer/tutorials/discover_js_framework/02_web_framework.rst
+++ b/content/developer/tutorials/discover_js_framework/02_web_framework.rst
@@ -96,7 +96,8 @@ services, and components can import a service with the `useService()` hook.
       <https://github.com/odoo/odoo/blob/1f4e583ba20a01f4c44b0a4ada42c4d3bb074273/
       odoo/addons/base/views/res_partner_views.xml#L525>`_).
    #. A button `New Orders`, which opens a list view with all orders created in the last 7 days. Use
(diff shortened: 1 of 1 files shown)


content/developer/tutorials/discover_js_framework/02_web_framework.rst
```diff
@@ -96,8 +96,7 @@ services, and components can import a service with the `useService()` hook.
       <https://github.com/odoo/odoo/blob/1f4e583ba20a01f4c44b0a4ada42c4d3bb074273/
       odoo/addons/base/views/res_partner_views.xml#L525>`_).
    #. A button `New Orders`, which opens a list view with all orders created in the last 7 days. Use
-      the `Domain <https://github.com/odoo/odoo/blob/1f4e583ba20a01f4c44b0a4ada42c4d3bb074273/
-      odoo/addons/web/static/src/core/domain.js#L19>`_ helper class to represent the domain.
```
AntoineVDV, line 100: It seems like only the extra `odoo/` part poses a problem, so you could keep the format as-is and remove the problematic part.
```rst
the `Domain <https://github.com/odoo/odoo/blob/1f4e583ba20a01f4c44b0a4ada42c4d3bb074273/
addons/web/static/src/core/domain.js#L19>`_ helper class to represent the domain.
```

AntoineVDV (changes requested):

AntoineVDV (approved):
@robodoo r+ squash

PR #5710, opened 2023-09-01
[Fix] Swapped Section Titles

motu-odoo:
The section titles for button types appear to be swapped. The action type section is discussing the object type while the object type section is discussing the action type. This commit fixes the titles to match the content.


AntoineVDV (commented):
Hi @motu-odoo, it looks like you have a `master` base for your branch. Could you rebase it onto `16.0`?

@fw-bot up to saas-16.4

PR #5716, opened 2023-09-04
[FIX] odoo.sh: typo in branches.rst

baro-odoo:
Original text: Similar to staging branches, the emails are not sent but are intercepted by a mailcatcher and scheduled actions are not triggered as often is the database is not in use.

Correction: as often as* the database is not in use.


jcs-odoo:
@baro-odoo 
I changed the base to 15.0 instead of master and changed the text. Shouldn't it be "as long as" instead of "as often as"?
Let me know if you need another change.

jcs-odoo:
@robodoo r+

PR #5719, opened 2023-09-04
[IMP] maintain: refresh the change hosting solution page

xpl-odoo:
This PR updates the page's content as it is a bit dated in some places (see task) and takes into consideration the latest rst guidelines.

task-3434787


xpl-odoo:
@robodoo retry

PR #5773, opened 2023-09-08
[IMP] Views howtos: add details on how to use a view in arch

### LINKED ISSUE #5772: Lack of good documentation
I am trying to follow the tutorial "Customize a view type" for Odoo version 16:
https://www.odoo.com/documentation/16.0/developer/howtos/javascript_view.html

But at the end, it leaves us with the point number 6 "Use the view in an arch." which is nothing useful.

Odoo, please, make a better documentation, there are a lot of missing parts or parts that leaves things as is, without going into the deep of it.

I would really like to create a custom view by myself, but this lack of effort from Odoo about documentation is really annoying.

fdardenne:
fix #5772

### DIFF
--- a/content/developer/howtos/javascript_view.rst
+++ b/content/developer/howtos/javascript_view.rst
@@ -253,10 +253,16 @@ Creating a new view is an advanced topic. This guide highlight only the essentia
 
       registry.category("views").add("beautifulView", beautifulView);
 
-#. Use the view in an arch.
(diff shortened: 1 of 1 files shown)


AntoineVDV:
@robodoo delegate+

fdardenne:
@robodoo r+

fdardenne:
@robodoo retry

PR #5786, opened 2023-09-08
[IMP] cli: add unit to memory options

loleloup:
The documentation is unclear about the unit used for the --limit-memory-soft and --limit-memory-hard options of odoo-bin.  The unit used is bytes, but it isn't mentioned and the default unit is shown in mebibytes, causing confusion.


AntoineVDV (approved):
@robodoo r+

PR #5800, opened 2023-09-11
[IMP] developer reference: update t-debug doc

xmo-odoo:
The value of `t-debug` is now optional, and leaving it out means the directive just invokes the `breakpoint` builtin, which was added in 3.7 and is configurable via an envvar or a hook callable.


AntoineVDV:
@robodoo r+

PR #5825, opened 2023-09-12
[IMP] install: clarify Windows is not recommended for prod

odony:
It has always been the case informally but clarify it in the doc.


AntoineVDV (commented):
Shouldn't it be mentioned in the Source Install section too?

PR #5762, opened 2023-09-07
[ADD] WhatsApp: Documentation

tiku-odoo:
Add WhatsApp configuration documentation

Dev PR https://github.com/odoo/enterprise/pull/45601#pullrequestreview-1608996613
Task [3495962](https://www.odoo.com/web#cids=3&model=project.task&view_type=form&id=3495962)

PR #5853, opened 2023-09-13
Update whatsapp.rst

fpodoo:


StraubCreative:
Thanks @fpodoo !
We added these changes on #5762 and deployed today 🚀

PR #5854, opened 2023-09-13
[IMP] better wording in Chapter 13

dispread:
More concise and straightforward way of wording this


AntoineVDV (commented):
Hello @dispread, and thank you for your suggestion.

I would rather not change the wording here because it's not entirely equivalent. In Odoo, it's all about *preventing* users from doing something, rather than *allowing* them to do it. So, using the term "prevent" is clearer in this case. Also, checking whether a key is not A or B is sometimes different than checking if the key is C or D, for example when there also exists values E, F...

dispread:
Fair enough

PR #5899, opened 2023-09-22
[FIX] developer/tutorials: wrong kanban view example

dispread:
Pretty sure the previous wasnt the intended result as it would show the "this is new" in a new kanban card. This just groups it with the property it belongs to.


Feyensv (approved):
@robodoo r+ squash

Thanks for the fix, it's indeed the expected definition, the `oe_kanban_global_click` div should encompass the whole kanban card content :+1:

PR #5938, opened 2023-09-27
[IMP] adyen: update prefix information

anko-odoo:
Since user will no longer be required to create link on their own they will not require information how to do so. Both Checkout API and Reccurring API were replaced by prefix.

task-33381

### DIFF
--- a/content/applications/finance/payment_providers/adyen.rst
+++ b/content/applications/finance/payment_providers/adyen.rst
@@ -30,8 +30,7 @@ Odoo needs your **API Credentials** to connect with your Adyen account, which co
 - :ref:`API Key <adyen/api_and_client_keys>`: The API key of the webservice user.
 - :ref:`Client Key <adyen/api_and_client_keys>`: The client key of the webservice user.
 - :ref:`HMAC Key <adyen/hmac_key>`: The HMAC key of the webhook.
-- :ref:`Checkout API URL <adyen/urls>`: The base URL for the Checkout API endpoints.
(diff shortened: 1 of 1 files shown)


content/applications/finance/payment_providers/adyen.rst
```diff
@@ -106,18 +105,12 @@ URLs, proceed as follows:
...
-   `https://yourprefix-pal-live.adyenpayments.com/pal/servlet/Recurring`.
+#. In the :guilabel:`API URL Prefix` field, enter the prefix you previously saved.
 
 .. note::
-   If you are trying Adyen as a test, you can use the following URLs instead:
+   If you are trying Adyen as a test, you can use the following Prefix instead:
 
-   - :guilabel:`Checkout API URL`: `https://checkout-test.adyen.com`
-   - :guilabel:`Recurring API URL`: `https://pal-test.adyen.com/pal/servlet/Recurring`
+   - :guilabel:`API URL Prefix`: `checkout-test`
```
auva-odoo, line 113: Since there is only one prefix, I would remove the bullet list and merge it with the previous sentence as follows:
If you are trying Adyen as a test, you can use the following :guilabel:`API URL Prefix` instead: `checkout-test`.
What do you think?

auva-odoo (changes requested):
Hey @anko-odoo I have one small suggestion, let me know if you have any questions :)

content/applications/finance/payment_providers/adyen.rst
anko-odoo: Yeah, sounds all right.

PR #5984, opened 2023-10-03
[FIX] developer/reference: outdated documentation of iap extract api

lobaudoux:
For security reasons, we had to drop the support of those routes.

### DIFF
--- a/content/developer/reference/extract_api.rst
+++ b/content/developer/reference/extract_api.rst
@@ -22,7 +22,7 @@ Overview
 ========
 
 The extract API uses the JSON-RPC2_ protocol; its endpoint routes are located at
-`https://iap-extract.odoo.com`.
+`https://extract.api.odoo.com`.
 
 Version
 -------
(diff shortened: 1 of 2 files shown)


Feyensv:
Are those routes only removed in 16+ or earlier as well ?
If so, it's maybe worth targeting older branches :thinking:.

lobaudoux:
Hello @Feyensv ,

> Are those routes only removed in 16+ or earlier as well ?

These routes weren't actually used in any Odoo versions, but it's only in version 16 of the documentation that is was documented, hence the 16.0 target.

The rationale for the extract API documentation is that we document it according to the version that was used at the time.
E.g. for Odoo 14, we're only documenting the old version of the API, version 120.

IMO, this is confusing as the extract API is independent of Odoo versions. I think we should document the latest version of the API in all versions of the documentation.

Feyensv:
Ok then, thanks for the explanation :+1: 

@robodoo delegate+

Feyensv (approved):

lobaudoux:
@robodoo r+

lobaudoux:
@robodoo rebase-ff

PR #6071, opened 2023-10-09
[IMP] administration: add instructions to enable HSTS

mart-e:


content/administration/install/deploy.rst
```diff
@@ -352,6 +352,32 @@ in ``/etc/nginx/sites-enabled/odoo.conf`` set:
...
+Strict Transport Security`_ or HSTS.
+
+On hosted odoo.com database
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+To enable HSTS on an Odoo instance, add the following element inside the `<head>`
+tag of the `web.layout` and `website.layout` view:
+
+.. code-block :: html
+
+  <link href="/_odoo/hsts/enable" rel="stylesheet"/>
```
johandem, line 369: If we're talking about SaaS, this is done automatically for new databases since March this year. Besides I'm not sure we should recommend this approach, I think it was more of a workaround at the time.
Customers can requests to enable security features through our support now (can be done via the support page)
These security features bundle the following:
- Strict-Transport-Security (HSTS)
- Secure attribute on session_id cookie
- SameSite=Lax on session_id cookie

content/administration/install/deploy.rst
```diff
@@ -352,6 +352,32 @@ in ``/etc/nginx/sites-enabled/odoo.conf`` set:
...
+
+On a self-hosted database
+~~~~~~~~~~~~~~~~~~~~~~~~~
+On your own server, add the `Strict-Transport-Security` to the requests. For
```
johandem, line 373:
```suggestion
On your own server, add the `Strict-Transport-Security` to the response. For
```

PR #6073, opened 2023-10-09
Replace technical-training-sandbox to technical-training

YousufSohail:


AntoineVDV:
Hi @YousufSohail, this section of the documentation is meant only for developers working at Odoo ;)
See the note above: "The following steps are intended only for Odoo employees. The mentioned repositories are not   accessible to third parties."

PR #6163, opened 2023-10-18
[FIX] partner_autocomplete: fix a typo

lman-odoo:


StraubCreative:
Good catch. Thank you @lman-odoo

@robodoo r+

StraubCreative (approved):
Good catch. Thank you @lman-odoo

@robodoo r+

PR #6191, opened 2023-10-23
[FIX] search panel: fix bad indentation

Polymorphe57:


AntoineVDV (approved):
@robodoo r+

AntoineVDV:
I didn't put a "fw up to" so you can double-check, but this is probably already fixed in master.

PR #6209, opened 2023-10-25
[IMP] extract_api: uniformize API documentation across all versions

lobaudoux:
We used to document the extract API according to the version of the documentation that is being viewed, i.e. in version 14, we documented the API version that was used in Odoo 14.

This is actually not a good idea as the extract API isn't strictly linked to Odoo versions, it has its own versioning system.

It can also be an issue if users don't realize they're viewing an old version of the API and start implementing customizations based on it.

This commit is an exact copy of the current `master` version of the extract API documentation.

### DIFF
--- a/content/developer/reference/extract_api.rst
+++ b/content/developer/reference/extract_api.rst
@@ -1,521 +1,756 @@
-:code-column:
-
 ===========
 Extract API
 ===========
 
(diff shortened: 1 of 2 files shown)


AntoineVDV (approved):
@robodoo r+

PR #6236, opened 2023-10-27
[FIX] developer: fix bad documentation of --unaccent

ryv-odoo:
In the CLI, `--unaccent` try to enable PostgreSQL unaccent extention when odoo is responsible to create new database(s).


AntoineVDV (approved):
@robodoo r+

PR #6250, opened 2023-10-30
[IMP] extract_api: document the new API version

lobaudoux:
We have introduced a second version of the Extract API that adds the IAP account token to the `get_result` and `validate` routes.

There are also other changes in this commit:
 - The deprecated stuff has been removed as it's been deprecated for a while and won't be supported anymore in the future.
 - The `status` table of the `get_result` route has been added as it was missing for some reason.
 - The documentation regarding the `/validate` routes have been removed for security reasons.


AntoineVDV (approved):
@robodoo r+

PR #6322, opened 2023-11-07
[IMP] sales: minor adjustment to dynamic text portion of PDF builder

ksc-odoo:
Made a minor adjustment to the dynamic text portion, based on the feedback provided by @dalaOdoo in PR#6029, to accommodate the "double-underscores" that were missing in the initial doc.


ksc-odoo:
@StraubCreative made that quick adjustment - and believe it's ready for a final look. Let me know if there's anything else I need to adjust, and I'll get to it right away. Thanks!

StraubCreative (approved):
@robodoo r+

PR #6274, opened 2023-11-02
[IMP] upgrade: overhaul upgrade doc

nmarotte:
Overhauling of the documentation for Upgrade as part of the need for Partner documentation for upgrade (task 3171474 )

This keep most of the information intact, merging the FAQ into the documentation itself, improves the structure (one page), remove redundancies, ...

It is part of a 2-phases project where the second phase would be to add information about upgrade for developers (currently sitting in developer/reference/upgrade.rst but that will be moved to developer/howtos.html) to contain informations about migration scripts, upgrading custom code, etc ...

Related PR : #5774 #5953

PR #6485, opened 2023-11-16
17.0 upgrade doc p2 3171474 nama

nmarotte:
tmp PR for runbot


nmarotte:
Documentation modified with #6274  #6757

PR #6493, opened 2023-11-17
[IMP] developer: document how to create standalone owl apps

sdegueldre:
Since we've introduced Owl, developers want more and more to create standalone Owl applications outside of the webclient. This commit introduces a How-to for doing just that.

### DIFF
--- a/content/developer/howtos/standalone_owl_application.rst
+++ b/content/developer/howtos/standalone_owl_application.rst
@@ -1,0 +1,172 @@
+===================================
+Create a standalone Owl application
+===================================
+
+For any number of reasons, you may want to have a standalone Owl application that isn't a part of
+the web client. One example in Odoo is the self-ordering application, that lets customers order
+food from their phone. In this chapter we will go into what's required to achieve something like this.
+
+Overview
+========
+
+To have a standalone Owl app, a few things are required:
+
+- a root component for the application
+- some setup code to start the services and mount the component
+- an assets bundle that contains the setup code
+- a QWeb view that calls the assets bundle
+- a controller that renders the view
+
+1. Root component
+=================
+
+To keep things simple, let's start with a very straightforward component that just renders
+"Hello, World!". This will let us know at a glance if our setup is working.
+
+First, create the template in :file:`/your_module/static/src/standalone_app/root.xml`.
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="UTF-8"?>
+    <templates xml:space="preserve">
+        <t t-name="your_module.Root">
+            Hello, World!
+        </t>
+    </templates>
+
+Then create the JavaScript file for that component in :file:`/your_module/static/src/standalone_app/root.js`.
+
+.. code-block:: js
+
+    /** @odoo-module */
+    import { Component } from "@odoo/owl";
+
+    export class Root extends Component {
+        static template = "your_module.Root";
+        static props = {};
+    }
+
+.. seealso::
+   :ref:`Owl components reference<frontend/components>`.
+
+2. Setting up the environment and mounting the application
+==========================================================
+
+An Owl application needs an environment, and the Odoo JavaScript framework needs that environment to
+contain the :ref:`services<frontend/services>`.
+Starting the services is also required to load the translations, which we need to do before mounting
+our Owl application, so that we can give Owl a working translation function.
+
+Create the JavaScript file that will mount the app in :file:`/your_module/static/src/standalone_app/app.js`.
+
+.. code-block:: js
+
+    /** @odoo-module */
+    import { whenReady } from "@odoo/owl";
+    import { mountComponent } from "@web/env";
+    import { Root } from "./root";
+
+    whenReady(() => mountComponent(Root, document.body));
+
+
+3. Creating an assets bundle containing our code
+================================================
+
+In the manifest of your module, create a new :ref:`assets bundle<reference/assets_bundle>`.
+It should include the `web._assets_core` bundle, which contains the Odoo JavaScript
+framework and the core libraries it needs (e.g. Owl and luxon), after which you can have a
+glob that adds all the files for your application in the bundle.
+
+.. code-block:: py
+    :emphasize-lines: 9-10
+
+    {
+        # ...
+        'assets': {
+            'your_module.assets_standalone_app': [
+                ('include', 'web._assets_helpers'),
+                'web/static/src/scss/pre_variables.scss',
+                'web/static/lib/bootstrap/scss/_variables.scss',
+                ('include', 'web._assets_bootstrap'),
+                ('include', 'web._assets_core'),
+                'your_module/static/src/standalone_app/**/*',
+            ],
+        }
+    }
+
+The other lines are bundles and scss files that are required to make Bootstrap work. They are
+mandatory, as the components of the web framework use bootstrap classes for their styling and
+layout.
+
+.. caution::
+    Make sure that the files for your standalone app are only added to this bundle, if you already
+    have a definition for `web.assets_backend` or `web.assets_frontend` and they have globs, make
+    sure these globs don't match the files for your standalone app, otherwise the startup code for
+    your app will conflict with the existing startup code in those bundles.
+
+.. seealso::
+    :ref:`Module manifest reference<reference/module/manifest>`.
+
+4. XML view that calls the assets bundle
+========================================
+
+Now that we have created our assets bundle, we need to create a
+:ref:`QWeb view<reference/view_architecture/qweb>` that uses that assets bundle.
+
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="utf-8"?>
+    <odoo>
+        <template id="your_module.standalone_app">&lt;!DOCTYPE html&gt;
+            <html>
+                <head>
+                    <script type="text/javascript">
+                        var odoo = {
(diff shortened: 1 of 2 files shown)


content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,163 @@
...
+    import { Component, xml, whenReady, App } from "@odoo/owl";
+    import { makeEnv, startServices } from "@web/env";
+    import { templates } from "@web/core/templates";
+    import { _t } from "@web/core/l10n/translation";
+
+    class Root extends Component {
+        static template = xml`Hello, World!`;
+    }
+
+    whenReady().then(async () => {
```
ged-odoo, line 57: i think it would make sense to split the code already in 3 files: `app.js`, `root.js` and `root.xml` to give them a better starting point, with the component definition separate from the setup code

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,163 @@
...
+
+Overview
+========
+
+To have a standalone Owl app, a few things are required:
+
+- a root component for the application
+- some setup code to start the services and mount the component
+- an assets bundle that contains the setup code
+- an QWeb view that calls the assets bundle
```
aab-odoo, line 17:
```suggestion
- a QWeb view that calls the assets bundle
```

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,163 @@
...
+
+    whenReady().then(async () => {
+        const env = makeEnv();
+        await startServices(env);
+        const app = new App(Root, {
+            templates,
+            env: env,
+            dev: env.debug,
+            translateFn: _t,
+            translatableAttributes: ["data-tooltip"],
```
aab-odoo, line 65: would be nice if we didn't have to setup this

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,163 @@
...
+.. code-block:: py
+
+    {
+        # ...
+        'assets': {
+            'your_module.assets_standalone_app': [
+                ('include', 'web._assets_helpers'),
+                'web/static/src/scss/pre_variables.scss',
+                'web/static/lib/bootstrap/scss/_variables.scss',
+                ('include', 'web._assets_bootstrap'),
```
aab-odoo, line 87: any idea/reason why the three instruction above aren't directly in assets_bootstrap? Would be nice if we could do 
```py
'your_module.assets_standalone_app': [
                ('include', 'web._assets_bootstrap'),
                ('include', 'web._assets_core'),
                'your_module/static/src/standalone_app/**/*',
]
```

sdegueldre: Not sure, I've added a point to the roadmap to look at scss bundles and clean things up, they're hard to use and it's easy to miss files. Since core depends on bootstrap I'd like to directly include `_assets_bootstrap` in `_assets_core` eventually.

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,176 @@
...
+                ('include', 'web._assets_core'),
+                'your_module/static/src/standalone_app/**/*',
```
ged-odoo, line 102: just one last comment: I get why you put the content of the app in a sub folder, but maybe we should simply do it in `src`. In most case, it will be fine. Not sure though, if you prefer `src/standalone_app`, i can live with it

sdegueldre: I think this avoids the pitfall of people including their entire src folder on their controller when their module already has some files in static that they use to customize the webclient. Maybe I should add a warning in that spirit so that people don't include this folder in the assets backend either though.

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
+=====================================
+Creating a standalone Owl application
```
ged-odoo, line 2: title should probably be `Create a standalon Owl application` to make it look more similar to other howtos

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+our Owl application, so that we can give Owl a working translation function.
+
+Create the JavaScript file that will mount the app in `/your_module/static/src/standalone_app/app.js`.
+
+.. code-block:: js
+
+    /** @odoo-module */
+    import { whenReady } from "@odoo/owl";
+    import { mountComponent } from "@web/env";
+    import { Root } from "@your_module/standalone_app/root";
```
ged-odoo, line 67: i would do a relative import here

ged-odoo (approved):

ged-odoo:
looks good to me

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+                'web/static/src/scss/pre_variables.scss',
+                'web/static/lib/bootstrap/scss/_variables.scss',
+                ('include', 'web._assets_bootstrap'),
+                ('include', 'web._assets_core'),
+                'your_module/static/src/standalone_app/**/*',
+            ],
+        }
+    }
+
+The lines before `web._assets_core` are bundles and scss files that are required to make
```
AntoineVDV, line 96: If that helps, you could use the [`emphasize-lines` option](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-emphasize-lines) of the code-block directive to simply mention "the highlighted lines" instead of "the lines before".

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+Bootstrap work. They are required, as the components of the web framework use bootstrap classes for
+their styling and layout.
+
+.. note::
```
AntoineVDV, line 100: That sounds more like an `important` or `caution` admonition to me.

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+    Make sure that the files for your standalone app are only added to this bundle, if you already
+    have a definition for `web.assets_backend` or `web.assets_frontend` and they have globs, make
+    sure these globs don't match the files for your standalone app, otherwise the startup code for
+    your app will conflict with the existing startup code in those bundles.
+
+.. seealso::
+    :ref:`module manifest reference<reference/module/manifest>`.
```
AntoineVDV, line 107: Use the `doc` role to link entire pages.
```suggestion
    :doc:`Module manifests reference </developer/reference/backend/module>`
```

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+
+4. XML view that calls the assets bundle
+========================================
+
+Now that we have created our assets bundle, we'll need to create a
```
AntoineVDV, line 112:
```suggestion
Now that we have created our assets bundle, we need to create a
```

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+.. code-block:: xml
+
+    <?xml version="1.0" encoding="utf-8"?>
+    <odoo>
+        <template id="your_module.standalone_app">&lt;!DOCTYPE html&gt;
+            <html>
+                <head>
+                    <script type="text/javascript">
+                        var odoo = {
+                            csrf_token: "<t t-nocache="The csrf token must always be up to date." t-esc="request.csrf_token(None)"/>",
```
AntoineVDV, line 124: Just out of curiosity: isn't `t-esc` deprecated?

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+                    <t t-call-assets="your_module.assets_standalone_app" />
+                </head>
+                <body/>
+            </html>
+        </template>
+    </odoo>
+
+This template only does two things: it initializes the odoo global variable, then calls the assets
+bundle we just defined. Initializing the odoo global variable is a necessary step. This variable
+should be an object that contains the following:
```
AntoineVDV, line 138:
```suggestion
This template only does two things: it initializes the `odoo` global variable, then calls the assets
bundle we just defined. Initializing the `odoo` global variable is a necessary step. This variable
should be an object that contains the following:
```

content/developer/howtos/standalone_owl_application.rst
```diff
@@ -0,0 +1,171 @@
...
+
+- The CSRF token, which is required to interact with HTTP controllers in many cases.
+- The debug value, which is used in many places to add additional logging or developer-friendly checks
```
AntoineVDV, line 141:
```suggestion
- The debug value, which is used in many places to add additional logging or developer-friendly checks.
```

AntoineVDV (approved):
Nice little addition! It will be useful for sure :+1: 

@robodoo delegate+

content/developer/howtos/standalone_owl_application.rst
sdegueldre: Definitely want to point to the manifest section here, if another section gets added on top of the manifest section this would be broken.

sdegueldre:
@robodoo r+

PR #6495, opened 2023-11-17
[IMP] website: use plausible.io servers and account

Donapi:
[id](https://www.odoo.com/web#id=3540753&cids=1&menu_id=4720&action=333&active_id=3835&model=project.task&view_type=form)


jcs-odoo (changes requested):
Mention Odoo's default behaviour is using Odoo's servers and that changing this parameter is to use plausible.io's servers. It's not only about the accounts.

jcs-odoo (approved):
@robodoo r+

PR #6496, opened 2023-11-17
[FIX] dev/reference: Fix external reference

rrahir:
The tour utils were red and moved in Odoo (odoo/odoo#107618).


AntoineVDV (approved):
@robodoo r+

PR #6522, opened 2023-11-21
[FIX] payroll: add missing show-content metadata

xpl-odoo:
Without the [:show-content: metadata markup](https://www.odoo.com/documentation/17.0/contributing/documentation/rst_cheat_sheet.html#document-metadata), it is not possible to open the page by navigating from the toctree.


xpl-odoo:
cc @jero-odoo @larm-odoo 
@robodoo r+

PR #6607, opened 2023-11-23
[FIX] getting_started: invalid addon-path and useless lambda

clbr-odoo:
- First commit: the addons-path adds `/technical-training-sandbox` before it becomes a valid addons folder.
- Second commit: remove a useless lambda that caused some incomprehension from newdooers.


AntoineVDV (approved):
@robodoo r+ rebase-ff

PR #6669, opened 2023-11-27
[FIX] developer: remove confusing section about starting services

sdegueldre:
When writing the documentation page for standalone Owl apps, it was originally required to create the env and start the services manually, and then mount the application by giving it a bunch of configuration parameters. During review, it was noted that this was error prone and confusing, and as such a corresponding PR added a helper to mount the application without needing to do much by hand. While the code was adapted in the documentation that was merged, the text still referred to starting the services and creating the environment which is no longer necessary.

This commit removes references to those things, and because the resulting section is very small and not very interesting it has been merged into the previous section.

### DIFF
--- a/content/developer/howtos/standalone_owl_application.rst
+++ b/content/developer/howtos/standalone_owl_application.rst
@@ -12,7 +12,6 @@ Overview
 To have a standalone Owl app, a few things are required:
 
 - a root component for the application
-- some setup code to start the services and mount the component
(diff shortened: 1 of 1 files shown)


fdardenne (approved):

Feyensv (approved):
@robodoo delegate+

fdardenne:
@robodoo r+

sdegueldre:
@robodoo r+

PR #6699, opened 2023-11-28
Update source.rst

jrr-odoo:
Currently, version 16 for PG is not supported or available. Testing is still ongoing for that version. Since this is the case. we should not have version 12 or above because that is not accurate. When ver16 is available, we can update the doc to show up to version 16.0 and for future releases. reference
https://discord.com/channels/678381219515465750/687339658694098958/1173638065911255062

### DIFF
--- a/content/administration/install/source.rst
+++ b/content/administration/install/source.rst
@@ -144,7 +144,7 @@ Odoo requires **Python 3.10** or later to run.
       to download and install Python 3 if needed.
 
 .. note::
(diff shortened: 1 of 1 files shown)


jrr-odoo:
I've been in discussion with R&D regarding upgrades. While Version 16 is compatible with Odoo, it's not yet supported on the upgrade server. We are likely to update the information on our page [Odoo Supported Versions](https://www.odoo.com/documentation/17.0/administration/maintain/supported_versions.html) to reflect this, as it pertains specifically to the upgrade server, rather than Odoo itself.

PR #6725, opened 2023-11-30
[IMP] accounting: Incoterms

toaa-odoo:
Updated the Incoterms page to add "location" field and fixed typos

taskid-3619104


toaa-odoo:
Wrong branch name

PR #6728, opened 2023-11-30
[FIX] install: do not create superuser for postgres

william-andre:
The preferred/safe way of running Odoo is with a standard user, with only the right to create a database.
See https://github.com/odoo/odoo/commit/b6b73551dbbb3079f043920f21554b945fa4870e


AntoineVDV:
The content was restructured in 16.0, and the instructions in < 16.0 are okay, so only >= 16.0 is fine in this case :)

@robodoo r+

PR #6729, opened 2023-11-30
[IMP] developer/reference/cli: --proxy-mode

Julien00859:
Many customers struggle with their web server configuration, notably regarding the `--proxy-mode` option and the way `X-Forwarded-*` HTTP request headers are interpreted within Odoo.

The `--proxy-mode` section has been updated to cover the most common misunderstandings and to give guidances on how to setup a web server.

Odoo always only takes the last entry of the `X-Forwarded-*` request header because there are situations where it is not possible to determine which nth last entry to use. Employees might access their odoo database via the internal network: connecting directly to nginx, while customers might access the database via an additional proxy such as cloudflare. The real IP of employees would be the last inside the `X-Forwarded-For` chain, while the real IP of customers would be the *second* last entry inside the chain. It would be incorrect to always take the same nth last entry inside the chain. The cloudflare's own IP address must be discarded from the chain. Web servers usually feature a way to ignore trusted IP from the chain, a way so that the real IP of the user is always the last entry inside the chain. Odoo relies on such feature to be active and configured.

Prior discussions about `X-Forwarded-For`:
* odoo/odoo#104947
* odoo/odoo#139536

Using any of the `X-Forwarded-*` headers is disabled, even with `--proxy-mode`, in case the `X-Forwarded-Host` header is missing. System admin might be tempted to not set this header and to set `Host` instead, this is broken as this a user-agent would be able to spoof `X-Forwarded-Host` and Odoo would use that instead of the correct `Host`.

Prior discussions about `X-Forwarded-Host`:
* odoo/odoo#63277


content/developer/reference/cli.rst
```diff
@@ -425,7 +425,21 @@ HTTP
 .. option:: --proxy-mode
 
     enables the use of ``X-Forwarded-*`` headers through `Werkzeug's proxy
-    support`_. Only enable this when running behind a trusted web proxy!
```
Julien00859, line 428: This line is duplicated with the `.. warning` below

content/developer/reference/cli.rst
```diff
@@ -639,6 +653,8 @@ Here is a sample file:
     https://www.postgresql.org/docs/9.2/static/libpq-connect.html#AEN38208
 .. _Werkzeug's proxy support:
     https://werkzeug.palletsprojects.com/en/0.16.x/middleware/proxy_fix/#module-werkzeug.middleware.proxy_fix
+.. _set_real_ip_from:
+    https://nginx.org/en/docs/http/ngx_http_realip_module.html
```
AntoineVDV, line 657: No need for a hyperlink alias here. It's fine as it is, but the link could be directly in the paragraph to ease maintenance.

content/developer/reference/cli.rst
```diff
@@ -425,7 +425,21 @@ HTTP
...
+    It ignores all ``X-Forwarded-*`` headers in case ``X-Forwarded-Host`` is
+    missing from the request.
+
+    It always gets the real IP from the last entry of the ``X-Forwarded-For``
+    chain. Configure your web server accordingly using directives such as
+    nginx's `set_real_ip_from`_ in case there are other trusted proxies along
+    the chain that must be ignored.
+
+    ``X-Forwarded-Proto`` and ``X-Forwarded-Host`` are used to update the
+    request root URL which in turn is used to update the ``web.base.url``
```
AntoineVDV, line 439:
```suggestion
    request root URL, which in turn is used to update the ``web.base.url``
```

content/developer/reference/cli.rst
```diff
@@ -425,7 +425,21 @@ HTTP
...
+    system parameter upon a successful admin authentication. This system
+    parameter is used to generate all links for the current database, see
```
AntoineVDV, line 441:
```suggestion
    parameter is used to generate all links for the current database. See
```
```suggestion
    parameter is used to generate all links for the current database; see
```

AntoineVDV (approved):
Kudos for targeting 15.0 :+1: 

@robodoo delegate+

Julien00859 (commented):
@robodoo r+

Julien00859:
@fw-bot up to master

content/developer/reference/cli.rst
```diff
@@ -425,7 +425,21 @@ HTTP
 .. option:: --proxy-mode
 
     enables the use of ``X-Forwarded-*`` headers through `Werkzeug's proxy
-    support`_. Only enable this when running behind a trusted web proxy!
+    support`_.
+
+    It ignores all ``X-Forwarded-*`` headers in case ``X-Forwarded-Host`` is
+    missing from the request.
```
amh-mw, line 431: I think this language should be stronger. Something like "A properly configured proxy upstream must provide `X-Forwarded-Host` or --proxy-mode is ignored.

Julien00859: Ah shit I only see your message now and the PR as been merged. I don't think using a stronger language is actually needed, the piece of information was missing and people had some expectations, now that the info is there I think they'll configure their proxies accordingly.

Julien00859: We'll see, in case we continue to get messages on the matter we can still open another PR and change the wording :)

PR #6754, opened 2023-12-05
[FIX] fix wront repo in getting_started tutorial

fja-odoo:


AntoineVDV:
Hello, the `technical-training-sandbox` is actually the correct repository for hosting trainees' implementations of the training. The `technical-training` repository, on the other hand, was used for another technical training that is now outdated and is not in use anymore.

fja-odoo:
> Hello, the `technical-training-sandbox` is actually the correct repository for hosting trainees' implementations of the training. The `technical-training` repository, on the other hand, was used for another technical training that is now outdated and is not in use anymore.

The `technical-training-sandbox` repository has restricted access so partners/customer/enthusiasts cannot access it

AntoineVDV:
Indeed, but it is stated in chapter 2 that it is reserved for Odoo employees. I have plans to switch to the odoo/tutorials repository instead to get rid of that limitation.

PR #6755, opened 2023-12-05
[FIX] fix wront repo in getting_started tutorial

fja-odoo:


AntoineVDV:
See https://github.com/odoo/documentation/pull/6754#issuecomment-1840763238

PR #6896, opened 2023-12-12
[FIX] accounting: cheat sheet reconcile

malb-odoo:
Before this commit, when doing the reconciliation of the table present in the cheat sheet, the second reconciliation was wrong because the "Invoice 1" should be reconciled with the twos partials payment. This Pr correct that by changing the find parameter to target the partials also.

task: 3633468


AntoineVDV (commented):
@robodoo r+

PR #6897, opened 2023-12-12
[IMP] orm : add info about boolean default

asno-odoo:
When a default value for boolean is set as `False`, the values of new column for existing records is set to `NULL`. In python, `FALSE` and `NULL` might mean same thing in certain contexts but its not the case for Sql(we allow to add sql constraints).


AntoineVDV:
Hello, this tutorial is outdated and has been replaced by the "Getting Started" tutorial. "Building a module" is not maintained anymore. Kudos for the PR, though!

PR #6933, opened 2023-12-14
[IMP] orm : add info about boolean default

asno-odoo:
When a default value for boolean is set as False, the values of new column for existing records is set to NULL. In python, FALSE and NULL might mean same thing in certain contexts but its not the case for Sql(we allow to add sql constraints).

concerned code: https://github.com/odoo/odoo/blob/15.0/odoo/models.py#L2767


AntoineVDV (changes requested):
Hi again,

Adding this note here makes very little sense; readers are just being introduced to the concept of fields and didn't learn about default values yet, even less about SQL constraints. In fact, I believe that this note shouldn't be included in a general-purpose tutorial at all.

PR #6947, opened 2023-12-15
[IMP] sign: tags, signing order, reminders, expiration dates

Donapi:
taskid-[3609727](https://www.odoo.com/web#id=3609727&cids=1&model=project.task&view_type=form)
New features are available in v17:
Modify tags on signed documents.
Send automatic email reminders to people who have yet to sign a document after a set delay.
Signing order available as soon as you have more than one signer 
Setting expiration dates on signature requests.

### DIFF
--- a/content/applications/finance/sign.rst
+++ b/content/applications/finance/sign.rst
@@ -100,6 +100,14 @@ and the status of the signatures.
 .. image:: sign/signature-status.png
    :alt: Signature status
 
+Validity dates and reminders
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can set **validity dates** on limited-duration agreement documents or send **automatic email
+reminders** to obtain signatures on time. From your dashboard, click :guilabel:`Send` on your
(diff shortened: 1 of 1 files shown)


content/applications/finance/sign.rst
```diff
@@ -100,6 +100,14 @@ and the status of the signatures.
 .. image:: sign/signature-status.png
    :alt: Signature status
 
+Expiration and reminders
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can set a **validity date** to your document for limited-time offers or send an
```
bouvyd, line 106: 'limited time offers' sounds very, very commercial - so much so that it kind of overpowers the meaning we put into the feature imho
it's also to ensure that requests won't rot in the system forever
perhaps a more generic wording like 'time-sensitive documents' would be better suited

toaa-odoo (approved):

xpl-odoo (changes requested):

xpl-odoo (approved):
@robodoo r+

PR #6969, opened 2023-12-18
[FIX] barcode: typo in example

Felicious:
Fix typo found in the barcode of an example. Found by @odoo-shke 

Intended to be FWported to future versions


StraubCreative (approved):
@robodoo r+

PR #6982, opened 2023-12-19
[IMP] localization: AU rebrands Keypay to Employment Hero

tong-odoo:
task-3250954


tong-odoo:
@fw-bot up to 13.0

toaa-odoo:
Hi @tong-odoo, is there any particular reason you are targeting 13.0? AFAIK it does not build under 15.0 anymore

tong-odoo:
Hi @toaa-odoo, it is because customers are checking on 13.0 documentation and found it quite confusing with the old name still existing.

toaa-odoo (approved):

jcs-odoo:
I'm afraid the documentation is not supported and not built anymore for any version below 15.0.
Unless I'm mistaken, @AntoineVDV and the pages are still built?

tong-odoo:
@jcs-odoo What should we do for version below 15.0 in this case? Because customers are confusing and keep asking our sales about Keypay and Employment Hero

jcs-odoo:
> @jcs-odoo What should we do for version below 15.0 in this case? Because customers are confusing and keep asking our sales about Keypay and Employment Hero. Could we disable the access for 13.0?

How do they access this old documentation, then? Normally it shouldn't be found with a search on Google.

It's a recurring question: we'd rather archive the old doc, but it remains online nonetheless, I'm afraid.

tong-odoo:
> > @jcs-odoo What should we do for version below 15.0 in this case? Because customers are confusing and keep asking our sales about Keypay and Employment Hero. Could we disable the access for 13.0?
> 
> How do they access this old documentation, then? Normally it shouldn't be found with a search on Google.
> 
> It's a recurring question: we'd rather archive the old doc, but it remains online nonetheless, I'm afraid.

I think the page link is from the KeyPay website. I will try to talk to them to update the link first. Thank you so much for your help

jcs-odoo:
Or maybe not Payroll, after all ^^

Issue #7210, opened 2023-11-27
Odoo is by default logging to stderr instead of stdout


kajedot:
Impacted versions: 14.x, 15.x, 16.x, 17.x

Steps to reproduce: 

Run odoo with commands:
`python odoo-bin 2>stderr.txt `
`python odoo-bin 1>stdout.txt`

Current behavior:
File stdout.txt is empty. All log levels (DEBUG and INFO for sure) are logged to stderr. This is the inconsistency with the Odoo [documentation](https://www.odoo.com/documentation/17.0/developer/reference/cli.html?highlight=logging#logging): "By default, Odoo displays all logging of [level](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel) info except for workflow logging (warning only), and log output is sent to **stdout**"

Expected behavior:
Assuming there is no mistake in the documentation, Odoo should log every message to the stdout stream (so in my case stdout.txt should be populated). But considering logging module which Odoo is using, which sends everything to stderr, it is a mistake in documentation. 

Disclaimer: 
`1>filename   # Redirect stdout to file "filename."`
`2>filename   # Redirect stderr to file "filename."`
Reference: https://tldp.org/LDP/abs/html/io-redirection.html.

Julien00859:
Indeed, we always only log on stderr, that's the default behavior of the python StreamHandler class that we are using. We should correct the documentation, I'll do it.

https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler 
https://github.com/odoo/odoo/blob/b3b9b2b49ef90d30bcce1e27a4ba0b24f888170c/odoo/netsvc.py#L156
https://github.com/odoo/odoo/blob/b3b9b2b49ef90d30bcce1e27a4ba0b24f888170c/odoo/netsvc.py#L198

PR #7253, opened 2024-01-04
[FIX] developer/reference/cli: log on stderr

Julien00859:
Odoo is actually logging on sys.stderr and not sys.stdout[^1][^2][^3].

Remove passage about the workflows, the workflows were an OpenERP feature removed almost a decade ago in odoo/odoo@98cb4719dbe90db.

Fixes: #7210

[^1]: https://github.com/odoo/odoo/blob/b3b9b2b49ef90d30bcce1e27a4ba0b24f888170c/odoo/netsvc.py#L156
[^2]: https://github.com/odoo/odoo/blob/b3b9b2b49ef90d30bcce1e27a4ba0b24f888170c/odoo/netsvc.py#L198
[^3]: https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler

### DIFF
--- a/content/developer/reference/cli.rst
+++ b/content/developer/reference/cli.rst
@@ -450,14 +450,13 @@ HTTP
 Logging
 ~~~~~~~
 
-By default, Odoo displays all logging of level_ ``info`` except for workflow
-logging (``warning`` only), and log output is sent to ``stdout``. Various
-options are available to redirect logging to other destinations and to
-customize the amount of logging output.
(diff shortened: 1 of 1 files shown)


Julien00859:
I think it will, we'll see

AntoineVDV:
At least the link that I created from the interface has no effect; so if it works it's due to the commit message.

Julien00859:
Yup, it is the commit message that will close the issue once merge, as it'll close this PR thanks to mergebot rewriting the message to include `close #7253`

PR #7403, opened 2024-01-16
[FIX] requirements: fix sphinxcontrib-applehelp version

d-fence:
It seems that the latest version if this lib (1.0.8) needs sphinx 5.0.


d-fence:
robodoo r+ p=0

PR #7251, opened 2024-01-04
[IMP] iot: several printers per report

viforget:
Before we could only associate one printer per reports,
Now we can associate several, a pop-up will appear for the first print to ask you to select which of the associated printers you want to print on, the selection will stay in memory in the local storage and the pop-up wont appear again.
To change the selection of printers, you can go in Clear Selected Devices to remove the selection from the local storage, doing so, the pop-up will appear the next time you'll print this report

PR #7431, opened 2024-01-17
[IMP] IOT Printer Doc Update

tiku-odoo:
Add clarification per: 
https://github.com/odoo/documentation/pull/7251/files

Task: https://www.odoo.com/web#cids=3&menu_id=4720&action=333&active_id=3835&model=project.task&view_type=form&id=3690324

### DIFF
--- a/content/applications/productivity/iot/devices/printer.rst
+++ b/content/applications/productivity/iot/devices/printer.rst
@@ -108,8 +108,9 @@ should be linked to this printer.
    :align: center
    :alt: The printer devices listed in the IoT Devices menu.
 
-Now, each time :guilabel:`Print` is selected in the control panel, instead of downloading a PDF,
-Odoo sends the report to the selected printer, and automatically prints it.
+Now, each time :guilabel:`Print` is selected in the control panel, instead of downloading a PDF, a
+pop-up appears which displays all the printer(s) linked to the report. Then Odoo sends the report to
+the selected printer(s), and automatically prints it.
 
 .. seealso::
    :doc:`POS Order Printing <../../../sales/point_of_sale/restaurant/kitchen_printing>`


tiku-odoo:
@lse-odoo 

Hi Loan, 
Can you review this doc when you have a chance? Just added some additional information per: 

https://github.com/odoo/documentation/pull/7251

lse-odoo:
@tiku-odoo I am not sure of the behavior in version < 17.0, I guess it would be better to ask to the functional support the intended behavior in this versions :/

tiku-odoo:
@wigu-odoo 

Can you review this doc when you have a moment?

wigu-odoo:
Hello @tiku-odoo 

I will test again on Monday when I am back in the office (thanks snowstorms) but it does look mostly correct! 

The one thing I did notice from when I tested this about a month ago in a ticket that I needed to escalate was that the popup to select printer does not appear unless I have developer mode active. 

In the ticket I have a video from runbot when I have dev mode enabled as well as a screenshot from runbot and in the screenshot, I do not get the popup since I don't have developer mode enabled, but in my recording, I do when I have developer mode enabled. 

The debug icon is also present in my recording with this popup. 

I am pretty sure developer mode needs to be enabled for the popup to select the printer, but I will test more when I am in the office on Monday to confirm.

lse-odoo:
@wigu-odoo if I am not mistaken, the pop-up show once then save into the browser cookie your result. So maybe the fact to switch to developer mode is not true and made you think that it was needed

wigu-odoo:
@lse-odoo You were indeed correct, developer mode is not necessary for the popup. 

@tiku-odoo this is indeed correct!

wigu-odoo (approved):
just tested to be sure developer mode is not necessary, everything looks correct!

StraubCreative (approved):
LGTM 👍 

@robodoo r+

PR #7510, opened 2024-01-23
[FIX] reference/user_interface: filter element name is mandatory

JulienAlardot:
The filter element attribute 'name' is now required since https://github.com/odoo/odoo/commit/4ddc3231

image


AntoineVDV (approved):
@robodoo r+

PR #7301, opened 2024-01-08
[ADD] Time Off: Adding new doc

larm-odoo:
No documentation exists for Time Off. Forward port this to V 16 and 17, then targeted edits can be made.

PR #7515, opened 2024-01-23
[FIX][CSS]: ul ol sub-bullet margins

samueljlieber:
This PR fixes "simple" bullet list spacing within a non-simple list.

docs-bullet-spacing-fs8

This issue was discovered in https://github.com/odoo/documentation/pull/7301#discussion_r1462524487.

The added CSS targets ⬇️ to add a bottom margin of 1em to the `<p>` element.
```
<ul or ol>
  <li>
    <ul class="simple">
      <li :last-child>
        <p>
```

### DIFF
--- a/extensions/odoo_theme/static/scss/_sphinx_basic.scss
+++ b/extensions/odoo_theme/static/scss/_sphinx_basic.scss
@@ -615,6 +615,11 @@ ul.simple p {
     margin-bottom: 0;
 }
 
+ol > li > ol.simple > li:last-child > p,
+ul > li > ul.simple > li:last-child > p {
+    margin-bottom: 1em;
+}
+
 dl.footnote > dt,
 dl.citation > dt {
     float: left;


extensions/odoo_theme/static/scss/_sphinx_basic.scss
```diff
@@ -615,6 +615,11 @@ ul.simple p {
     margin-bottom: 0;
 }
 
+ol > li > ol.simple > li:last-child > p,
+ul > li > ul.simple > li:last-child > p {
```
Brieuc-brd, line 619: Hello 👋 
What happens if the element inside the last item is not a paragraph? (e.g. a link, a button, an image, etc.) ?
Not sure it will work 🤔 .

What you could do is apply the margin to the `ul`.

```suggestion
li > .simple {
```

☝️ This way, you'll select all `simple` lists which are in a parent list and apply on it a margin-bottom.

You could go a step further and make the selector more generic. Instead of using a class (here `.simple`) to target lists within another list, you could use this (to be tested) :

```suggestion
li > ul, li > ol {
```

Brieuc-brd: Can you tell me how I can test your branch? :)

samueljlieber: Hi @Brieuc-brd I had to make a small adjustment to my styles, it didnt account for each scenario, it is now:
```
ol.simple > li > p,
ul.simple > li > p,
ol.simple > li > ul > li > p,
ul.simple > li > ul > li > p,
ol.simple > li > ol > li > p,
ul.simple > li > ol > li > p {
    margin-top: 0;
    margin-bottom: 0;
}

ul > li > ul.simple > li:last-child > p,
ol > li > ul.simple > li:last-child > p {
    margin-bottom: 1em;
}
```

You can test this by either pulling down this branch `15.0-scss-sub-bullet-fix-sali` to your local computer by following [these instructions](https://github.com/odoo/documentation/blob/17.0/README.md), or open this branch on runbot to [view the built documentation with these styles](https://runbot207.odoo.com/runbot/static/build/57847964-15-0/logs/build/html/applications/hr/time_off.html#time-off-requests-section).

Once viewing the documentation, navigate to the **Human Resources > Time Off** page and scroll to the **Time off requests section**.  Here you will see a non-simple `ul` with a sub-`ul` that is `.simple`.

Then further down in the document are different variations of these list types.

You can also look at the Table of Contents files to verify that these style changes are not affecting them.

I believe that these style changes fix the issue that this PR aims to address, let me know your thoughts! 
Thanks 🙏

extensions/odoo_theme/static/scss/_sphinx_basic.scss
```diff
@@ -598,21 +598,19 @@ ol.upperroman {
...
+ul.simple > li > ol > li > p {
     margin-top: 0;
+    margin-bottom: 0;
 }
 
-ol.simple p,
-ul.simple p {
-    margin-bottom: 0;
+ul > li > ul.simple > li:last-child > p,
+ol > li > ul.simple > li:last-child > p {
```
Brieuc-brd, line 612: As I said before, you can cover this case with a simpler selector :
```suggestion
li > .simple {
```
By doing this, it's easier to maintains the `.simple` spacing. 
You don't need to specify the paragraph margin to solve this issue.

Brieuc-brd (changes requested):
Hey 👋 
> So the issue this PR is addressing is the lack of bottom margin on the last <p> tag in a .simple list.

There is indeed a lack of bottom margin for `.simple` lists, but to solve this, you don't have to apply the margins on the last `p`. You can target a simpler selector that will be much easier to maintain 😉 (comments below).

extensions/odoo_theme/static/scss/_sphinx_basic.scss
```diff
@@ -598,23 +598,15 @@ ol.upperroman {
...
-
 ol.simple p,
 ul.simple p {
     margin-bottom: 0;
 }
 
+li > .simple {
+    margin-bottom: 1em;
+}
+
```
Brieuc-brd, line 609: We could go further, I'd use `.simple` as the main selector and use the power of SCSS 🙂 
```scss
.simple {
    p {
        margin-bottom: 0;
    }
    
    li > & {
        margin-bottom: 1em;
    }
}
```
But, again, is the `simple` class used on other elements than `ul` & `ol` ? @AntoineVDV maybe you know it? 🙂

extensions/odoo_theme/static/scss/_sphinx_basic.scss
```diff
@@ -902,4 +894,4 @@ div.math:hover a.headerlink {
     #top-link {
         display: none;
     }
-}
+}
```
Brieuc-brd, line 897: You should keep the empty line at the end of the file 😉

extensions/odoo_theme/static/scss/_sphinx_basic.scss
AntoineVDV: No idea^^

samueljlieber: @Brieuc-brd `.simple` is also used on `dl` elements. [This SCSS](https://github.com/odoo/documentation/pull/7515#discussion_r1486018508) works well for the `dl` element too. 

I searched through the documentation and did not find any other instances of where the class `.simple` is used on other elements. This [docutils Examples of Syntax Constructs](https://docutils.sourceforge.io/docs/user/rst/demo.html#table-of-contents) document also only shows the class being used on only the `ul`, `ol`, and `dl` elements.

The `dl`, `dt`, and `dd`  elements are heavily used on the `applications/finance/accounting/get_started/cheat_sheet.rst` doc [for example](https://www.odoo.com/documentation/15.0/applications/finance/accounting/get_started/cheat_sheet.html#accounting-cheat-sheet).

I will push up a commit to update the CSS to SCSS since it is a more elegant solution 👍

Brieuc-brd (approved):
LGTM 👍

AntoineVDV (approved):
Thanks @Brieuc-brd for reviewing :)

This also adds a margin to [nested lists that are the last list item of the parent list](https://www.odoo.com/documentation/15.0/contributing/documentation/rst_cheat_sheet.html#nested-lists), but I think that it's okay.

Could you just change the commit message to something more Odooesque? For example `[FIX] odoo_theme: add bottom margin to inner lists`

@robodoo delegate+

samueljlieber:
Updated commit message in https://github.com/odoo/documentation/commit/38dcb529b0f8e28c858470291ef45a90034aaf65.

Thank you both for the insight and reviews 🙏

@robodoo r+

PR #7660, opened 2024-02-07
[IMP] iot: add documentation regarding IoT updates

lse-odoo:
Would be more convenient for the support to avoid repeating the same instructions when they need to perform a certain kind of update


lse-odoo:
@tiku-odoo 
Could you please do the necessary to adapt the PR to Odoo documentation standard please ? :)
I quickly add the main part, feel free to force push on it without explanations (except if you have a doubt on some points)

PR #7664, opened 2024-02-08
[ADD] developer/frontend: beware orphan assets

jdkirkwood:
Add note that orphan asset records may be left after removing/renaming via the manifest - causing bundle compilation errors.

Task: [3336371](https://www.odoo.com/web#id=3336371&model=project.task&view_type=form)


jdkirkwood:
Too specific to upgrades.

PR #7678, opened 2024-02-09
Update source.rst

sarmad90:
The usage of however seems incorrect. A small edit to make to improve readability and clarity.


xpl-odoo:
Thanks for the suggestion @sarmad90.
While I'm not the original writer of this documentation page, the phrasing using "however" seems clear and correct to me, so I'll leave it as it is.

PR #7715, opened 2024-02-13
[FIX] payment_providers: add payment journal selection

auva-odoo:
task-3718255

Up to saas-16.4


auva-odoo:
@robodoo r+
@fw-bot up to saas-16.4

PR #7749, opened 2024-02-15
[IMP] developer/reference/cli: --smtp-ssl [encryption]

Julien00859:
The `--smtp-ssl` flag can now work as a regular `--option value` option where the value can be any value from the `smtp_encryption` field of `ir.mail_server`.

community: odoo/odoo#154077
task-2861790

### DIFF
--- a/content/developer/reference/cli.rst
+++ b/content/developer/reference/cli.rst
@@ -315,9 +315,18 @@ Emails
 
 .. option:: --smtp-port <port>
 
-.. option:: --smtp-ssl
+.. option:: --smtp-ssl [encryption]
 
-    If set, odoo should use SSL/STARTSSL SMTP connections
+    Connection encryption, accepted values are:
+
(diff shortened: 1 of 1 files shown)


content/developer/reference/cli.rst
```diff
@@ -314,9 +314,18 @@ Emails
 
 .. option:: --smtp-port <port>
 
-.. option:: --smtp-ssl
+.. option:: --smtp-ssl [encryption]
 
-    If set, odoo should use SSL/STARTSSL SMTP connections
+    Connection encryption, accepted values are:
```
AntoineVDV, line 319: The "if set" start of the previous version was helping understand that this flag is not required.

```suggestion
    If specified, this flag indicates the connection encryption. Accepted values are:
```

AntoineVDV (approved):
@robodoo delegate+

PR #7887, opened 2024-02-26
Fix snippets templates url in building_blocks.rst

zeroheure:
the file snippets.xml (/addons/website/views/snippets/snippets.xml) contains the snippets options and is wrongly linked as snippets templates


zeroheure:
The same can be applied to previous documentation versions

PR #8004, opened 2024-03-05
[FIX] deploy: enable HSTS also for websocket

Julien00859:
Fine tunning of 6a2725e604


Julien00859:
https://github.com/odoo/documentation/pull/6086#issuecomment-1974156278

> Julien00859:
> mart-e why not include HSTS to the /websocket server block too?

> mart-e:
> Julien00859 no reason, we should probably add it too

mart-e (approved):

mart-e:
@robodoo r+

PR #8129, opened 2024-03-13
[IMP] web: create how-to for using Owl components in the frontend

sdegueldre:


content/developer/howtos.rst
```diff
@@ -50,6 +51,11 @@ How-to guides
       Learn how to create a public-facing Owl application outside of the web client using a
       controller and the Odoo JavaScript framework.
 
+   .. card:: Using Owl components in the portal and website
```
AntoineVDV, line 54: Suggestion for consistency but both are fine.
```suggestion
   .. card:: Use Owl components in the portal and website
```

content/developer/howtos/frontend_owl_components.rst
```diff
@@ -0,0 +1,187 @@
...
+        # ...
+        'data': {
+            'views/templates.xml',
+        }
+    }
+
+And that's it! If you open the home page of the portal you should see the message
+"Hello, World!" at the top of the page.
+
+Problems with using Owl components in the portal and website
```
AntoineVDV, line 110: There is no need to repeat the page's title. The title could be shorter, something like
```suggestion
Points of caution
```

content/developer/howtos/frontend_owl_components.rst
```diff
@@ -0,0 +1,187 @@
...
+
+This can be a frustrating experience, so you should be careful when designing your
+page that the Owl component will not move elements around. This can be achieved in
+various ways, e.g. by positioning it below all other existing elements, by not having
+other interactive elements around, or by reserving a fixed space for the Owl component
+using CSS.
+
+
+.. seealso::
+    `cumulative layout shift on web.dev <https://web.dev/articles/cls>`_
```
AntoineVDV, line 144:
```suggestion
  `Cumulative Layout Shift on web.dev <https://web.dev/articles/cls>`_
```

content/developer/howtos/frontend_owl_components.rst
```diff
@@ -0,0 +1,187 @@
...
+and element in the page that was initially rendered and with which they want to interact,
+so they move their cursor or finger above that element. Just as they're about to click,
+the Owl component is mounted and the element with which they wanted to interact moved.
+They click and interact with the Owl app instead.
+
+This can be a frustrating experience, so you should be careful when designing your
+page that the Owl component will not move elements around. This can be achieved in
+various ways, e.g. by positioning it below all other existing elements, by not having
+other interactive elements around, or by reserving a fixed space for the Owl component
+using CSS.
```
AntoineVDV, line 140:
```suggestion
When a page initially renders content, and that content subsequently moves ("shifts")
within the page, this is referred to as a layout shift. When using Owl components in
the portal or website, all of the HTML that surrounds the Owl component is rendered
by the server and is the first thing that will be displayed to the user. When JavaScript
starts running, Owl will mount your component, which will likely cause the surrounding
elements to move around on the page. This can cause poor user experience: the user sees
an element on the page that was initially rendered and with which they want to interact,
so they move their cursor or finger above that element. Just as they're about to click,
the Owl component is mounted, and the element they want to interact with is moved.
They click and interact with the Owl app instead.

This can be a frustrating experience, so you should be careful when designing your
page that the Owl component will not move elements around. This can be achieved in
various ways, e.g. by positioning it below all other existing elements, not having
other interactive elements around, or reserving a fixed space for the Owl component
using CSS.
```

content/developer/howtos/frontend_owl_components.rst
```diff
@@ -0,0 +1,187 @@
...
+to find pages and analyze their content so that they can show these pages in their
+search results. While modern search engines are generally capable of executing JavaScript
+code, and should generally be able to view and index content rendered in JavaScript,
+they may not index the content as fast and may penalize the page in search results.
+
+Because most search engines do not reveal the exact way that they crawl and index
+web pages, it's not always easy to know the extent of the impact client side rendering
+can have on your search engine scores. While it's unlikely to make or break your SEO
+strategy, you should still only use Owl components when they are adding real value
+over server-side rendering.
```
AntoineVDV, line 160:
```suggestion
When search engines build their index of the content of the web, they use web crawlers
to find pages and analyze their content to show these pages in their
search results. While modern search engines are generally capable of executing JavaScript
code and should generally be able to view and index content rendered in JavaScript,
they may not index the content as fast and penalize the page in search results.

Because most search engines do not reveal the exact way that they crawl and index
web pages, it's not always easy to know the extent of the impact that client-side rendering
can have on your search engine scores. While it's unlikely to make or break your SEO
strategy, you should still only use Owl components when they are adding real value
over server-side rendering.
```

content/developer/howtos/frontend_owl_components.rst
```diff
@@ -0,0 +1,187 @@
...
+
+When to use Owl components on the portal and website
+====================================================
+
+As explained in the previous sections, using Owl component can slightly degrade user
+experience if you're not careful, and may also hinder your SEO. So when should you
+choose to use Owl components in these places? Here are some general guidelines.
```
AntoineVDV, line 167:
```suggestion
As explained in the previous sections, using Owl component can slightly degrade user
experience if you're not careful and may also hinder your SEO. So when should you
choose to use Owl components in these places? Here are some general guidelines.
```

content/developer/howtos/frontend_owl_components.rst
```diff
@@ -0,0 +1,187 @@
...
+
+When you need strong interactivity
+----------------------------------
+
+The decision to use Owl is a trade-off between the previously mentioned disadvantages,
+and effort that Owl saves you by making it easy to build richly interactive user
+experiences. The main reason to use Owl is when you want to build an interface that
+can react in real time to user inputs wihout requiring the page to reload. If you
+mainly want to show static content to the user, you probably shouldn't use an Owl
+component.
```
AntoineVDV, line 187:
```suggestion
The decision to use Owl is a trade-off between the previously mentioned disadvantages
and the effort that Owl saves you by making it easy to build richly interactive user
experiences. The main reason to use Owl is when you want to build an interface that
can react in real time to user inputs without requiring the page to reload. If you
mainly want to show static content to the user, you probably shouldn't use an Owl
component.
```

AntoineVDV (approved):
Interesting read :)

I assume I can assign the JS team as a codeowner for this page (done)?

@robodoo delegate+

sdegueldre:
@AntoineVDV yep, thanks for assigning codeowners and for the quick review! Comments have been applied.

@robodoo r+

PR #8131, opened 2024-03-13
[IMP] core: remove user_has_groups()

rco-odoo:
Simply align the documentation with the new API and usage of method user.has_group().

see: https://github.com/odoo/odoo/pull/151597


AntoineVDV (approved):
@robodoo r+ delegate+

rco-odoo:
@robodoo r+

PR #8172, opened 2024-03-15
[FIX] website: anchors in analytics hyperlinks

jcs-odoo:
Custom anchors can be used as hyperlinks to target a specific heading on the page. However, the google analytics one didn't work well because of the capital letters used.


d-fence:
robodoo retry
Failing performance test is now deactivated

PR #8193, opened 2024-03-17
[FIX] dev/view_architectures: domain attribute of search view fields

glovebx:
The domain in the `field` component cannot accept user-provided values or access self variables.
This commit removes the misleading sentence about it from the documentation.


Feyensv (approved):
@robodoo r+ squash

Thanks for the correction :+1:

PR #8196, opened 2024-03-18
[ADD] Website: inner content building blocks

Donapi:
taskid-3733888


content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
```diff
@@ -0,0 +1,39 @@
+=============================
+Inner content building blocks
+=============================
+
+You can use the :guilabel:`Inner content` building blocks, such as the :guilabel:`Button`,
+:guilabel:`Image`, :guilabel:`Video`, or  :ref:`Social Media <inner_content_blocks/social_media>`
+blocks, to enhance your website's view by providing a structure and design for the content available
+on your web pages. Select a block from the :guilabel:`Inner content` section in the
+:guilabel:`Blocks` tab of the website builder, and drag and drop it onto your web page.
```
LoredanaLrpz, line 9: I think there's too much information in the first sentence when it should provide "this is what we're talking about" 
Here's a rewriting suggestion, I just separated into two paragraphs to make it a bit clearer:
```suggestion
The inner content building blocks allow you to add elements such as videos, images, buttons,
:ref:`Social Media <inner_content_blocks/social_media>` redirection buttons, and more into
pre-existing structure blocks.

To do so, access a website page and open the website builder by clicking :guilabel:`Edit`. Then,
drag and drop a block from the :guilabel:`Inner content` section of the :guilabel:`Blocks` tab onto
your web page.
```
I'd remove the guilabel when talking about images and stuff because this is more the concept, readers don't know yet where to look so they don't see image or video written on their screen

content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
```diff
@@ -0,0 +1,39 @@
...
+.. _inner_content_blocks/social_media:
+
+Social Media block
+------------------
+
+Displaying your company's social media on your website is a good manner to grow brand awareness and
+improve :doc:`SEO <../../../website/pages/seo>`.
+
+Select the :guilabel:`Social Media` building block from the :guilabel:`Inner content` section and
+drag and drop it anywhere on your website, including in existing building blocks.
```
LoredanaLrpz, line 26: If you don't drag and drop into existing building blocks, it works, but the button goes to the footer -I think there should be a note about that -either here or in the introduction (probably better in the intro -or configuration depending on what you do x))

content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
LoredanaLrpz, line 1: I think there should be a "configuration" h2 with the second part of the intro (what I suggested) and social media block could just be an example, I don't understand why this deserves a title and not all the other blocks

content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
```diff
@@ -0,0 +1,39 @@
...
+
+.. image:: inner_content_blocks/social_media_block.png
+  :alt: The social media block.
+
+Once added to your web page, select the :guilabel:`Social Media` block, go to the
+:guilabel:`Customize` tab, and go to the :guilabel:`Social Media` section to edit it.
```
LoredanaLrpz, line 32: Same as the general comment: this part could go into the config h2 and this paragraph could be part of the social media example

LoredanaLrpz (changes requested):
There you go :) Do not hesitate if you have any questions!

content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
Donapi: I do not think a note about that would add any value to the documentation. The inner blocks are meant to be added to existing building blocks, as explained in the intro. If users do not drag and drop them to existing blocks (I'm not sure in which case that would be their intention...?), they will see they go to the footer.

content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
Donapi: The idea is to be consistent with the other building blocks pages (Structure, Features, Dynamic pages that are being created). We decided to only present for each category of building block one or more blocks. In the future, we can add additional blocks if it makes sense.

content/applications/websites/website/web_design/building_blocks/inner_content_blocks.rst
Donapi: As explained before, the idea is not to have a config section to be consistent with the other pages.

LoredanaLrpz (approved):
<3

xpl-odoo:
@robodoo r+

PR #8197, opened 2024-03-18
[REM] crm: track prospects visits

jero-odoo:
The information in this document is no longer relevant in V15+ as the Lead Scoring module was deprecated. Removed this document and added a redirect.
https://www.odoo.com/documentation/15.0/applications/sales/crm/track_leads/prospect_visits.html

Also removed doc link (and reference to the deprecated module) from Link Trackers doc https://www.odoo.com/documentation/17.0/applications/websites/website/reporting/link_tracker.html#follow-up-on-tracked-links

Should only need a tech review since there were no content changes.


jero-odoo:
@samueljlieber I think this can go right to tech review, since content was only removed, not added/updated.

redirects/15.0.txt
```diff
@@ -168,6 +168,7 @@ applications/general/voip/voip_widget.rst applications/productivity/voip/voip_wi
 # applications/sales
 
 applications/sales/crm/optimize/outlook_extension.rst applications/general/integrations/mail_plugins/outlook.rst
+applications/sales/crm/track_leads/prospect_visits.rst applications/sales/crm/track_leads.rst                                   #track_leads/prospect_visits -> track_leads
```
samueljlieber, line 171: Super small nitpick, add a space in the comment please :)
```suggestion
applications/sales/crm/track_leads/prospect_visits.rst applications/sales/crm/track_leads.rst                                   # track_leads/prospect_visits -> track_leads
```

samueljlieber (approved):
Hi @jero-odoo, nice job– this PR looks good to me! One quick adjustment and you can pass to ZST. Thank you!

jero-odoo:
@StraubCreative This is ready for you to review. (Only removing content, not adding or updating anything)

StraubCreative (approved):
Nice catch, approved @jero-odoo 👍 
..
@robodoo r+

xpl-odoo:
@robodoo r+

xpl-odoo:
@robodoo r-

xpl-odoo:
@robodoo r+

xpl-odoo:
Hello @AntoineVDV !
Any idea why this PR is "stuck"?

AntoineVDV:
Absolutely not :man_shrugging:

@Xavier-Do I thought that the CI status was not sent, so I resent it from the commit page, but it doesn't seem to be the problem.
image
I started a new build and now it seems to be the Python step that is crashing.

Xavier-Do:
The python step is crashing because the automatic rebase failed. 

You didn't rebuild, you started a [new batch with automatic rebase](https://runbot.odoo.com/runbot/batch/1402260), a feature that is hidden for most dev. Why did you use it?

The message you received is not clear, but give the relevant information

```
Apply patch failed for 363d0a582f4b449e313aef2249bcf13615dcd624...e4633374c27e34405fb28d3ef8bc7733b9977263 with error code 0+1. (b'patching file content/applications/sales/crm/track_leads.rst
...
```

Anyways, this means that this pr needs a manual rebase. 

I created a [new batch](https://runbot.odoo.com/runbot/batch/1403433) without automatic rebase, it is green 

Resending the status could have been enough but the automatic rebase and stuff made too much noise to investigate. Anyway, i'll hide those button to avoid confusion. 

This pr have a good change to fail because robodoo won't be able to cherrypick it on top of the current 15.0, not 100% sure cherrypick are more robust than apply patch but this is a possibility and it will need a manual rebase.

Xavier-Do:
> Because I couldn't find a clear explanation in the logs, and it has often fixed unclear issues to start an automatic rebase when a simple new batch didn't do the trick.

The reason was most likely an issue with github, we could have investigated if you asked us at this point if the resend didn't work. Resend should have been enough, if it wasn't the case any other operation could have failed at the same point when sending the status.

> Rather than hiding a feature that saves time for both reviewers and PR authors, I suggest doing the opposite: fix issues if any and add warnings if necessary.

The issue is that there is too much actions in the interface confusing dev, this is why we hide it. The rebuild is accessible to everyone in the dropdown menu since it is what you intended to do.

This new batch action and especially with automatic rebase creates more load than needed in most cases, and has no guarantee to have the expected result. The rebuild button is accessible for everyone, we added a group to hide the other one since they are confusing and needed in rare specific cases, mostly after changing configurations on runbot, willing to avoid loosing a r+ in one repo when rebasing another one, or to merge pr from the community opened outside odoo-dev.  
This is unfortunately randomly fixing issues by luck when used blindly, this is why we prefer to limit this feature only to user who need it and know when to use it waiting for a clearer integration.

Xavier-Do:
> devs are asking each other to use this or that runbot feature on their PR because the other guy belongs to the right permission group

Can you give an example ?

AntoineVDV:
> needed in rare specific cases, mostly after changing configurations on runbot, willing to avoid loosing a r+ in one repo when rebasing another one, or to merge pr from the community opened outside odoo-dev

I think you strongly underestimate the benefits of this particular feature. I agree that it is not that useful for the author of the bundle, but it's a completely different story for reviewers:
- With the increasing number of both linked repos and newbie developers, mismatching bases are way more common than you think, certainly not "rare".
- Rebasing PRs of the community is an absolute PITA because we need to add the remote, checkout locally, rebase, force-push, etc. It takes time and isn't always possible, for example when the remote belongs to an organization or when the author doesn't allow us to push on their branch.
- Last but not least, bugs fixed in later commits ("this error has already been fixed"), runbot exceptions removed, etc. all create the need to rebase. When a bundle has 3-4 branches, the last thing you want is rebasing all of that and re-r+ everything again when there exists a feature that does exactly that for you.

If it were really useless, I wouldn't be arguing over this and asking to give reviewers access. It fills its role just fine (where a regular rebuild could not)  99% of the time; does it make sense to require extra manual work from everyone to spare some load for the remaining 1%?

> Can you give an example ?

I can only speak for myself, but the top 3 are the automatic rebase, disabling codeowners on noisy (and not interesting) PRs, and updating codeowners (through regex and/or `runbot.team`). Others have other needs, but this is regularly discussed.

Xavier-Do:
> With the increasing number of both linked repos and newbie developers, mismatching bases are way more common than you think, certainly not "rare".

The issue with the automatic rebase is that it doesn't work all the time, like here, you said "the step failed" when it was the automatic rebase. Moreover then automatic rebase is done only one and is lost on first push. 

We are planning to forbid building if base mismatches, it is the dev task to keep base synced, or it could also fail locally, and a dev should start by testing locally.  But if you think it will be really needed, I can continue to work on my branch that will make the rebase more robust (after docker registrie and other priorities), and in this case we may propose to it in this specific case: if bases do not match, no build start, but anyone can start them manually from the interface with an autofix.

> Rebasing PRs of the community is an absolute PITA because we need to add the remote, checkout locally, rebase, force-push, etc. It takes time and isn't always possible, for example when the remote belongs to an organization or when the author doesn't allow us to push on their branch.

For this case specifically it is now open to every user, no need to have the specific group to have the button. 

> Last but not least, bugs fixed in later commits ("this error has already been fixed"), runbot exceptions removed, etc. all create the need to rebase. When a bundle has 3-4 branches, the last thing you want is rebasing all of that and re-r+ everything again when there exists a feature that does exactly that for you.

For this one it is sometimes required, but even if it is marked as already known it may be a real error that reappeared. It is usually better to ask us in this case because it will also give us feedback that we may have re-enable an test to early or that we should have a way to keep a test disable if the merge-base is older than the fix, .... Also if random and already know, it can still be random, and rebuilding may be enough and way faster. We are usually use to take those decision.  

> If it were really useless, I wouldn't be arguing over this and asking to give reviewers access. It fills its role just fine (where a regular rebuild could not) 99% of the time; does it make sense to require extra manual work from everyone to spare some load for the remaining 1%?

The load is actually more than you imagine I think, because in some case you could just restart a 30 minutes build when this button will create 10 hour of load. It is ok when needed but as in you last example it was not the case. And it's maybe not hudge, but people think that is is the way to go, create load and then complain when it doesn't work, making the issue harder to understand when they come to us. It is usually less work for us to check the specific case and do it ourselves when it is needed, and also ensure that we are aware of existing issues and possible improvement.

> I can only speak for myself, but the top 3 are the automatic rebase, disabling codeowners on noisy (and not interesting) PRs, and updating codeowners (through regex and/or runbot.team). Others have other needs, but this is regularly discussed.

I took a look at the last one you did

- https://runbot.odoo.com/runbot/bundle/master-payment-remove-view-configuration-fields-matp-209792

I don't get why, if it was a missing ci a resend should have been enough and if not you should contact us because it could hide a real issue. 

- https://runbot.odoo.com/runbot/batch/1402260
Same here but more than not solving the issue, it failed. 

- https://runbot.odoo.com/runbot/batch/1398067
- https://runbot.odoo.com/runbot/batch/1396014
External prs, should be now possible to autorebase them without having the group. 

- https://runbot.odoo.com/runbot/bundle/saas-17-2-17-0-sale-fix-unsudoed-invoice-creation-vcr-8ydf-fw-227736
Looks like it is for an upgrade error. This one was actually linked to the freeze, we add to add an exception for it
https://runbot.odoo.com/web/#id=313&menu_id=495&cids=1&action=758&model=runbot.upgrade.exception&view_type=form
So the autorebase worked by chance, a rebuild was enough since we added an exception for that. More that that, it is a single branch in our repo, a manual rebase was not so hard to do and would have ensure that following builds were also up to date if a change was needed. 

So looking at those last one (except external prs) it looks like you think it was needed, when it actually wasn't the case and it worked by chance.

AntoineVDV:
I'm not particularly a fan of buttons *per se*, but I just cannot afford the time for the extra manual work that is often required anymore. So, if the load itself is an argument against this feature, individual fixes for each of the problems mentioned above would be a viable solution :+1:

However, even though I'm okay with notifying you each time I spot something suspicious, I don't think that waiting for a reply is the way to go; the added bureaucracy would just slow down a merge process that is already quite heavy. Instead, problem-specific solutions (like those you mentioned, or a per-repo "sync head" button, for example) would allow reviewers to keep their autonomy while addressing your main concerns.

PR #8242, opened 2024-03-20
[I18N] *: export latest terms + add missing resources

ticodoo:
This commit does 2 things:

- exports the latest terms to be translated
- adds missing pot files + adds them to be translated on transifex (essentials, hr, and studio)


mart-e (approved):
@robodoo r+

mart-e:
robodoo r+

PR #8293, opened 2024-03-21
[FIX] subscriptions: updated and corrected mistake re: eLearning

ksc-odoo:
PROJECT TASK: https://www.odoo.com/web#id=3811470&cids=3&model=project.task&view_type=form

- fixed issue regarding eLearning, product variants, and recurring prices
- updated the doc with new screenshots and more accurate process


content/applications/sales/subscriptions/ecommerce.rst
```diff
@@ -19,33 +20,38 @@ recurrence period.
...
+To set up each recurrence period as a product variant, go to :menuselection:`Subscriptions app -->
+Products`, and select a product. In the :guilabel:`Attributes & Variants` tab, click :guilabel:`Add
+a line`.
 
-Create an :guilabel:`Attribute` called `Billing Period` (or something similar) by typing in the
-name and clicking :guilabel:`Create`. This attribute name appears as the option heading on the
-product page of the eCommerce shop.
+Then proceed to create an :guilabel:`Attribute` called `Billing Period` (or something similar), by
+typing in the name, and clicking :guilabel:`Create` from the mini drop-down menu that appears. This
+attribute name appears as option heading on the product page of the eCommerce shop.
```
Felicious, line 29: Totally personal preference, but when it comes to naming things, since users don't HAVE to call it `Billing Period`, let's try describing what they should type in, instead of telling them what to type.
```suggestion
Then proceed to create an :guilabel:`Attribute` by typing in a name to represent the billing period, and clicking :guilabel:`Create` from the mini drop-down menu that appears. This
attribute name appears as option heading on the product page of the eCommerce shop.
```

content/applications/sales/subscriptions/ecommerce.rst
```diff
@@ -19,33 +20,38 @@ recurrence period.
...
+:guilabel:`Recurring Prices` tab. Proceed to assign the different :guilabel:`Product Variants` to
+their corresponding recurrence :guilabel:`Period` and :guilabel:`Price`.
 
 .. image:: ecommerce/product-variants-time-based-pricing.png
    :align: center
    :alt: Product variants on the "Time-based pricing" tab of the product form.
 
-The product variants are now available for selection on the eCommerce product page.
+After following those aforementioned steps, the product variants are available for selection on the
+eCommerce product page.
```
Felicious, line 54: it might be helpful to include menuselection steps to get to the eCommerce product page?

Felicious (approved):
Hi @ksc-odoo ! These were good updates to this doc. I only had very minor comments, so feel free to move this onto the next stage after you address these!

content/applications/sales/subscriptions/ecommerce.rst
ksc-odoo: I totally understand (and appreciate) where you're coming from. I, too, feel that same way. Personally, I think my parenthetical suggestion of "(or something similar)" suffices in this regard, and doesn't make the user feel beholden to one *specific* title. So, I think I'm going to leave it as-is, but thanks for the suggestion. 👍

content/applications/sales/subscriptions/ecommerce.rst
ksc-odoo: This would be a difficult task, as they can implement the lesson from this doc on *any* subscription product, which could be on *any* page, of their online/eCommerce store. So, I think this more generic allusion to where it would be located is the best/most apt approach for this particular circumstance.

ksc-odoo:
Thank you for the always helpful feedback, @Felicious -- i made the necessary modifications, and believe this one is now ready for Tech Review from you, @samueljlieber -- whenever you get a chance! Thanks! 🙏

samueljlieber (approved):
Looks good to me @ksc-odoo, nice job  👍
..
@robodoo r+

PR #8409, opened 2024-03-28
[IMP] website: domain names CNAME records

xpl-odoo:
Clarify that it is required to create a redirection from the naked domain to the www. subdomain


xpl-odoo:
@robodoo r+

PR #8469, opened 2024-04-01
[ADD] extensions: icon_role font awesome

samueljlieber:
Task: [#3843168](https://www.odoo.com/web#id=3843168&cids=3&menu_id=4720&action=333&active_id=3835&model=project.task&view_type=form)

This PR adds a custom extension for an `:icon:` role to allow the use of Font Awesome icons in documentation.

With the new role, icons can be called by their classname. 

```rst
This is a Font Awesome :icon:`fa-step-forward` icon.
```

Screenshot 2024-04-02 at 11 07 01 AM

---

If this PR is approved, it will likely need to be manually forward ported to each version.


samueljlieber:
Hello @odoo/doc-review and @odoo/design-doc-review 👋 I have an exciting PR that implements icons into documentation via a new custom role. I decided to make this custom role its own extension, rather than add it directly to the odoo_theme, however I added the icon sets the role uses to the odoo_theme.

I tested icon RST formatting in all of the directives, and the :card: directive is the only one that the new :icon: role does not format in (title or body). I did not have any issues building with latex either, however, I did not try to use the :icon: role in the build.

The need to implement icons into Odoo documentation stems from writing effectively for the end-user. By using the UI icons from Odoo in the documentation, we can directly reference the visual elements of the product when instructing the user. This has many benefits:

- Increases the skim-ability of the document
- Provides clarity of information
- Reduces technical writer load

The US Content Marketing team has had to resort to using emojis and icon descriptors to instruct the user for UI elements. This has resulted in a cumbersome review process and multiple variations of emojis and descriptions used to depict an icon.

Im looking forward to hearing your thoughts on this implementation of icons in documentation! :)

AntoineVDV (commented):
Hello, good initiative! :)

Given that this role is specific to Odoo and its resources will live in the `odoo_theme` directory, a separate extension is not needed.

Regarding the rendering of the role, I think we should limit it to just the <icon>, instead of <icon> (<icon_class>). The reasons are that the writer might want to hide or rename the class, and I'm not convinced that the class would be translatable anyway.

samueljlieber:
Hi @AntoineVDV, thank you for your insightful review :) in de2dd98 I moved the role code to `odoo_theme` and removed any `icon_role` extension declarations I had in the `conf.py`. I also removed the `:guilabel:` icon descriptor feature from the `icon_role()` function, so now only the icon is rendered. Should be ready for another look 👍

Brieuc-brd:
Hello 👋 
Can you tell me where I can see an example of icon rendering in your branch?
Is there a specific page in the documentation?

samueljlieber:
> Hello 👋 Can you tell me where I can see an example of icon rendering in your branch? Is there a specific page in the documentation?

Hi @Brieuc-brd, in 20b0760 I added a section in `content/contributing/documentation/rst_cheat_sheet.rst` for icons so you can see the formatting– we will likely remove this content add before merge to keep this PR's changes strictly to the implementation of the :icon: role, but for now feel free to use this [list of all icon classes in RST](https://drive.google.com/file/d/1RYjEbg-1KBsPoxJKeZopGHUjo3_G0NwK/view?usp=sharing) that I generated for testing. 

See the icon section here: https://runbot145.odoo.com/runbot/static/build/60967664-15-0/logs/build/html/contributing/documentation/rst_cheat_sheet.html#icons

jcs-odoo:
THIS IS SUPER EXCITING!

Brieuc-brd (changes requested):
Hello @samueljlieber 👋 
Thanks for your work 👍 

I've done some tests on runbot and I'm concerned about a few things:

- This PR targets documentation for Odoo 15, while `oi` icons were introduced in `16.0`. This means that if the documentation uses these icons, users using Odoo before `16.0` won't find them in the backend.
IMO, `oi` should be added from `16.0`.
Here's an example below :

| In the Doc | In the backend |
| - | - |
| Capture d’écran 2024-04-08 à 16 34 20 | Capture d’écran 2024-04-08 à 16 32 14 |



- `oi` icons have been created to fit the backend's font base, which is `14px`. In the Documentation, the font base is `16px`, which creates antialiasing when you render `oi` icons. I think a CSS adjustment would be necessary :)

| Current | Expected | 
| - | - |
| Capture d’écran 2024-04-08 à 16 39 50 | Capture d’écran 2024-04-08 à 16 41 20 |

- What about classes to handle sizes and fixed widths ? (`fa-lg` =>`fa-5x`, `oi-small` => `oi-larger`, `fa-fw` & `oi-fw`) 
Is this something you don't need to add in the Documentation?

- Also, I don't see adaptations for RTL (https://github.com/odoo/odoo/blob/master/addons/web/static/lib/odoo_ui_icons/style.css#L63), is it something you don't need in the Documentation ?

- Odoo UI icons is a library that will evolve in the future (icons may be added or updated in future versions). Do you have any plans to keep the library up to date for Documentation? :)

I think adding FontAwesome and UI icons in the Documentation is a good idea, but we need to be sure that everything is well defined to maintain consistency with the Backend :)

Courage 💪 

cc. @stefanorigano

samueljlieber:
Thank you so much @Brieuc-brd for your thorough review and feedback on this PR- it is greatly appreciated 🙏 I want to share my thoughts on some of your points while I am working on implementing corrections for others.

> This PR targets documentation for Odoo 15, while oi icons were introduced in 16.0. This means that if the documentation uses these icons, users using Odoo before 16.0 won't find them in the backend.
IMO, oi should be added from 16.0.

- We certainly should only include the version specific icons. I agree oi should be added from 16.0, this PR can still target 15.0, however, I will be removing oi from its contents and adding that in, in the 16.0 version of the PR.

> oi icons have been created to fit the backend's font base, which is 14px. In the Documentation, the font base is 16px, which creates antialiasing when you render oi icons. I think a CSS adjustment would be necessary :)

- This is a great catch. I believe a simple add of the `font-size: 14px;` rule in `extensions/odoo_theme/static/scss/_font_awesome.scss` would do the trick. 

> What about classes to handle sizes and fixed widths ? (fa-lg =>fa-5x, oi-small => oi-larger, fa-fw & oi-fw)
Is this something you don't need to add in the Documentation?

- Personally, I think that this would add more complications than what is needed for documentation. At the moment I don't see an instance where we would need to "alter" icons in this way in docs, however, I am completely open to everyones thoughts on the matter. @jcs-odoo and @StraubCreative do you see any need to have different sizes and fixed widths options when using icons in docs? (also see the next point on rotation) ⬇️

> Also, I don't see adaptations for [RTL](https://github.com/odoo/odoo/blob/master/addons/web/static/lib/odoo_ui_icons/style.css#L63), is it something you don't need in the Documentation ?

- I also don't think we need adaptations for RTL as these adaptations seem to be already covered by existing icons. For example, `oi-arrow-down-right` with the `o_rtl` adaptation just becomes `oi-arrow-up-left` correct?

> Odoo UI icons is a library that will evolve in the future (icons may be added or updated in future versions). Do you have any plans to keep the library up to date for Documentation? :)

- Most definitely, I believe the implementation of this :icon: role lends itself to being flexible with the icon classes used. If/when Odoo UI is expanded, an improvement PR can be made on the documentation repo to include those classes. A good workflow would likely be periodically reviewing the library and making adjustments on doc quarterly. 

--- 

Looking forward to continuing the discussion here, so the implementation of icons in docs is accurate and consistent with Odoo and set up to be scalable as we progress! Thanks again!

samueljlieber:
Hi everyone, in https://github.com/odoo/documentation/commit/f1e54b6cf5366395fc5a7669a6a0b5892aa3f06f I did the following:

- Removed `oi` icons, leaving this PR to only introduce Font Awesome to 15.0. `oi` will be introduced from `16.0`+.
- Added the 14px font base to `extensions/odoo_theme/static/scss/_font_awesome.scss` to correct the antialiasing.
- Modified to support handling sizes, fixed widths, and colors (e.g. `fa-code-fork fa-flip-vertical text-success`)

Please see the [RST Cheatsheet Icons section](https://runbot215.odoo.com/runbot/static/build/61655123-15-0/logs/build/html/contributing/documentation/rst_cheat_sheet.html#icons) for formatting examples.

I am curious on your thoughts on including the color variations (`text-success`, `text-warning`, etc) in documentation. I think it is a good idea to exactly match the UI, and it is helpful when documenting the icon color changes that occur on the UI after an event.

Thank you!

xpl-odoo:
Thanks for the initiative, @samueljlieber ! It should really help both documentation writers and Odoo users alike.

AntoineVDV:
@Brieuc-brd Did you want to make another pass?

extensions/odoo_theme/static/scss/_font_awesome.scss
```diff
@@ -0,0 +1,947 @@
...
+ * -------------------------- */
+[class^="fa-"] {
+  display: inline-block;
+  font-family: FontAwesome !important;
+  speak: never;
+  font-style: normal;
+  font-weight: normal;
+  font-variant: normal;
+  text-transform: none;
+  line-height: 1;
```
Brieuc-brd, line 15: I assume this part of the code comes from the `.oi` selector of Oddo UI icons and that you've changed the font-family value? 🙂 
This part isn't necessary, as you've already defined the Font Awesome properties just below (lines 16 to 21).

extensions/odoo_theme/static/scss/_font_awesome.scss
```diff
@@ -0,0 +1,947 @@
...
+  font-display: block;
+}
+.fa {
+  display: inline-block;
+  font: normal normal normal 14px/1 FontAwesome;
+  font-size: inherit;
+  text-rendering: auto;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
```
Brieuc-brd, line 44: Since you don't use this class in favor of `[class^="fa-"]`, this is not necessary.

extensions/odoo_theme/static/scss/_font_awesome.scss
```diff
@@ -0,0 +1,947 @@
+/*!
+ *  Based on Font Awesome 4.7.0 by @davegandy - http://fontawesome.io - @fontawesome
+ *  License - http://fontawesome.io/license (Font: SIL OFL 1.1, CSS: MIT License)
+ */
+/* FONT PATH
+ * -------------------------- */
+[class^="fa-"] {
```
Brieuc-brd, line 7: Why not use the library as is, and therefore the `.fa` class?

Is it to simplify the use of `::icon::` and avoid having different "tags" when you import other libraries ? (e.g. avoid having something like `:icon_fa:`, `:icon_oi:`)

If so, I agree, but with this selector, it will only work if you define the icon first:

```xml
:icon:`fa-code-fork text-success` <!-- will work ✅  -->
:icon:`text-success fa-code-fork` <!-- won't work ❌ -->
```

To fix that, you should use :
```suggestion
[class*="fa-"] {
```

Brieuc-brd (changes requested):
Hey ! @samueljlieber 👋 
Thanks for your work 👍 

I've left a few comments below 😉

Brieuc-brd:
@samueljlieber 
I forgot something: `:icon:` should return a `<i/>` tag instead of `<span/>`.

extensions/odoo_theme/static/scss/_font_awesome.scss
```diff
@@ -0,0 +1,947 @@
...
+ */
+/* FONT PATH
+ * -------------------------- */
+@font-face {
+  font-family: 'FontAwesome';
+  src: url('../fonts/fontawesome-webfont.woff2?v=4.7.0') format('woff2'), url('../fonts/fontawesome-webfont.woff?v=4.7.0') format('woff');
+  font-weight: normal;
+  font-style: normal;
+  font-display: block;
+}
```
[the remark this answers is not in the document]
samueljlieber: Oops– this was left over from copying over from Odoo community. Will remove 👍

extensions/odoo_theme/static/scss/_font_awesome.scss
samueljlieber: Yes the intention is to simplify both the `:icon:` role and the classes added to it to render an icon. For Font Awesome specifically, I wanted to allow the use of just the icon class name without needed to prefix the `fa` class and the icon class (e.g. `fa-code-fork` instead of `fa fa-code-fork`).

Great suggestion to update the selector 👍

content/contributing/documentation/rst_cheat_sheet.rst
```diff
@@ -211,6 +211,48 @@ Use the `command` markup to highlight a command.
           Run the command :command:`make clean html` to delete existing built files and build the
           documentation to HTML.
 
+.. _contributing/icons:
+
+Icons
+-----
+
+Use the `icon` markup to add a class name of an icon. The icon set used is *Font Awesome*. It is
+recommended to accompany an icon with a label (`guilabel`), however, it is not mandatory.
```
jcs-odoo, line 220: maybe refer to "GUI element" with a ref link to `contributing/markups/guilabel` instead of "label (`guilabel`)"?

samueljlieber:
> I forgot something: `:icon:` should return a `<i/>` tag instead of `<span/>`.

@Brieuc-brd is using the `<i>` element a requirement of the icon libraries or purely semantic? I am asking because I believe this is a limitation of Docutils, the `<inline>` element used to generate the `:icon:` role renders as a `<span>`, and I am unaware of a way to specify using the `<i>` element 🤔 https://docutils.sourceforge.io/docs/ref/doctree.html#inline 

There is an inline `<emphasis>` element that returns an `<em>` tag rather than an `<i>`.

samueljlieber:
590ddca Implemented @Brieuc-brd 's [suggestions](https://github.com/odoo/documentation/pull/8469#pullrequestreview-2018985980), except for the <i> element– [see my reasoning](https://github.com/odoo/documentation/pull/8469#issuecomment-2075580121).

I also implemented @jcs-odoo 's [suggestion](https://github.com/odoo/documentation/pull/8469#discussion_r1578082734).

extensions/odoo_theme/static/scss/_font_awesome.scss
```diff
@@ -0,0 +1,923 @@
...
+  font-size: inherit;
+  text-rendering: auto;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+/*!
+ *  Based on Font Awesome 4.7.0 by @davegandy - http://fontawesome.io - @fontawesome
+ *  License - http://fontawesome.io/license (Font: SIL OFL 1.1, CSS: MIT License)
+ */
```
Brieuc-brd, line 19: I'm not sure this comment is necessary since you already defined it at the top of the file 😉 .

samueljlieber:
I forgot about [the `<raw>` element](https://docutils.sourceforge.io/docs/ref/doctree.html#raw)! 

In d6993b9 I implemented this to output an `<i>` instead of `<span>` for the `:icon:` role 👍

Brieuc-brd (approved):
LGTM, good job 👍

AntoineVDV (approved):
Thanks @Brieuc-brd for the design review :+1: 

@samueljlieber Good work!
I pushed some changes in a fixup! commit to:
- use `register_canonical_role` which should be more appropriate;
- remove the second call to `handle_error` because empty roles seems not to be interpreted;
- merge `handle_error` into `icon_role` since it's small enough and called in only one place;
- use the provided `text` variable rather than building `full_text` which, from what I can see, is essentially `test` + a space character at the end that is not interpreted when used in a `class` attribute;
- insert back a `inliner.problematic` node into the rendered document.
If that's ok for you, let's merge :rocket: 

@robodoo delegate+

jcs-odoo (changes requested):
Hi everyone :)

Here are some cases (in 15.0) where a class could be useful to be defined, as the color of the icon indicates a status:

- activities (in Project, for example)
  - `o_activity_color_planned`
  - `o_activity_color_planned`
  - `o_activity_color_overdue`
- messages status in the chatter (if the mail bounces back for example)
- the components' availability in manufacturing orders
  - No class on the FA icon, but the parent container has a green `btn-link`class
  - `text-danger`
- the favorite button, although it's not a class, but a different FA icon
  - one is specific to Odoo
  - the other is the FA icon but with a custom yellow color I think

However, the Odoo-related classes are not included here, and I wouldn't use the `text-info`, `text-success`, `text-danger`, and `text-warning` classes just to display a resembling color, since the only occurrence I could find was the use of `text-danger` in Manufacturing.

Unless we add "Odoo classes" to this repo, and define clearly when to use them or not in the writing guidelines, I think it's best to avoid mentioning that we can add classes in the RST cheat sheet.

In any case, I think we'll have to add some instructions in the writing guidelines when we revisit them :)

What do you think?

jcs-odoo:
Indeed, let's not use color classes for now.
Thanks for you work :)

samueljlieber:
3ebf78f removed `SUPPORTED_ICON_CLASSES` from the icon_role() and removed the FA color class examples from `rst_cheet_sheet.rst`, squashed and rebased. Ready for merge 👍

samueljlieber:
[9428a84](https://github.com/odoo/documentation/pull/8469/commits/9428a84da2148315a811254fcc89527cbe9edf33) Updated commit message:
`[ADD] extensions: icon_role odoo ui and font awesome` ➡️ `[ADD] extensions: icon_role font awesome`

samueljlieber:
Merging and FWP up to master. I will be opening a new PR to introduce Odoo UI icons in 16.0+.
Thanks everyone! 🚀 
@robodoo r+

jcs-odoo:
congrats ^^

PR #8483, opened 2024-04-02
[IMP] accounting: transactions update

toaa-odoo:
taskid-3835958


content/applications/finance/accounting/bank/transactions.rst
```diff
@@ -27,8 +27,13 @@ Odoo supports multiple file formats to import transactions:
 - Quicken Interchange Format (.QIF);
 - Belgium: Coded Statement of Account (.CODA).
 
-To import a file, go to the :guilabel:`Accounting Dashboard`, and in the :guilabel:`Bank` journal,
-click on :guilabel:`Import Transactions`.
+To import a file, go to the **Accounting Dashboard**, and in the :guilabel:`Bank` journal,
+click on :guilabel:`Import File`.
+
+.. tip::
+   You can also click :guilabel:`⋮` on the :guilabel:`Bank journal` when on the **Accounting
```
dade-odoo, line 34: Here, "Bank journal" is the guilabel, but above it's "Bank" journal, where only "bank" is the guilabel. Unless there's a specific reason for them to be different, I'd make them both the same.

dade-odoo (approved):
Just one small comment from me :) Otherwise looks good.

content/applications/finance/accounting/bank/transactions.rst
```diff
@@ -27,8 +27,13 @@ Odoo supports multiple file formats to import transactions:
...
 - Belgium: Coded Statement of Account (.CODA).
 
-To import a file, go to the :guilabel:`Accounting Dashboard`, and in the :guilabel:`Bank` journal,
-click on :guilabel:`Import Transactions`.
+To import a file, go to the **Accounting Dashboard**, and in the :guilabel:`Bank` journal,
+click on :guilabel:`Import File`.
+
+.. tip::
+   You can also click :guilabel:`⋮` on the :guilabel:`Bank` journal when on the **Accounting
+   Dashboard**, or click the gear icon :guilabel:`(⚙)` on the transaction view and then
```
auva-odoo, line 35: Sorry, it's not super clear to me what the "transaction view" is? Do you mean the list of transactions that are displayed when you click the 3 dots and then select transactions? If that's the case, I think I would rephrase to 1) avoid using the word "view" in this case (which is used to refer to actual views, i.e. list, kanban, form, etc.) and 2) maybe explain how to access the list of transactions since I haven't found anwhere where this is mentioned (but ofc I may be wrong!)

auva-odoo (changes requested):
@toaa-odoo Hi! I have one small comment, sorry about that!

auva-odoo (approved):
@robodoo r+

PR #8774, opened 2024-04-16
Updates the default From email address configuration instruction

kajedot:
I tried to force the e-mail address from which e-mails in Odoo are sent. 

I added `mail.default.from` and `mail.default.from_filter` parameters to my configuration as the current version of the documentation says. Unfortunately, it did not changed anything. 

Real change happened when I changed the "Default From Alias" field in  General Settings > Discuss > Alias Domain > Arrow near the text field. 
image

I think the current version of documentation is misleading. It says how to "force the email address from which emails are sent" but provides the wrong steps to do it. 

Probably this [instruction](https://odoo-documentation.oeerp.com/applications/general/email_communication/email_domain.html#use-a-default-email-address) is wrong as well. 

BR, Kajetan.

### DIFF
--- a/content/applications/general/email_communication/email_servers.rst
+++ b/content/applications/general/email_communication/email_servers.rst
@@ -131,6 +131,19 @@ If the `from_filter` contains a full email address, and if the `mail.default.fro
 this address, then all of the email addresses that are different from `mail.default.from` will be
 encapsulated in `mail.default.from`.
 
+If the following configuration won't work configure the following settings:
+
+1. Go to the :menuselection:`Settings --> Discuss --> Alias Domain` and click on the arrow near the text field.
+
+2. Type a domain name from which you want to send your E-Mails into the `Domain` field (like `example.com`).
+
+3. Into the `Default From Alias` field type an e-mail sender user name (like `donotreply`).
+
+Now all mails will be send from the e-mail: `donotreply@example.com`. 
+
+Remember to add the whole e-mail address the the `FROM Filtering` field in the outgoing email server's configuration. 
+
+
 .. _email_communication/from_filter:
 
 Utilizing the "From" filter on an outgoing email server


content/applications/general/email_communication/email_servers.rst
```diff
@@ -131,6 +131,19 @@ If the `from_filter` contains a full email address, and if the `mail.default.fro
...
+
+1. Go to the :menuselection:`Settings --> Discuss --> Alias Domain` and click on the arrow near the text field.
+
+2. Type a domain name from which you want to send your E-Mails into the `Domain` field (like `example.com`).
+
+3. Into the `Default From Alias` field type an e-mail sender user name (like `donotreply`).
+
+Now all mails will be send from the e-mail: `donotreply@example.com`. 
+
+Remember to add the whole e-mail address the the `FROM Filtering` field in the outgoing email server's configuration.
```
ksc-odoo, line 144:
```suggestion
If the configuration does not work, configure the following settings:

1. Go to the :menuselection:`Settings app --> Discuss section --> Alias Domain`, and click on the
arrow beside the text field.

2. Type a domain name from which emails should be sent into the :guilabel:`Alias Domain` text field
(e.g. `example.com`).

3. Into the `Default From Alias` field, type an email sender user name (like `donotreply`).

Now all mails are sent from the email address: `donotreply@example.com`. 

Remember to add the entire email address the the :guilabel:`FROM Filtering` field, located on the
:guilabel:`Outgoing Mail Servers` form, which is accessible via :menuselection:`Settings app -->
Discuss section`, and clicking :guilabel:`Outgoing Email Servers` beneath the :guilabel:`Use Custom
Email Servers` field.
```
I'm not seeing a 'Default From Alias' field on the Settings app page, nor on the Email Server page -- can you please specify where this particular field is located?

Also can you please specify where in this process the user should click 'Save' on the Settings page to activate the changes/settings. And can you please be a bit more specific about when they should navigate to the Outgoing Email Server configuration page to adjust the 'FROM Filtering' field?

Once you make those changes, please tag me again for another look. Thanks!

ksc-odoo (changes requested):
Hello, @kajedot -- thanks for bringing this to our attention. Once you implement the necessary adjustments, per the feedback that I left, please tag me again for another quick look, and then we'll be able to move this to the next stage. Thanks again! 👍

PR #8925, opened 2024-04-23
[FIX] Fix a typo

ChloeLiang:


Feyensv:
Indeed, this should be fixed.
I've created a PR to fix the typo from the earliest version (#9043).
Thanks !

PR #8927, opened 2024-04-23
[IMP] upgrade: modify support information

xpl-odoo:
task-3473194


xpl-odoo:
@odoo/manufacturing-doc-review I have replaced `:ref:` to the upgrade page with a `:doc:`, as `:ref:` should not be used to lead to the H1 of a page.

xpl-odoo:
@robodoo r+

PR #9065, opened 2024-04-29
[IMP] extensions: icon_role odoo ui

samueljlieber:
Task: [#3900566](https://www.odoo.com/web#id=3900566&cids=3&menu_id=4720&action=333&active_id=3835&model=project.task&view_type=form)

Extension of https://github.com/odoo/documentation/pull/8469. This PR adds support for Odoo UI icons with the icon_role().

### DIFF
--- a/content/contributing/documentation/rst_cheat_sheet.rst
+++ b/content/contributing/documentation/rst_cheat_sheet.rst
@@ -216,18 +216,21 @@ Use the `command` markup to highlight a command.
 Icons
 -----
 
-Use the `icon` markup to add a class name of an icon. The icon set used is *Font Awesome*. It is
-recommended to accompany an icon with a :ref:`contributing/markups/guilabel` as a descriptor,
-however, it is not mandatory.
+Use the `icon` markup to add a class name of an icon. There are two icon sets used in Odoo:
+`FontAwesome4 <https://fontawesome.com/v4/icons/>`_ and :doc:`Odoo UI
+</developer/reference/user_interface/icons>`. It is recommended to accompany an icon with a
+:ref:`contributing/markups/guilabel` as a descriptor, however, it is not mandatory.
 
 .. list-table::
    :class: o-showcase-table
 
-   * - The graph view is represented by the :icon:`fa-bar-chart` :guilabel:`(bar chart)` icon.
+   * - The graph view is represented by the :icon:`fa-area-chart` :guilabel:`(area chart)` icon. The
+       pivot view is represented by the :icon:`oi-view-pivot` icon.
 
    * - .. code-block:: text
 
-          The graph view is represented by the :icon:`fa-bar-chart` :guilabel:`(bar chart)` icon.
+          The graph view is represented by the :icon:`fa-area-chart` :guilabel:`(area chart)` icon.
+          The pivot view is represented by the :icon:`oi-view-pivot` icon.
 
 .. _contributing/lists:
 
--- a/extensions/odoo_theme/__init__.py
+++ b/extensions/odoo_theme/__init__.py
@@ -115,12 +115,12 @@ def resolve(old_resolve, tree, docname, *args, **kwargs):
 def icon_role(name, rawtext, text, lineno, inliner, options=None, content=None):
     """ Implement an `icon` role for Odoo and Font Awesome icons. """
     for icon_class in text.split():
-        if not icon_class.startswith('fa-'):
+        if not (icon_class.startswith('fa-') or icon_class.startswith('oi-')):
             report_error = inliner.reporter.error(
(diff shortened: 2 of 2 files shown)


content/contributing/documentation/rst_cheat_sheet.rst
```diff
@@ -216,18 +216,26 @@ Use the `command` markup to highlight a command.
 Icons
 -----
 
-Use the `icon` markup to add a class name of an icon. The icon set used is *Font Awesome*. It is
-recommended to accompany an icon with a :ref:`contributing/markups/guilabel` as a descriptor,
-however, it is not mandatory.
+Use the `icon` markup to add a class name of an icon. There are two icon sets used in the Odoo user
+interface: `FontAwesome4 <https://fontawesome.com/v4/icons/>`_ and :doc:`Odoo UI
+<../../developer/reference/user_interface/icons>`. It is recommended to accompany an icon with a
```
AntoineVDV, line 221: Nitpicking: if you have to go back all the way up to the root, you could as well start from the root:
```suggestion
</developer/reference/user_interface/icons>`. It is recommended to accompany an icon with a
```

AntoineVDV (approved):

samueljlieber:
55e78ad Thank you @AntoineVDV, updated developer doc path :)

content/contributing/documentation/rst_cheat_sheet.rst
```diff
@@ -216,18 +216,26 @@ Use the `command` markup to highlight a command.
...
 
    * - .. code-block:: text
 
           The graph view is represented by the :icon:`fa-bar-chart` :guilabel:`(bar chart)` icon.
+          The pivot view is represented by the :icon:`oi-view-pivot` icon.
+
+.. note::
+   Both *FontAwesome* and *Odoo UI* icons **do not** need the main class (`fa` or `oi`) to be added
+   in conjunction with the specific icon class when used in the :abbr:`RST (reStructuredText)`
+   documentation.
```
jcs-odoo, line 238: Is this note necessary? The example above is quite self-explanatory, IMO: the class name within an `:icon:`role, and that's it ^^

jcs-odoo (commented):
Hi @samueljlieber . Here are my comments. Again, thanks for your work!

Brieuc-brd:
Thanks for your work, I'll check asap 👍

extensions/odoo_theme/__init__.py
```diff
@@ -115,12 +115,12 @@ def _set_docname_as_class(_reference_node, _node_docname):
...
     for icon_class in text.split():
-        if not icon_class.startswith('fa-'):
+        if not (icon_class.startswith('fa-') or icon_class.startswith('oi-')):
             report_error = inliner.reporter.error(
                 f"'{icon_class}' is not a valid icon formatting class.", lineno=lineno
             )
             error_node = inliner.problematic(rawtext, rawtext, report_error)
             return [error_node], [report_error]
-    icon_html = f'<i class="{text}"></i>'
+    icon_html = f'<i class="oi {text}"></i>' if text.startswith('oi-') else f'<i class="{text}"></i>'
```
Brieuc-brd, line 124: Oh I see you're still using the `oi ` class prefix. I thought the point was to avoid using them (`fa` and `oi`) to ease the implementation (https://github.com/odoo/documentation/pull/8469#discussion_r1578065195).

It's better to use it tho, it's the right way to handle font icons.
But now it doesn't make sense not to do the same for Font Awesome and therefore modify the library as you've done [here](https://github.com/odoo/documentation/pull/8469) 😬.

The way you've done it here makes it possible to use the library correctly without complicating its implementation. Font Awesome should be implemented in the same way 🙂.

But I think this should be done in another PR targeting `15.0`. Could you handle it ?

Brieuc-brd (changes requested):
Hello @samueljlieber 👋 
Thanks again for your work 👍 .

Just a comment below.

By the way, do you plan to update the [UI icon list](https://www.odoo.com/documentation/17.0/developer/reference/user_interface/icons.html?highlight=icon) in `17.0`? A few new icons have been added : https://github.com/odoo/odoo/blob/17.0/addons/web/static/lib/odoo_ui_icons/style.css#L24-L61

extensions/odoo_theme/__init__.py
samueljlieber: Thats a good point 👍 I will make a 15.0 [FIX] to update the FA implementation to use the same methodology that OI uses for the main class prefix.

samueljlieber:
Hi @Brieuc-brd, thanks for the review! Yes I will create an `17.0` [IMP] PR to add these new OI icons 👍

samueljlieber:
@robodoo r+

PR #9185, opened 2024-05-06
[FIX] reference/view_architectures/*: remove `nosearch` meta directives

AntoineVDV:
The `nosearch` meta directive is not fully supported in included files: the generated file is excluded from search results, but the "nosearch" term is present in the included content.


AntoineVDV:
@robodoo r+

Issue #9041, opened 2024-04-27
Loyalty Cards since version 16 are applied on current order or future, not on future only


SBOSolutions:
https://github.com/odoo/documentation/blob/3179286430737032b6181d53f9867fba2d7729ac/content/applications/sales/sales/products_prices/loyalty_discount.rst?plain=1#L88

_change_ : 
Loyalty Cards: When making purchases, the customer accumulates points to exchange for rewards on future orders.

to : 
Loyalty Cards: When making purchases, the customer accumulates points to exchange for rewards on current and/or future orders.

V15 : 
![CleanShot 2024-04-27 at 11 58 23](https://github.com/odoo/documentation/assets/8417180/4052e7ab-ff1d-443e-a7ab-39e956d1d19b)

V16 and further : 
![CleanShot 2024-04-27 at 11 55 11](https://github.com/odoo/documentation/assets/8417180/c0a185c0-00bf-456f-965c-c1c6494c48d9)

xpl-odoo:
Hello @SBOSolutions
Thanks for the feedback!
My colleague @ksc-odoo will have a look at the issue.

ksc-odoo:
@SBOSolutions a fix/adjustment for this is currently in the review stages. https://github.com/odoo/documentation/pull/9225

PR #9225, opened 2024-05-06
[FIX] sales: fixed a loyalty card specificity

ksc-odoo:
implemented a necessary fix, as requested here: https://github.com/odoo/documentation/issues/9041

only line 88 was adjusted to adhere to the suggested fix (found in the link above)

This change **should be forward ported**

### DIFF
--- a/content/applications/sales/sales/products_prices/loyalty_discount.rst
+++ b/content/applications/sales/sales/products_prices/loyalty_discount.rst
@@ -86,7 +86,7 @@ The different :guilabel:`Program Types` available on the program form are:
 - :guilabel:`Next Order Coupons`: Generate and share single-use coupon codes that grant access to
   rewards on the customer's next order.
 - :guilabel:`Loyalty Cards`: When making purchases, the customer accumulates points to exchange for
-  rewards on future orders.
+  rewards on current and/or future orders.
 - :guilabel:`Promotions`: Set conditional rules for ordering products, which, when fulfilled, grant
   access to rewards for the customer.
 - :guilabel:`Discount Code`: Set codes which, when entered upon checkout, grant discounts to the


ksc-odoo:
hey @larm-odoo this super small adjustment is ready for a quick peer review from you whenever you get a chance (just one line - line 88/89) Thanks! 👍

content/applications/sales/sales/products_prices/loyalty_discount.rst
```diff
@@ -86,7 +86,7 @@ The different :guilabel:`Program Types` available on the program form are:
 - :guilabel:`Next Order Coupons`: Generate and share single-use coupon codes that grant access to
   rewards on the customer's next order.
 - :guilabel:`Loyalty Cards`: When making purchases, the customer accumulates points to exchange for
-  rewards on future orders.
+  rewards on current and/or future orders.
```
larm-odoo, line 89: I thought we were supposed to use just "and" instead of "and/or" according to the style guide?

larm-odoo (approved):
Hey @ksc-odoo - approving now! I just had one comment, lol - feels silly to leave one comment for one line change =D .

content/applications/sales/sales/products_prices/loyalty_discount.rst
ksc-odoo: Oh, really? I am not aware of that. Also, in this case, the statement requires the use of it. So, with that in mind, I am going to proceed with that wording -- but I'll be sure to keep it in mind moving forward. Thanks!

ksc-odoo:
Thanks for the quick review, @larm-odoo -- Looks like this one is ready for a Final Review, @StraubCreative 👍

StraubCreative (approved):
Looks good, thanks @ksc-odoo 
..
@robodoo r+

PR #9370, opened 2024-05-16
[IMP] website: multiple website

Donapi:
[taskid- 3716945](https://www.odoo.com/web#id=3716945&cids=1&model=project.task&view_type=form)


content/applications/websites/website/configuration/multi_website.rst
```diff
@@ -52,7 +55,7 @@ top right corner and select the website you want to switch to.
    :alt: Website selector
 
 .. note::
-   When you switch websites, you are redirected to the other website, to the same page (URL) as the
+   When you switch websites, you are :ref:`redirected <website/URL-redirection>` to the other website, to the same page (URL) as the
    current one. If the page you are currently viewing does not exist on the other website, you are
    redirected to a 404 error page. Once redirected, click :guilabel:`Create page` to create the
    page.
```
auva-odoo, line 61: This doesn't seem to be true anymore in 17.0. I may be wrong but it seems we're always redirected to the Homepage now: when I tested in 16.0, when I was on the /shop page and switched to another website, i landed on the /shop page of the other website. In 17.0, I land on the Home page.
Also, I wouldn't add a ref to the redirections part of the Pages doc page. In my understanding, this behavior doesn't quite fit the definition of a traditional redirection, as no redirection record is created and this doesn't concern website visitors. Instead, it's simply how the feature works when administrators switch between websites.
Of course, I could be mistaken, so it might be a good idea to verify this.

auva-odoo (changes requested):
Hi @Donapi I finished reviewing this PR as well, let me know if you have any questions!

content/applications/websites/website/configuration/multi_website/create-website.png
auva-odoo, line 1: I think you forgot to compress the image?

Donapi: no, it was already compressed :) 
image

content/applications/websites/website/configuration/multi_website.rst
Donapi: You're absolutely right. Updated accordingly.

auva-odoo (approved):
@Donapi LGTM, I just force-pushed one minor correction :slightly_smiling_face: 
@robodoo r+

PR #9435, opened 2024-05-21
Update products.rst

ray-odoo:
Corrected the part of the Customer Portal where digital products can be found after purchase.


StraubCreative (approved):
Looks good to me @ray-odoo :) 
Thanks for the fix!

StraubCreative:
Moved over to #9437 since tests aren't passing due to branch name issue.

PR #9440, opened 2024-05-22
[MOV] helpdesk: move getting started, reorg toctree

jero-odoo:
Task: ﻿https://www.odoo.com/mail/view?model=project.task&res_id=3939324&access_token=81e62b8d-9d4d-4bc9-91f6-173fb833755f

Moving Helpdesk "Getting Started" doc to Parent page, including images
Removed old getting started doc
Added redirects
Reorganized toctree


jero-odoo:
@samueljlieber hey sam, this is ready for tech review/merge (no content updates, just redirects/toctree edits)

samueljlieber (changes requested):
Hi @jero-odoo! Nice job on this restructure, I agree with these changes! I am requesting changes because I have one correction and then one suggestion that we can discuss:

- Please remove the `:nosearch:` meta tag on helpdesk.rst :)

Suggestion: While we are restructuring Helpdesk, what do you think about removing the "Overview" folder entirely? It doesn't seem to serve much purpose, and can actually hide all the sub-docs under another drop-down. Let me know your thoughts!

jero-odoo:
Hey @samueljlieber I remove the nosearch tag. I am not sure about removing the Overview folder at the moment, only because I think it makes sense to delineate between the basics and the more advanced settings. Also since most of the other apps seem to follow a similar setup, I think it is ok to leave it that way for now. If you think it is important, I don't have strong feelings either way. Let me know. Thanks!

samueljlieber (approved):
Thanks @jero-odoo, looks good to me! 👍
...
@robodoo r+

PR #9441, opened 2024-05-22
[IMP] conf: automatically generate the sitemap.xml for SEO

lse-odoo:
Before this commit:
Odoo's documentation does not have a sitemap.xml file. This file it usefull to better reference the documentation accross search engines in order for customers to find more relevant research.

After this commit:
The sitemap.xml is generated automatically for all documentation pages automatically with link to the translated versions following the guidelines, see:
https://developers.google.com/search/docs/specialty/international/localized-versions

The code is heavily inspired by:
https://github.com/jdillard/sphinx-sitemap
with some adaptations to fit odoo's needs

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -1,8 +1,11 @@
 import os
+import queue
 import re
 import shutil
 import sys
+from multiprocessing import Manager
 from pathlib import Path
+from xml.etree import ElementTree
 
 import docutils
 import sphinx
@@ -392,6 +395,15 @@ def setup(app):
     app.add_lexer('xml', XmlLexer)
 
     app.connect('html-page-context', _generate_alternate_urls)
+    # sitemap related:
+    if not app.config.language or app.config.language == 'en':
+        # As the URLs are the same for all languages except for the language code,
+        # we will only generate for the english (default) language and assume only the language code change
+        app.connect("builder-inited", _create_sitemap_queue)
+        app.connect("html-page-context", _add_html_url)
+        app.connect("build-finished", _create_sitemap)
+    else:
+        _logger.info("No sitemap generation as non-english language.")
 
     # Add a `condition` option on directives to ignore them based on config values
     app.add_config_value('odoo_dir_in_path', None, 'env')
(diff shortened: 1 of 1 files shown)


lse-odoo:
TODO: check how odoo.com handle documentation generation. 
- How many times is it ran ? (avoid creating one sitemap per language for instance).
- Do they `make clean` before? (otherwise only the new links will be in the sitemap, thus losing the previous ones)

Feyensv:
@AntoineVDV I don't have access to the script anymore I think, but it'd probably be difficult to handle a sitemap of all versions and languages we build the doc for no?

AntoineVDV:
@Feyensv @lse-odoo to check with internal (MAT, JDE or JKE) for the details, but it runs every night for every version and for every language in each version.

lse-odoo:
thanks for the info @AntoineVDV :) would you mind reviewing the PR please ? maybe @Feyensv you would be interested as well if you have some time

lse-odoo:
@odoo/doc-review would you have some time to review it please ? :)

lse-odoo:
@Feyensv would you have some time to review it please ? :)

Feyensv:
Started reviewing it, but other priorities came up :c.
I'm keeping it in my todo list but dunno exactly when we'll be able to focus on it, probably not before the 17.4 freeze

Feyensv (commented):
fixup commit to only build the sitemap when requested (is_remote_build), and to drop language-specific urls.
I'll go see with internal how they handle their sitemaps, and how we could integrate the documentation sitemaps.

lse-odoo:
@Feyensv ok! LGTM thanks for the review :)

Feyensv (changes requested):
After some investigation with JKE, we don't think that this sitemap would be useful right now, at least not in its current state.

A sitemap is mostly a way to inform search engines of new entry points they are not aware of, for a given domain name.  Considering (nearly) all the links are already referenced on `odoo.com/documentation`, itself being referenced on `odoo.com`, a sitemap wouldn't help much (and looking for stuff on google already points to odoo documentation, in the latest version most of the time).

The biggest improvement a sitemap could bring would be to inform the search engines of the pages last modification date (`lastmod`), so that modified pages are prioritized when crawling content, so that it's refreshed in priority.  But this feature is not included in the current version of this PR, nor guaranteed to work (i suppose we'd have to rely on the file last modification date but it doesn't seem like github propagates that information).

PR #9548, opened 2024-05-30
[IMP] elearning: editing content

Donapi:
taskid-3908034

### DIFF
--- a/content/applications/websites/elearning.rst
+++ b/content/applications/websites/elearning.rst
@@ -184,6 +184,11 @@ Manage your content by going to :menuselection:`eLearning --> Courses --> Conten
    :align: center
    :alt: Create your content.
 
+.. note::
+   Alternatively, you can also manage your content from the :guilabel:`All Courses` dashboard, by
+   clicking the course's name. Then, select the content you want to manage or click :guilabel:`Add
+   Content` to create new content.
+
 Document tab
 ------------
 


content/applications/websites/elearning.rst
```diff
@@ -184,6 +184,11 @@ Manage your content by going to :menuselection:`eLearning --> Courses --> Conten
    :align: center
    :alt: Create your content.
 
+.. note::
+   Alternatively, you can also manage your content from the :guilabel:`All Courses` dashboard, by
+   clicking the course's name. Then, select the content you want to manage or click :guilabel:`Add
+   Content` to create new content.
```
auva-odoo, line 190:
```suggestion
   Alternatively, you can manage your content from the :guilabel:`All Courses` dashboard: Click the course, then select the content you want to manage, or click :guilabel:`Add Content` to create new content.
```
- To try and avoid using the "you can do xxx *by doing* yyyy" wording and replace it with a more direct approach: to do xxx, do yyy (in this case, I left "you can do xxx" since it works better with "Alternatively, ...")
- From what I tested, you don't need to click the course's *name*, you can click anywhere on the card
- Other corrections are from grammarly (alternatively + also= no, apparently; comma before "or")

auva-odoo (changes requested):
Hi @Donapi, I added a small suggestion for you. I also think there might be a little mix-up with the task link. It left me a bit confused trying to figure out the purpose of the task. No big deal, I just wanted to get some more context. Thank you! :slightly_smiling_face:

PR #9661, opened 2024-06-05
[FIX] howtos/javascript_view: fix typo

Enmanuel-Otero-Montano:
A typo was found. The word "customeKanbanView" is misspelled and should be "customKanbanView", which is the constant exported correctly in the code. Therefore, the corrected line would be:


fdardenne (approved):
Thank you!

AntoineVDV (approved):
@robodoo r+

AntoineVDV:
@robodoo squash

PR #9703, opened 2024-06-07
[FIX] point_of_sale: replace receipt image

guga-odoo:
task-3834714


guga-odoo:
@LoredanaLrpz Hi! For the sake of gaining experience in github I've just replaced the receipt. Can you check if everything's alright to you? Thank you!

LoredanaLrpz (changes requested):
Hey @guga-odoo, the new screenshot looks great, but it appears larger. Could you make it about 75%? 
To do so, just add a line under `.. image::` with  `:scale: 75%` (like an `:alt:` for example) 
I'm not sure if 75% would be best, you'll have to try different percentages until you think it's good =)
Thank you! :)

LoredanaLrpz (approved):
Splendid

PR #9744, opened 2024-06-10
[FIX] tutorials/server_framework_101: position of form's notebook element 

cuongnmtm:
The notebook element should be nested in the sheet element.


AntoineVDV (approved):
Indeed! Thanks for fixing :)

@robodoo r+ squash

PR #9757, opened 2024-06-11
[FIX] voip: broken contact link

auva-odoo:
Changed https://www.axivox.com/contact into https://www.axivox.com/en/contact as it didn't exist (anymore).


auva-odoo:
@robodoo r+

PR #9795, opened 2024-06-13
[ADD] field_service: create new page about worksheet

guga-odoo:
Task-3991994


content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,72 @@
...
+:guilabel:`Task Title` and :guilabel:`Customer` fields are mandatory. Click :guilabel:`Save`.
+
+Assign a default Worksheet Template
+===================================
+
+If you have configured multiple Worksheet Templates, you can select one to be set up by
+default when you create a new Field Service task. Go to your task, drag your mouse over the name of
+your project to make the green arrow and the **internal link** tab appear, click on it.
+
+.. image:: worksheets/template-default.png
```
auva-odoo, line 52: I would crop this screenshot to only show the left column (which will highlight what you really want to show.

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Go to your Field Service task and select a :guilabel:`Worksheet Template`. Remember that the
+:guilabel:`Task Title` and :guilabel:`Customer` fields are mandatory. Click :guilabel:`Save`.
+
+Assign a default Worksheet Template
+===================================
+
+If you have configured multiple Worksheet Templates, you can select one to be set up by
+default when you create a new Field Service task. Go to your task, drag your mouse over the name of
+your project to make the green arrow and the **internal link** tab appear, click on it.
```
auva-odoo, line 50: I don't think the title is necessary here and it looks a little odd with just 1 line udner the previous one. Instead, I would maybe do something like this to 1) introduce the concept of a default template (which was done in the title) and 2) be a little more direct:

```suggestion
By default, the :guilabel:`Default worksheet` template is selected. To define another default worksheet template,  
click the :icon:`oi-arrow-right` (right-arrow) button that appears when you hover your mouse over the :guilabel:`Project` field on the task form. 

```
see this doc for how to add icons directly into the text like I did in my suggestion https://www.odoo.com/documentation/master/contributing/documentation/rst_cheat_sheet.html#icons

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. image:: worksheets/template-default.png
+   :scale: 75%
+   :alt: Mouse on project name to setup default worksheet template
+
+Then click
+on the :guilabel:`Settings` tab. Under **Field Service**, make sure the :guilabel:`Worksheet`
+checkbox is ticked and select the :guilabel:`Worksheet Template` you want to set up as default.
+:guilabel:`Save` your changes.
```
auva-odoo, line 59:
```suggestion
Then, in the :guilabel:`Settings` tab,  scroll down to the :guilabel:`Field service` section and select the :guilabel:`Worksheet Template` you want to set up as default.
```
- make sure the Worksheet checkbox is ticked: since we enabled it earlier, i don't think we need to do it again?
- Field service = guilabel and not bold since it's a UI element
- no need to really manually save so i remvoed that as well

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Use worksheets on site
+======================
+
+Field service workers can complete the worksheet to document the work they are doing on site. Go to
+your :guilabel:`Task` and click the :guilabel:`Worksheet` Smart Button. Fill the blank fields
+and :guilabel:`Save`.
```
auva-odoo, line 66:
```suggestion
To complete the worksheet on site, access the task and click the :guilabel:`Worksheet` smart button.
```
- I suggest removing the last part because what's in the worksheet can vary a lot and are not just "blanl fields", if that makes sense? (e.g., you can also have check boxes, or dropdown lists, or images, etc.) and i think it's kind of implied in the "complete the worksheet" part at the beginning?
- Saving is done automatically here (manual saving is optional so no need to mention it I think)

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. note::
+   - As soon as you save a worksheet, the appearance of the **Worksheet** smart button on
+     your task will change and display **Worksheet Complete** instead, regardless some fields are
+     left blank.
+   - Any field set up as **required** has to be filled for a worksheet to be saved.
```
auva-odoo, line 72:
```suggestion
   - As soon as you save a worksheet, the appearance of the :guilabel:`Worksheet` smart button on
     the task changes to :guilabel:`Worksheet Complete` instead, even if some fields are
     left blank.
   - Any field defined as :guilabel:`Required` has to be filled for a worksheet to be saved.
```
- these are all UI elements and should therefore be guilabel (+ Required with a capital R)
- regardless needs "of" afterwards so the sentence would be "regardless of some fields being left blank", which is very heavy IMO

auva-odoo (changes requested):
This is a super thorough and well-researched doc content-wise! I’ve added a few comments and suggestions to improve the formatting and structure a bit. I also recommend simplifying a few sentences to make them more direct. A few general comments (that are also included in the review):
- Use sentence case in titles
- If it's a UI element, it needs guilabel

I know it seems like a lot of comments, but please believe me when I say you did a really good job here! We can sit down together to discuss it if you’d like, or you can review my comments and get back to me with any questions. Please don't hesitate to reach out if you need anything! Thanks 🙂

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Because it is common for businesses to have their workers perform the same type of field service
+repeatedly, Odoo offers the option to create custom **Worksheet Templates**. By doing so, you
+don’t have to create the same worksheet every time you plan the same type of field service task.
+
+Configuration
+=============
+
+To use **worksheets** in **Field Service**, the **Worksheet Template** feature must be enabled
+first. Go to :menuselection:`Field Service --> Configuration --> Settings --> Worksheet Templates`
```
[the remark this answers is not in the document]
guga-odoo: Yes that's what I meant (question 2), your suggestion makes it more clear :)

content/applications/services/field_service/worksheets.rst
guga-odoo: Thank you for this tip!

content/applications/services/field_service/worksheets.rst
guga-odoo: Done, including png quant :)

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,68 @@
...
+page to leave **Studio**.
+
+.. seealso::
+   - :doc:`Fields and widgets in Studio <../../studio/fields>`
+
+Add a worksheet template to a Field Service task
+================================================
+
+Go to your field service task and select a :guilabel:`Worksheet Template` and click
+:guilabel:`Save`.
```
auva-odoo, line 46:
```suggestion
Go to your field service task, select a :guilabel:`Worksheet Template`, and click :guilabel:`Save`.
```

content/applications/services/field_service/worksheets.rst
auva-odoo: I think something went wrong somewhere because from what I can see the image is not compressed:
image
I also think we could get rid of the blank part

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,68 @@
...
+
+By default, the :guilabel:`Default worksheet` template is selected. To define another default
```
auva-odoo, line 48:
```suggestion
By default, the :guilabel:`Default Worksheet` template is selected. To define another default
```

content/applications/services/field_service/worksheets.rst
```diff
@@ -0,0 +1,68 @@
...
+worksheet template, click the :guilabel:`➔` (:guilabel:`Internal link`) icon that appears when you
```
auva-odoo, line 49: I am pretty sure it was an arrow before but now I see this instead (this means we would need to change the screenshot as well ...)
image

auva-odoo (changes requested):
HI @guga-odoo let me know when you have time to do this second round together :slightly_smiling_face:

auva-odoo (approved):
@robodoo r+

PR #9848, opened 2024-06-18
[IMP] eCommerce: update digital files section + remove screenshots al…

masp-odoo:
…ignment

task:3997287


masp-odoo:
@masp-odoo il faut changer la version pour 17.2 + 17.3 + Master (task à part pour Master). 
17.0 + 17.1 sont identiques.

auva-odoo (changes requested):
Dear @masp-odoo, here's my review after testing this more extensively. We should also definitely wait for feedback before we finalize this of course!
Also, I don't think we need to mention the MRP field in 17.2 + 17.3; it's not there unless Manufacturing is installed so shouldn't be mentioned here :slightly_smiling_face: 

Also, side note, but ideally, we try to keep the PR title under 50 characters if possible, could you please change it? Thank you!

auva-odoo:
Will be addressed in the ongoing restructure of the Products page; closing this for better management.

PR #9879, opened 2024-06-21
[I18N] *: export latest terms

ticodoo:


ticodoo:
Do I have the power?

@robodoo r+

ticodoo:
Do I have the power now?

@robodoo r+

PR #9945, opened 2024-06-25
[IMP] website: translations

Donapi:
[taskid-3617120](https://www.odoo.com/odoo/project.task/project.task/3617120?cids=1)
Changes in V17


content/applications/websites/website/configuration/translate.rst
```diff
@@ -2,101 +2,111 @@
...
-#. Click :guilabel:`Save`.
+.. tip::
+  Once the desired language is installed, you can translate some items from the backend (e.g., the
+  :guilabel:`product’s name` in the product template). To do so, click :guilabel:`EN` next to the
+  text you want to translate and add the translation.
+
+Block customization per language
+--------------------------------
+
+You can hide content (such as images or videos) that can not be translated. To do so,
```
auva-odoo, line 105:
```suggestion
You can hide content (such as images or videos, for example) depending on the language. To do so:
```
it's not just content that cannot be translated, for example, you could decide to hide an entire block just because the content doesn't apply for a language, or any other reason.

content/applications/websites/website/configuration/translate.rst
```diff
@@ -2,101 +2,111 @@
...
+To replace the original text with the translation, click the block, edit its contents, and save.
 
-#. Click :guilabel:`Save`.
+.. tip::
+  Once the desired language is installed, you can translate some items from the backend (e.g., the
+  :guilabel:`product’s name` in the product template). To do so, click :guilabel:`EN` next to the
+  text you want to translate and add the translation.
+
+Block customization per language
+--------------------------------
```
auva-odoo, line 103: - I would not call this section "Block customization" since you can't really customize the block based on the language, you can just hide/show it (like you explained directly below).
- Also, from 17.0, you could just refer to the "Conditional visibility" section in the Building blocks doc and then just keep the last bullet point in your list if you want ("Go to languages to configure ..." - you'll probably need to change it a little bit so it fits the text).

auva-odoo (changes requested):
Hi @Donapi thank you for updating this documentation, there were indeed some major things missing from it. I explained everything in my comment on line 16, but I force-pushed some changes to the section from "Add languages" until "Default language". Before you make any changes based on my review on your local branch, **don't forget to do a git pull** so your local branch will be updated with the changes I pushed. 
As usual, don't hesitate if there's anything you want to discuss or if you have any questions :slightly_smiling_face: Thank you!

auva-odoo (approved):
Thank you for your work @Donapi! I pushed some minor changes (remove capital letters, etc.), I hope you don't mind!

Just a quick note: from version 17.0, you can’t add languages from the language selector (see the tip on lines 26-29) so you might need to make some updates on the fw-port branch :slightly_smiling_face: 

@robodoo r+

PR #9820, opened 2024-06-14
[IMP] accounting/l10n: tell users to install l10n_fr_invoice_addr

sovi-odoo:
In Odoo 15.0, an extra module is necessary to conform to the french legislation. This will tell users to install it.

task-3856826

**Community PR:** odoo/odoo#162458

Forward up to 15.4

PR #10025, opened 2024-07-02
[IMP] accounting/l10n: tell users to install l10n_fr_invoice_addr

sovi-odoo:
In Odoo 15.0, an extra module is necessary to conform to the french legislation. This will tell users to install it.

This is a forward port of odoo/documentation#9820

task-3856826

Forward up to 17.3


dade-odoo (approved):
Thanks for the update :)

AntoineVDV:
@robodoo r+ up to saas-17.3

PR #10028, opened 2024-07-02
[IMP] loca: Kenya menu update

toaa-odoo:
Menus moved around since last PR. This updates it.

taskid-4028177


auva-odoo (approved):
Thank you @toaa-odoo ! I pushed a few changes; you can r+ if you're okay with them.
@robodoo delegate+

toaa-odoo:
@robodoo r+

PR #10065, opened 2024-07-03
[IMP] CRM: Inactive lost leads

pabr-odoo:
Task: https://www.odoo.com/odoo/my-tasks/4013046

Adding as small note admonition explaining how to mark an inactive lead as lost.

### DIFF
--- a/content/applications/sales/crm/pipeline/lost_opportunities.rst
+++ b/content/applications/sales/crm/pipeline/lost_opportunities.rst
@@ -45,6 +45,10 @@ corner of the opportunity.
    :align: center
    :alt: A lost opportunity with the lost banner added.
 
+.. note::
+   To mark an *inactive* (archived) opportunity as lost, set the :guilabel:`Probability` field to
+   `0` percent.
+
 .. _crm/lost-reasons:
 
 Create/edit lost reasons


pabr-odoo:
Hey @StraubCreative tagging you in for first review on this small addition when you get the chance. Thanks!

content/applications/sales/crm/pipeline/lost_opportunities.rst
```diff
@@ -45,6 +45,12 @@ corner of the opportunity.
    :align: center
    :alt: A lost opportunity with the lost banner added.
 
+.. note::
+   Only *active* opportunities can be marked as *Won* or *Lost*.
+
+   To mark an *inactive* opportunity as lost, first set the probability of the opportunity to 0%.
+   The :guilabel:`Lost` button can now be used to mark the opportunity as lost.
+
```
StraubCreative, line 53: >    Only *active* opportunities can be marked as *Won* or *Lost*.

Just confirmed in runbot— I don't think this is true, as I was able to mark an inactive lead as Lost.

>    The :guilabel:`Lost` button can now be used to mark the opportunity as lost.

The **Lost** button will only appear again if the opp is unarchived. Setting the prob % to 0 automatically marks it as lost.

```suggestion
   To mark an *inactive* (archived) opportunity as lost, set the :guilabel:`Probability` field to `0` percent.
```

StraubCreative (changes requested):
Hi @pabr-odoo 
Quick suggestion for you below, can you confirm please?
If correct, let's make any necessary changes and then we can pass to tech review, ty!

StraubCreative (approved):
LGTM! 
@samueljlieber you're up for quick tech review, thanks!

samueljlieber (approved):
Thanks for this quick improvement PR @pabr-odoo, looks good to me 👍
..
@robodoo r+

PR #10074, opened 2024-07-04
[FIX] german localization: Fixed broken GoBD link

dade-odoo:
If merged, this commit will replace the broken GoBD link with the correct link.

task ID: 4034487


auva-odoo (approved):
@robodoo r+

auva-odoo:
@robodoo retry

PR #10142, opened 2024-07-09
[IMP] Field_Service: add product to a task

guga-odoo:
taskid-4043214


content/applications/services/field_service/product_management.rst
```diff
@@ -1,10 +1,30 @@
+==================
+Product management
+==================
+
+When field service workers perform on-site, they commonly use products for the service. Odoo’s
+**Field Service** app allows them to record the products they’re using. Doing so keeps your
+inventory up-to-date in real-time and automatically adds these products to the invoice.
```
auva-odoo, line 7:
```suggestion
When field service workers perform tasks on-site, they commonly use products to complete their work. Odoo
Field Service allows them to record these products. Doing so keeps your
inventory up-to-date in real-time and automatically adds the products to the invoice.
```
- to avoid repetitions (use/they're using)
- we usually mention the app either on its own (Field Service) or you can also have Odoo + name of the app, without the possessive form (check out the POS or Accounting doc for reference). You can also have "app" somewhere in there (especially if you want to say "Open the xxx app") but in this case it's not necessary. Bold is also optional for app names (including "Odoo" in this case), but you can keep it if you want.
- Since this is the introduction to the doc page, I think it might be helpful to add a little something about the User default warehous, maybe? Maybe we could move part of the introduction that's under the User dfault warehouse heading here somewhere? I'm not sure, let's discuss this!

content/applications/services/field_service/product_management.rst
```diff
@@ -1,10 +1,30 @@
...
+
+Add products to a task
 ======================
```
auva-odoo, line 10: - This might be a question for @zulu-odoo since this was her suggestion, but she mentioned the term "product catalog" in the discord discussion. Is this was is commonly used to refer to this feature? Should we use/mention it somewhere in here?
- I think that you need to enable the Time and Material Invoicing setting to be able to use this feature so, depending on Zuza's answer to my question above, we might want to do something like this:

Product catalog
============

To enable the product catalog, go to :menuselection:`Field Service --> Configuration --> Settings` and enable the :guilabel:`Time and Material Invoicing` feature.

To add products to a task, proceed as follows:

(then keep your steps below)

auva-odoo (changes requested):
Hello there @guga-odoo thank you for giving so much love to the Field Service documentation. I added a few comments and suggestions in my review :slightly_smiling_face: Also:
- I think we could find a more accurate PR title for this (and remove the space after "Field", yes, that's clearly nitpicking, I'm sorry!) so that it mentions *what* you changed in the doc instead (e.g., Add products to field service tasks or Product catalog for field service tasks, etc.)
- Usually when list items include full sentences, we add a period at the end.
As usual, let me know if you have any questions or if there's anything you want to discuss :slightly_smiling_face: Thank you!

content/applications/services/field_service/product_management.rst
```diff
@@ -1,10 +1,30 @@
...
+
+#. Go to :guilabel:`Field Service`
+#. Select a task
+#. Click the :guilabel:`Products` Smart Button
+#. Click a product card to add it to your task. You can click anywhere on the card to add the
```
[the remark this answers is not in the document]
guga-odoo:
```suggestion
#. Click :guilabel:`Add` on a product card to add it to your task.
```

I agree that the user interface is a little weird. I wouldn't be surprised if they improved it one day by keeping the Add button and removing the option to click anywhere else. I've made a suggestion in that regard, what do you think?

auva-odoo:
@robodoo r+

auva-odoo (approved):

auva-odoo:
@robodoo retry

PR #10227, opened 2024-07-16
[MOV] ecommerce: moving product management page

Donapi:
[task](https://www.odoo.com/odoo/project/3835/tasks/4056328?cids=1)


content/applications/websites/ecommerce/products.rst
```diff
@@ -1,3 +1,9 @@
+:show-content:
+
+========
+Products
+========
+
 ==================
 Product management
 ==================
```
auva-odoo, line 9: Did you mean to remove this? There are currently 2 h1 in this rst ...

auva-odoo (changes requested):
Hi @Donapi! I noticed two small issues:
- There's a double h1 (see my comment in the review).
- It looks like the redirect from managing_products to products might have been missed.
I also suggest changing the PR name slightly to include the page you moved. Since you'll be moving a lot of pages while restructuring the commerce doc, this will help keep things organized better.
Thanks!

content/applications/websites/ecommerce/products.rst
ksc-odoo, line 9:
```suggestion
Product management
==================
```
I agree with @auva-odoo ... this is how this header should be formatted in this case

ksc-odoo (approved):
Apart from the one comment I left (which supported @auva-odoo 's original comment), this one looks ready to go. Approving now. 👍

redirects/17.0.txt
```diff
@@ -75,3 +75,7 @@ applications/sales/point_of_sale/pricing/digital_labels.rst applications/sales/p
 
 applications/services/helpdesk/overview/getting_started.rst applications/services/helpdesk.rst                                                            #helpdesk/overview/getting_started --> helpdesk
 applications/services/project/tasks/email_alias.rst applications/services/project/tasks/task_creation.rst                                                 #project/tasks/email_alias --> task_creation
+
+# applications/website
+
+applications/websites/ecommerce/managing_products/products.rst applications/websites/ecommerce/products.rst
```
auva-odoo, line 81: Sorry if I wasn't clear earlier: You need to add one for each of the sub-pages as well (catalog, variants, etc.)

auva-odoo (approved):
@robodoo r+

PR #10341, opened 2024-07-25
[IMP] website: landing page

masp-odoo:
taskID-3710109

### DIFF
--- a/content/applications/websites/website.rst
+++ b/content/applications/websites/website.rst
@@ -1,16 +1,49 @@
 :nosearch:
 :show-content:
 :hide-page-toc:
-:show-toc:
 
 =======
 Website
 =======
 
-Discover the best **Open-Source Website Builder** and learn how to build beautiful websites that
-convert visitors into leads or revenues.
+**Odoo Website** offers a user-friendly platform for creating and managing your website. It includes
+various tools and features to help you design, publish, and maintain web pages without needing
+advanced technical skills. You can easily customize layouts, add multimedia content, and integrate
+with other Odoo apps to expand your website's functionality.
 
-.. important::
+.. cards::
+
+   .. card:: Web design
+      :target: website/web_design
+      :large:
+
+      Design your website using building blocks and website themes.
+
+   .. card:: Pages
+      :target: website/pages
+      :large:
+
+      Manage website pages, menus, and search engine optimization.
+
+   .. card:: Configuration
+      :target: website/configuration
+      :large:
+
+      Configure domain names, address autocompletion, Google Search Console, cookies bar,
+      translations, multiple websites, form spam protection, content delivery network (CDN).
+
+   .. card:: Reporting
+      :target: website/reporting
+
+      Monitor your website's traffic with website analytics and set up link trackers.
+
+   .. card:: Mail groups
+      :target: website/mail_groups
+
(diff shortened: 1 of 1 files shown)


content/applications/websites/website.rst
```diff
@@ -1,14 +1,51 @@
...
 =======
 Website
 =======
 
-Discover the best **Open-Source Website Builder** and learn how to build beautiful websites that
-convert visitors into leads or revenues.
+Odoo provides a user-friendly website builder that empowers you to create and manage your online
+presence effectively. Enhance lead generation and conversions effortlessly while ensuring seamless
+maintenance. Odoo covers all your needs from web design to comprehensive reporting and
+:abbr:`SEO (Search Engine Optimization)`.
```
auva-odoo, line 13:
```suggestion
**Odoo Website** offers a user-friendly platform for creating and managing your website. It includes a variety of tools and features to help you design, publish, and maintain web pages without needing advanced technical skills. You can easily customize layouts, add multimedia content, and integrate with other Odoo apps to expand your website's functionality. 
```
You can of course tweak this, it was just easier for me to illustrate what we're aiming for using an example :slightly_smiling_face:

content/applications/websites/website.rst
```diff
@@ -1,14 +1,51 @@
...
+
+.. cards::
+
+   .. card:: Webdesign
```
auva-odoo, line 17:
```suggestion
   .. card:: Web design
```
Both options are correct, but since we've been using the two-word version, I suggest we stick with that.

content/applications/websites/website.rst
```diff
@@ -1,14 +1,51 @@
...
+      :target: website/reporting
+
+      Efficiently manage website analytics and integrate link trackers for comprehensive insights.
+
+   .. card:: Mail groups
+      :target: website/mail_groups
+
+      Easily configure mail groups for effective communication.
 
 .. important::
```
auva-odoo, line 50: I know you didn't change it but I don't think it needs to be an "important" admonition. It could be a note or a tip instead?

auva-odoo (changes requested):
Hi @masp-odoo! After discussing this with Xavier, we think it would be best if the main application page included a brief introduction, followed by either very simple cards like this https://www.odoo.com/documentation/master/applications/finance/accounting.html or actual content when possible (which isn’t the case here). The goal should be to inform, not to sell the product, as that's covered on the app-specific pages on odoo.com. So, no marketing language —maybe just a tiny bit in the introduction to highlight key features, but always in an objective way (e.g., "user-friendly" is fine). I didn't add suggestions for changing the cards but it shouldn't take long I think.

I've added a few comments in my review as well 😊

Sorry about this! I wasn't sure which direction to take after our last conversation, as it's been a long, long time since someone reviewed or created a main application page. I definitely think we should do it for all pages, so we don't end up with "empty" TOC tree items or ugly-looking pages. Thank you for starting that process :slightly_smiling_face:

masp-odoo:
Dear @auva-odoo, thanks for your feedback! I updated it in this direction, what do you think?

auva-odoo (changes requested):
@masp-odoo another little review as discussed :slightly_smiling_face: Let me know what you think!

auva-odoo (approved):
@robodoo r+

PR #10448, opened 2024-07-30
[IMP] Website: GTM integration

masp-odoo:
taskID-4006105

### DIFF
--- a/content/applications/websites/website/reporting/analytics.rst
+++ b/content/applications/websites/website/reporting/analytics.rst
@@ -137,7 +137,9 @@ measurement codes and related code fragments, collectively known as tags on your
 app, directly through the code injector.
 
 .. warning::
-   Google Tag Manager may not be compliant with local data protection regulations.
+   - Some GTM tags use data layers (e.g., advanced eCommerce tracking data layers) to retrieve
+     variables and send them to Google Analytics. Data layers are currently not managed in Odoo.
+   - Google Tag Manager may not be compliant with local data protection regulations.
 
 To use GTM, proceed as follows:
 


masp-odoo:
Hey @auva-odoo, here's an exponation from APL if it can help:
_"Odoo integrates well with GTM, you can use GTM and manage all the tags from GTM. What we're talking about here is adding code to odoo (in the form of a datalayer) to make the information available to GTM so that it can be sent to analytics. We don't do that. But we do allow GTM to be integrated in full standard via code injection. Here, when we talk about datalayers, we're really talking about adding, for example, the name, price, taxes, etc. for each product in a specific datalayer code so that GTM can send the information back to analytics. We don't do that at the moment.
Some GTM tags use a data layer to retrieve variables and send them to Google Analytics. This information is not currently managed in Odoo. Which in itself is quite logical. This could be done (check with the POs for this), but to use them, the customer would have to put GTM on their site and then configure their tag correctly. Now, if users want to track advanced ecommerce in Google Analytics, that's not possible at the moment. This represents a small proportion of our users.

With GTM, we can already do a lot of things outside of this tag, so the standard integration works."_

content/applications/websites/website/reporting/analytics.rst
```diff
@@ -137,7 +137,9 @@ measurement codes and related code fragments, collectively known as tags on your
 app, directly through the code injector.
 
 .. warning::
-   Google Tag Manager may not be compliant with local data protection regulations.
+   - GTM integration is possible in Odoo standard, although some data layers might not be installed
+     (e.g. advanced eCommerce tracking data layers).
```
auva-odoo, line 141: What if we said this instead (copied from your explanation); I think by providing a little bit more context and explanations, it makes it easier for people to understand?

```suggestion
   - Some GTM tags use data layers (e.g., advanced eCommerce tracking data layers) to retrieve variables and send them to Google Analytics. Data layers are currently not managed in Odoo. 
```
(I changed it a little bit, just trying to make it as complete as possible while staying as concise as possible; is it still accurate?)

auva-odoo (changes requested):
@masp-odoo Little suggestion for you, what do you think?

content/applications/websites/website/reporting/analytics.rst
masp-odoo: Hi, it seems accurate to me :) !

auva-odoo (approved):
@robodoo r+

PR #10567, opened 2024-08-02
[ADD] sales: point of sale qr code payment

tong-odoo:
This PR adds a new doc detailing how to setup point of sale qr code payment.


fja-odoo:
@tong-odoo This feature landed in 17.2 :)

tong-odoo:
Continue on #10597

PR #10599, opened 2024-08-05
[FIX] developer/reference/data: fix typo in data file example

AntoineVDV:
The attribute of `record` data operations should be `id` and not `name`.


AntoineVDV:
@robodoo r+

kmagusiak (approved):

PR #10614, opened 2024-08-06
Added information about known bug with authentication.

oskarenablebanking:


FlorianGilbert (changes requested):
Hello Oskar, thank you for your contribution!

But it's not really up-to-date :sweat_smile: 
For example the `addaccount.png` is different now :grimacing:.

Concerning the content, I have to admit that I'm not sure to understand the solution where the customer has to click "10-20 times" on the button.

Could you please elaborate? Now the "here" has been moved to the "Continue" button if the flow is now finished. 

If I'm not wrong, the bug that you're trying to avoid is not a bug but only caused by the popup blocker from the web browser / popup blocker / adblocker. 

Thank you!

PR #10562, opened 2024-08-01
[IMP] l10n_pe: Update current Peruvian modules and description

gmz-odoo:
1. **Remove** older ‘Introduction section’
2. **Add a table** listing and update of the currently supported features and modules with a description of its functionality (like the format/structure displayed for [Chile Localization](https://www.odoo.com/documentation/17.0/applications/finance/fiscal_localizations/chile.html) page) 
3. **Update** the description of suggested and related resources, such as Smart Tutorial videos
4. Update the reference for current Bank sync options
5. Update image for current listed Bank sync options
6. Correct the tittle of the exporting invoicing section

Expected result and comparison of actual info:
https://docs.google.com/document/d/1uyB_2sKIbab26LHO8mx9ikPOUukN6-DjHUz0jkcgXx0/edit

PR #10635, opened 2024-08-07
[IMP] l10n_pe : Export Invoice section

gmz-odoo:
- Correct the **tittle** of the _exporting invoicing section_
- Current: the section is using with the wrong tittle (and a duplicate keyword on the user doc )


samueljlieber:
Closing because the changes in this PR are being handled in https://github.com/odoo/documentation/pull/10562.

PR #10654, opened 2024-08-08
[IMP] ecommerce : zero priced products

masp-odoo:
task-3711225


masp-odoo:
I'll close this PR because @lman-odoo updated the whole Prices page. 
CC: @auva-odoo

PR #10729, opened 2024-08-13
[FIX] reference/view_architectures: remove outdated and broken link

AntoineVDV:
The external link to `AbstractController` class was broken, both because the placeholder substitution mechanism doesn't work in included RST excerpts and because the class has long been removed.

Since it makes little sense to replace the link with another link to the `useActionLinks` hook in the context of this reference doc, this commit removes the link altogether, as well as the erroneous mention of class limitations.


AntoineVDV:
@robodoo r+

PR #10735, opened 2024-08-13
[FIX] add missing language names to the language switcher

oomsveta:
Even though the Thai documentation has been built and is available, it didn't show up in the language switcher. This is because it needs to be given a "language name" first.

This commis add the language names corresponding to the newly added translations (namely, into Thai and Japanese).

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -252,11 +252,13 @@ languages_names = {
     'es': 'ES',
     'fr': 'FR',
     'it': 'IT',
+    'ja': 'JA',
     'ko': 'KO',
     'nl': 'NL',
     'pt_BR': 'PT',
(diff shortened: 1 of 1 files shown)


oomsveta:
> only in 17 and not earlier versions ? (cf #8194)

We don't build documentation for Japanese and Thai for earlier versions (cf https://github.com/odoo/saas-automation/blob/master/roles/doc/files/odoo/doc/doc.toml)

Feyensv (approved):
@robodoo delegate+

(don't have saas-automation rights anymore, but I'll trust you on this :D)

ticodoo:
@robodoo r+

PR #10757, opened 2024-08-14
[IMP] sales: parent doc restructure

ksc-odoo:
As discussed with ZST, this PR will focus mainly on restructuring the current format of the Sales parent doc to imbue the same look/feel as the recent marketing app restructures, where cards for every child doc is readily available at the top, ensuring for an enhance "ease of access" for the reader/user.

PROJECT TASK: https://www.odoo.com/odoo/project/3835/tasks/4115079


ksc-odoo:
Hey @samueljlieber  - since this is another one of those doc restructure PRs, I figured I'd pass it along to you for a quick Tech Review (whenever you get a chance). Thanks! 🙏

samueljlieber (changes requested):
Hi @ksc-odoo! 
Thank you for your work so far on this PR, I am requesting changes because I don't completely agree with this restructure in its current state. 

When using cards, they should resolve to a content page– and not another blank TOC page. Each of the cards added in this PR land on TOC pages, which takes **more** clicks than [the current structure](https://www.odoo.com/documentation/17.0/applications/sales/sales.html) to get to content.

sales-restructure-fs8

See in this image^ how using cards currently takes 3 clicks to get to content page, whereas the existing structure only takes 2 clicks.

Let me know if you want to brainstorm ways to make this restructure more effective. Thank you!

StraubCreative:
@ksc-odoo 

I agree with Sam's logic here and don't think now is the time to perform this action on the Sales app anyway since there's a lot of content for us to add still. The immediate impact is minimal and the medium/long-term impact is negligible since we'd have to redo it anyway with the addition of more content; I'd rather wait until there's a compelling need instead of restructuring for aesthetic (and not functional) reasons.

Open to discussing further, otherwise we can close this PR.

cc: @pabr-odoo @dikd-odoo

PR #10777, opened 2024-08-16
[FIX] change the display name of Korean to KR in the switcher

oomsveta:
According to feedback from our Korean translator,the abbreviation for Korean is usually KR instead of KO.

This commit updates the name of the language as displayed in the language switcher.


AntoineVDV (approved):
@robodoo r+

PR #10844, opened 2024-08-23
[FIX] inventory: DHL service

Felicious:
Specify that only the DHL Express service is supported for the DHL shipping connector

[Task](https://www.odoo.com/odoo/project/3835/tasks/4115174)

Fwport: all the way to master


StraubCreative (approved):
@robodoo r+

StraubCreative:
@robodoo retry

PR #10867, opened 2024-08-29
[IMP] upgrade: add a note about on-prem script requirements

cawo-odoo:
We often get questions from customers about errors they see running the upgrade-script, which are caused by restrictions of their environment. Add a note that mentions the most important requirements to maybe avoid some of these questions.

### DIFF
--- a/content/administration/upgrade.rst
+++ b/content/administration/upgrade.rst
@@ -156,6 +156,17 @@ project <https://odoo.sh/project>`_.
 
          $ python <(curl -s https://upgrade.odoo.com/upgrade) test -d <your db name> -t <target version>
 
+      .. note::
+         This command has some requirements on the environment it runs in:
+
+         - Some external commands. It will complain if one is missing.
+         - The system user that executes the command needs to be configured with access to the
+           database.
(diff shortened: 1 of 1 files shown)


cawo-odoo:
> The Odoo 14 documentation isn't maintained anymore, so you should target Odoo 15 instead :)

Yes, thanks, I know, but it is still online, though. And my thought was that a client who upgrades from 14 may think he'll better read that one. But if this argument is not good enough, I can retarget, sure.

xpl-odoo:
> > The Odoo 14 documentation isn't maintained anymore, so you should target Odoo 15 instead :)
> 
> Yes, thanks, I know, but it is still online, though. And my thought was that a client who upgrades from 14 may think he'll better read that one. But if this argument is not good enough, I can retarget, sure.

Some might get the impression that the Odoo 14 documentation is still being maintained if we start targeting it.
I guess this case could be an exception though, if you feel it's important.

cawo-odoo:
> Some might get the impression that the Odoo 14 documentation is still being maintained if we start targeting it. I guess this case could be an exception though, if you feel it's important.

Thanks, it's OK, I don't feel that strongly about it and already opened a PR for 15.0 :-)

PR #10904, opened 2024-09-02
[IMP] accounting: change Financial budget title

afma-odoo:
task-4147298


afma-odoo:
@auva-odoo I just changed the title (even though the whole doc needs to be updated) but there was a risk people would get confused with the new Financial Budget feature. Thanks!

auva-odoo (approved):
@robodoo r+

PR #10945, opened 2024-09-05
[IMP] payment_providers: PCI DSS compliance, payment methods on portal

auva-odoo:
task-4042449
task-2949379

### DIFF
--- a/content/applications/finance/payment_providers.rst
+++ b/content/applications/finance/payment_providers.rst
@@ -179,8 +179,8 @@ Bank payments
 
 .. _payment_providers/add_new:
 
-Enable a payment provider
-=========================
+Enabling a payment provider
+===========================
 
 To add a new payment provider and make its related payment methods available to your customers,
 proceed as follows:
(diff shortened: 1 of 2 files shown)


content/applications/finance/payment_providers.rst
```diff
@@ -38,15 +38,15 @@ subscriptions with recurring payments using their favorite payment methods, such
    card numbers) is stored on Odoo servers or Odoo databases hosted elsewhere. Instead, Odoo apps
    use a unique reference number for the data stored safely in the payment providers' systems.
 
-.. _payment_providers/supported_providers:
+.. _payment-providers-supported-providers:
```
AntoineVDV, line 41: Why? The `/` was there to scope the ref, but without it it doesn't make much sense to have a ref named `something-supported-something`.

content/applications/finance/payment_providers/mollie.rst
```diff
@@ -8,14 +8,14 @@ Configuration
...
+   - :ref:`payment-providers-add-new`
 
 Credentials tab
 ---------------
 
-Odoo needs your **API Credentials** to connect with your Mollie account, which comprise:
-
-- **API Key**: The test or live API Key depending on the configuration of the provider.
+Odoo needs your **API Credentials** to connect with your Mollie account, including your
+:guilabel:`**API Key`, i.e., the test or live API Key depending on the configuration of the
```
AntoineVDV, line 17: It sounds like the API is only one of the credentials that Odoo needs.

auva-odoo: I'm a bit confused, am I missing something? I can only see one field in the Credentials tab for Mollie. Having just bullet point felt a bit odd, so I reworded the sentence a little bit (I just came accross it while changing the refs).

content/applications/finance/payment_providers.rst
auva-odoo: It seems we can't use `/` in doc links in Odoo if we want to point to a custom anchor. Technically, it would be `AppName-FileName-something`, but I guess payment providers are a bit of a special case.

AntoineVDV: Ok, I see. The recommendation must be because HTML anchors are computed based on the heading's title whereas RST anchors are identical to whatever you typed in the sources.

I disagree though, because:
- Most people will rely on HTML anchors when creating documentation links in Odoo with the `documentation=` attribute, thus completely ignoring the carefully crafted RST anchor.
- RST anchors are very often updated when the corresponding heading titles are updated, thus canceling the only pro argument for RST anchors.
- Clearing scoping RST anchors is important because they are sometimes referenced without context from other documentation files and because they might be conflicting anchor names across the documentation. I think the usage of `/` in RST anchors largely outweighs the benefits (if any) of dashed RST anchors.

content/applications/finance/payment_providers/mollie.rst
AntoineVDV: No, I mean that the instructions make it sound like there exist other API credentials fields: "API credentials [...] including your API key"

content/applications/finance/payment_providers.rst
jcs-odoo: I'm not 100% sure anymore. I don't think I had a problem with the `/` but rather with letters written in capital letters in the code.
@AntoineVDV @xpl-odoo @auva-odoo

content/applications/finance/payment_providers/mollie.rst
auva-odoo: (I will take care of this in another PR as the whole doc page needs work)

auva-odoo:
@robodoo r+

PR #10950, opened 2024-09-06
[IMP] odoo_sh : add a warning to remove saas modules after upload

loleloup:
Some people forget to remove the saas-modules from their database after upload, making them open tickets to support because keeping them installed causes some difficult to debug tracebacks


xpl-odoo:
@qsaodoo and @frps1 are revamping the whole Odoo.sh documentation.
Unless it's urgent to have that online, I would just check with them if they can include that info :)

loleloup:
Didn't know about that, I'll go talk with them thanks for the info :)

PR #11003, opened 2024-09-13
[IMP] POS: add information about POS categories

LoredanaLrpz:
task-3984698

Still have to come back in 16 (restrict and start categories) and saas-17.4 (point of sale tab in product form)


LoredanaLrpz:
@robodoo r+

LoredanaLrpz:
@robodoo r+

PR #11041, opened 2024-09-18
[IMP] accounting: split deferred settings

RicGR98:
This has been done in the following commit
https://github.com/odoo/enterprise/commit/39fffc5541eb51d7a3ae25405abe90fe6031001a

### DIFF
--- a/content/applications/finance/accounting/customer_invoices/deferred_revenues.rst
+++ b/content/applications/finance/accounting/customer_invoices/deferred_revenues.rst
@@ -33,18 +33,16 @@ Make sure the default settings are correctly configured for your business. To do
 
 Journal
   The deferral entries are posted in this journal.
-Deferred Expense Account
-  Expenses are deferred on this Current Asset account until they are recognized.
-Deferred Revenue Account
+Deferred Revenue
   Revenues are deferred on this Current Liability account until they are recognized.
 Generate Entries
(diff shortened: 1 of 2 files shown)


dade-odoo:
Hey @RicGR98 , as always thanks for keeping our accounting documentation up to date! I have two questions:
1. In the Configuration sections of each doc, it looks like you removed the wrong lines for each - the deferred revenue doc includes the line explaining deferred expenses and vice versa. Was that a mistake or am I misunderstanding something?
2. I see the value of `Full Months` is available for the `Based on` field. Could you add a brief explanation of how that would work (what makes it different from just `Months`)?
Thanks!

RicGR98:
Hello @dade-odoo 

Sorry for the delayed answer with the freeze, OXP, ...

1. Indeed you're right, I've fixed that :)
2. The `full_months` mode has been added in 17.0 (https://github.com/odoo/enterprise/pull/62470) whereas the split of the settings only on 17.3. So I was planning on doing the explanation of the `full_months` in another PR targeting 17.0 instead (I'll let you know when it's done of course).

dade-odoo (approved):
@RicGR98 no problem, I was also really busy with OXP so I'm just getting back to documentation now. This looks good to me, thanks for the addition and the explanation of why full months wasn't included yet :)

RicGR98:
@robodoo r+

PR #11046, opened 2024-09-18
Removed "See Also Odoo Tutorial"

adam-rocska:
The linked page yields an error message:
"You do not have permission to access this course." 

The error message is yielded as:
- a guest
- an average joe registered user

My fix proposal is to just remove the link from the docs altogether. Alternatively Odoo could make the linked tutorial publicly accessible if it's part of the public docs.


adam-rocska:
Actually as I click around, it's not only an Odoo 16 issue, but is present witth Odoo 17 and Odoo Online.

Maybe a pythoner of this repo knows how to attack this problem better?

PR #11116, opened 2024-09-27
[IMP] developer/testing: update tours documentation

pipu-odoo:
The towers interface was modified during version 17. Therefore, in this commit, we update the documentation of this interface.


Feyensv (approved):
@robodoo r+ squash

(why didn't you retarget the other PR? :thinking:)

PR #11150, opened 2024-10-07
[FIX] administration,developer: Odoo 18 now require PG15

KangOl:
See odoo/odoo#182645


AntoineVDV (approved):
@robodoo r+ delegate+

PR #11231, opened 2024-10-18
Update setup_guide.rst

rayenchtioui:
While learning Odoo and exploring the documentation, I noticed that users without a Python background might be unfamiliar with the concept of virtual environments. This can lead to confusion and frustration when trying to run the server without properly setting up the environment. I've added a section explaining how to create a virtual environment and install dependencies, which will save time for new users and make the setup process smoother.


Younes-ch:
Good idea!

I think you should mention on how to activate the virtual env before installing the dependencies (Both on windows/Linux):

Windows: `<myenvname>\Scripts\activate`
Linux/MacOS: `source <myenvname>/bin/activate`

PR #11263, opened 2024-10-22
[FIX] l10n_de: Fix second broken GoBD link

dade-odoo:
There were two broken links originally, one of which was fixed by PR#10074. This commit will remove the first link entirely and fix the second link, as suggested by TSB.

task-4088554


xpl-odoo (changes requested):
You will need to fix the `ci/documentation` error first:
image
If the error isn't clear: it means the same hypertext (or explicit target name) doesn't point to the same hyperlink. So you need to change one of the two hypertext so they match :)

@robodoo delegate+

dade-odoo:
@xpl-odoo oops, my mistake. Should be fixed now.

xpl-odoo:
@robodoo r+

PR #11283, opened 2024-10-25
[ADD] purchase: EDI with SPS commerce

Felicious:
Document types to facilitate EDI integration in Odoo in version 16. 

Available data exchange protocols that are available in Odoo 16 are:
- 850 inbound
- 810 outbound


Felicious:
Followed up with Fernando, and we're planning to freeze this document until the feature is more developed!

PR #11292, opened 2024-10-28
[IMP] core: controllers json -> jsonrpc

Julien00859:
See https://github.com/odoo/odoo/pull/183636

Task-4257153


AntoineVDV (approved):
@robodoo r+

PR #11334, opened 2024-11-06
[IMP] developer: added "--shell-file" CLI option

lordkrandel:
Added the description for the new `--shell-file` option, which overrides the the `$PYTHONSTARTUP` env variable to initialize the shell session with a startup Python script.

Related docs: https://docs.python.org/3/using/cmdline.html#envvar-PYTHONSTARTUP

Community PR: odoo/odoo#185075
Task [link](https://www.odoo.com/odoo/project/967/tasks/4306704)
task-4306704

### DIFF
--- a/content/developer/reference/cli.rst
+++ b/content/developer/reference/cli.rst
@@ -695,10 +695,16 @@ interaction with the :ref:`orm <reference/orm>` and its functionalities.
       By default, the shell is running in transaction mode. This means that any change made to the
       database is rolled back when exiting the shell. To commit changes, use `env.cr.commit()`.
 
+
+.. option:: --shell-file <init_script.py>
+
+   Specify a Python script to be run after the start of the shell. Overrides the environment
+   variable `PYTHONSTARTUP`.
+
 .. option:: --shell-interface (ipython|ptpython|bpython|python)
 
-   Specify a preferred REPL to use in shell mode. This shell is started with the `env` variable
-   already initialized to be able to access the ORM and other Odoo modules.
+   Specify a preferred `REPL` to use in shell mode. This shell is started with the `env` variable
+   already initialized to be able to access the `ORM` and other Odoo modules.
 
 .. seealso::
    :ref:`reference/orm/environment`


content/developer/reference/cli.rst
```diff
@@ -695,6 +695,11 @@ interaction with the :ref:`orm <reference/orm>` and its functionalities.
       By default, the shell is running in transaction mode. This means that any change made to the
       database is rolled back when exiting the shell. To commit changes, use `env.cr.commit()`.
 
+
+.. option:: --shell-file <init_script.py>
+   Specify a python script to be run after the start of the shell.
+   Overrides the environment variable PYTHONSTARTUP.
```
AntoineVDV, line 701: Single line breaks are not taken into account in RST. You need to add a blank line after `option`, otherwise it will look like [that](https://runbot117.odoo.com/runbot/static/build/70270290-master/logs/build/html/developer/reference/cli.html#cmdoption-odoo-bin-shell-file).

```suggestion
.. option:: --shell-file <init_script.py>

   Specify a python script to be run after the start of the shell.
   Overrides the environment variable PYTHONSTARTUP.
```

AntoineVDV: For the same reason, you might want to *not* return to a new line at the end of your sentence if you don't intend to have a line return in the rendered HTML:
```suggestion
.. option:: --shell-file <init_script.py>
   Specify a python script to be run after the start of the shell. Overrides the environment variable PYTHONSTARTUP.
```

AntoineVDV, line 701: Highlight technical terms:
```suggestion
.. option:: --shell-file <init_script.py>
   Specify a python script to be run after the start of the shell.
   Overrides the environment variable `PYTHONSTARTUP`.
```

AntoineVDV (changes requested):
The commit message should be adapted to the repository in which it will be merged. "base" is not a documentation directory, and the commit is not about adding a shell feature but about documenting it.

lordkrandel:
@robodoo r+

PR #11415, opened 2024-11-19
[IMP] documentation: update many2many_binary widget

amdi-odoo:
The widget now also display an image preview for the files of MIME type "image".
A new option 'reverse_order' has been added to visually reverse the files order.

### DIFF
--- a/content/developer/reference/frontend/javascript_reference.rst
+++ b/content/developer/reference/frontend/javascript_reference.rst
@@ -1202,6 +1202,8 @@ Many2many (`many2many`)
 Many2many Binary File (`many2many_binary`)
     This widget helps the user to upload or delete one or more files at the same time.
 
+    An image preview is available for files with the MIME type "image".
+
     Note that this widget is specific to the model `ir.attachment`.
 
     - Supported field types: `many2many`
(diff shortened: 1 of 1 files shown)


content/developer/reference/frontend/javascript_reference.rst
```diff
@@ -1212,6 +1214,8 @@ Many2many Binary File (`many2many_binary`)
 
         (cf: ``accept`` attribute on `<input type="file" />`)
 
+    - `reverse_order`: set to `true` to add new files at the top of the list (default: `false`, new files are added at the bottom).
```
aab-odoo, line 1217: This is not what you implemented, and this is not what the specs say.

In your implementation, files are received in a given order, and you display them the order way around. It's thus indeed also the same for added files: as the list is reversed, files added in the end of the list are displayed first.

The specs say that we want the same order as in the post. By reversing the files, new files are indeed added to the top, but when you open an existing record, files are ordered from less recent to more recent, which isn't the order you want. I don't know the business logic, so I don't know how to enforce that, but I don't think such an option is the way to go (and your comment on the specs makes me think that you doubt as well). I just confirmed with framework py that for now, there's no way to enforce an order on a many2many relation (it's the order of the comodel that's used, which in this case is `id desc`).

I showed your usecase to cth, he'll take a look to determine whether there's something we could/should do in the many2many_binary field (all the time, not through an option).

amdi-odoo: @aab-odoo 
Ok got it :smile: 
I'll try to find another way to fix the functional issue without messing with the widget or I'll simply wait for a potential widget improvement.
Thanks for looking into it! :slightly_smiling_face:

aab-odoo:
@amdi-odoo can we close this PR?

PR #11462, opened 2024-11-27
[17.0][IMP] Update LDAP documentation: disable referral chasing in case of Microsoft AD

vehi-invitu:
Description of the issue :
We needed to use Microsoft Active Directory (AD) for authentication in Odoo, but the existing documentation didn't address this case. This led to login issues for users even with valid credentials, so i aimed to update the documentation to provide clear instructions on how to disable referral chasing, which is crucial for successful authentication.

cc @cvinh

### DIFF
--- a/content/applications/general/users/ldap.rst
+++ b/content/applications/general/users/ldap.rst
@@ -53,3 +53,18 @@ Sign in with LDAP
 .. image:: ldap/ldap07.png
     :align: center
     :alt: User information highlighted on LDAP server setup on Odoo.
+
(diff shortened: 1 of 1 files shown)


auva-odoo:
@vehi-invitu thank you for your work! I’ve addressed it in [PR #11555](https://github.com/odoo/documentation/pull/11555/files) and took the opportunity to reformat the page to align with our [documentation guidelines (https://www.odoo.com/documentation/master/contributing/documentation.html)].

PR #11484, opened 2024-11-29
[FIX] developer/tutorials/importable_modules: no self in actions

vava-odoo:
In ir.actions.server, the variable containing the record(s) is not `self`, but `record` and `records`.
Also, since there is a word about `store` attribute for computed fields, this commit specifies the behaviour of `readonly` attribute.


Feyensv (approved):
@robodoo delegate+

bouvyd:
@vava-odoo 👍

vava-odoo:
@robodoo r+

PR #11520, opened 2024-12-03
[FIX] Upgrade: Correct URL and text for post-upgrade issues

nmarotte:
As stated on the sentence above, in case of an issue with the **production** database, we should open a ticket with the label (production) and not testing an upgrade.

This commits adapt the test to match the label currently present on odoo.com/help and the URL to pre-select the correct stage as well


xpl-odoo (approved):
@robodoo delegate+
@nmarotte you can r+ if ready to go ;)

nmarotte:
@robodoo r+

PR #11582, opened 2024-12-13
[IMP] point_of_sale: remove start category feature from pos

jipr-odoo:
In this commit:
==============
The "Start Category" feature has been removed from the Point of Sale. This feature is now redundant due to the existing "Restrict Category" feature, which provides similar functionality by limiting the available categories in POS

Task- 4405337


LoredanaLrpz:
Hey @jipr-odoo , thanks for this, I'm thinking about maybe changing the page a bit more, do you mind if I push on your branch to do so?
Thanks again =)

jipr-odoo:
> Hey @jipr-odoo , thanks for this, I'm thinking about maybe changing the page a bit more, do you mind if I push on your branch to do so? Thanks again =)

Hello @LoredanaLrpz 
Sure, you can make a second commit if it doesn’t take much time.

We already removed this feature in version 18.0, but it was left in the documentation PR. As a result, we recently received a query from customer support regarding customer confusion about this feature. To avoid any further confusion in the future, we want to remove it from the documentation as soon as possible.  

Thanks!

LoredanaLrpz:
@robodoo r+

PR #11620, opened 2024-12-20
[FIX] Egypt fiscal localizations: Update proxy install instructions

iCart:
Since Odoo 16, to install the local proxy, users need to select the "Odoo IoT" option install of the "Local Proxy Mode" to install the local proxy.

This commit updates both the text and accompanying screenshot

See also ticket 4347475


auva-odoo (approved):
Hello @iCart thank you for your work!

@robodoo r+

PR #11644, opened 2024-12-26
[IMP] website_sale: Google Merchant Center data source

lipiraux:
To ensure the multi-channel promotion of products through various platforms (e.g., Google, Meta, Pinterest, TikTok), we need to provide data feeds so that our users can easily input their catalog into these platforms.

This commit adds the documentation on the new /gmc.xml file for Google Merchant Center. This XML file serves as a product feed to Google, ensuring synchronization between our platform and Google's services.

The product feed can be configured to utilize a pricelist and can be translated to target different locations. For instance, if the French language is installed and there exists a pricelist named “UE” with the currency EUR, then “/fr/gmc-UE.xml” represents a feed in French, where all prices are displayed in EUR and computed according to the UE pricelist.

See also:  odoo/odoo#186976

task-4319930


lipiraux:
closed in favor of odoo/documentation#12261

PR #11674, opened 2025-01-01
Update 02_build_a_dashboard.rst

khad-odoo:
The useService("rpc") function is deprecated in odoo 18.

Now, the default way to call a controller which is not associated with a model is using the rpc function located in the web/static/src/core/network/rpc.js.

Therefore updated the "4. Call the server, add some statistics"  section to reflect the correct references in the odoo 18 tutorials/discover_js_framework docs.


ged-odoo:
nice find

ged-odoo (approved):

khad-odoo:
Thanks @ged-odoo , 
now if there are no changes needed, can you please merge this or tag the concerned person into this discussion?

PR #11676, opened 2025-01-02
[FIX] accounting/tax-return: fix tax return menu path

dade-odoo:
Fix the menu navigation path of the tax return. To fw-port through 17.4. Small changes to be made from 18.0

task-4424499


auva-odoo (approved):
@robodoo r+

PR #11721, opened 2025-01-10
[IMP] javascript reference: document domain field options

aab-odoo:
Before this commit, options of the domain field weren't documented. Now, they are.


AntoineVDV (approved):
Good to know :+1: 

@robodoo delegate+

By the way, the linked PR should be updated or removed.

aab-odoo:
robodoo r+

PR #11747, opened 2025-01-13
[IMP] account_reports: Unambiguous cross_report aggregations

lost-odoo:
Before this commit, when using cross_report subformula it would search for a report that has an line with this code. We had to ensure there was no duplicate codes and so it was difficult to maintain.

Now, when we use a cross_report subformula, we need to explicitly target the report from where the code is from.

task-4457667


afma-odoo (approved):
Thanks @lost-odoo for the update. Can you just capitalize "ID"?
Thank you!

auva-odoo (approved):
Thanks @lost-odoo and @afma-odoo for your work!

@robodoo delegate+

hupo-odoo:
@robodoo r-

lost-odoo:
@robodoo r+

PR #11823, opened 2025-01-22
[IMP] applications/sales: prepare files for future Gelato documentation

anko-odoo:
The content of the `gelato.rst` file will be added later on. The file itself is already added so that it can be referenced from Odoo through a documentation link.

task-3935688

See also:
- https://github.com/odoo/odoo/pull/193457
- https://github.com/odoo/enterprise/pull/77839

### DIFF
--- a/content/applications/sales/sales/gelato.rst
+++ b/content/applications/sales/sales/gelato.rst
@@ -1,0 +1,7 @@
+:orphan:
+
+======
+Gelato
+======
+
+The documentation for Gelato will be added soon.


StraubCreative:
Hi @anko-odoo and @AntoineVDV 

For the new content like Gelato, we would typically roll out with an overview doc first which would include basic details of what the feature is, how to configure it, and setting definitions. 

After that is released we would drop more flow-specific content underneath that in later PRs. 

A good example of that is the [manufacturing subcontracting section](https://www.odoo.com/documentation/18.0/applications/inventory_and_mrp/manufacturing/subcontracting.html). 

Is this amenable to you?

cc: @Felicious @dikd-odoo

AntoineVDV:
Hey @StraubCreative, sure, we can proceed like that :)

In the meantime, we'll still merge the current PR as-is because we need the doc URL to exist ASAP. But it's fine because of the `orphan` meta tag that makes the page inaccessible from the TOC.

AntoineVDV (approved):

AntoineVDV:
@robodoo r+

anko-odoo:
@robodoo r+

anko-odoo:
@robodoo r+

anko-odoo:
@robodoo rebase-ff

PR #11825, opened 2025-01-22
[FIX] mobile: SS0 -> SSO typo

tanyabouman:


xpl-odoo (approved):
Hi @tanyabouman, thanks for the fix ;)

@robodoo r+

PR #11857, opened 2025-01-24
[IMP] accounting: ocr free iap test credits for enterprise users

mcsi-odoo:
task-4328876


dade-odoo (approved):
Nice work @mcsi-odoo , thanks!

PR #11865, opened 2025-01-26
[FIX] tutorials/server_framework_101: update link for example of list

sk-nada:
previous link was for code that used tree in xml. The tree tag throws Traceback error in 18.0. This commit will provide the correct example with list tag instead of tree tag.


AntoineVDV (approved):
Thanks for the fix :)

@robodoo r+ squash

PR #11947, opened 2025-02-03
[FIX] administration: correct Fedora version

mege-odoo:
The Odoo 17 and 18 RPM builds target Fedora 38 instead of 36:

16: https://github.com/odoo/odoo/blob/16.0/setup/package.dffedora#L3
17: https://github.com/odoo/odoo/blob/17.0/setup/package.dffedora#L3
18: https://github.com/odoo/odoo/blob/18.0/setup/package.dffedora#L3

Note that both Fedora 36 and 38 are EOL:
https://docs.fedoraproject.org/en-US/releases/eol/


mege-odoo:
Confirmed with MOC we only target Fedora at the moment.

PR #11954, opened 2025-02-04
[IMP] pos_urban_piper: clarify combo product handling

meji-odoo:
In this commit:
===
- Added a note explaining that UrbanPiper does not support combo products.
- Users need to create a new product and add combo items as variants to use them with UrbanPiper.

task-4546712


LoredanaLrpz:
@robodoo r+

PR #11983, opened 2025-02-07
[FIX] coding_guidelines: fix the alphabetic orders for the import

hamo-odoo:
In odoo, as we suggest everyone to follow an alphebetic order.

The translator(`_`), ASCII is 95 and `a` ASCII value is 97 which '_' < 'a', the translator should be defined first in the import before the `api` not at the end of the import

This commits adds an example to import static class `Command` with a Capital alphabet with a ASCII value of `C` as 67 which will be before than the `_`
'C' < '_' < 'a'

This order of import should be followed while the import

### DIFF
--- a/content/contributing/development/coding_guidelines.rst
+++ b/content/contributing/development/coding_guidelines.rst
@@ -425,7 +425,7 @@ Inside these 3 groups, the imported lines are alphabetically sorted.
     from datetime import datetime
     # 2 : imports of odoo
     import odoo
-    from odoo import api, fields, models, _ # alphabetically ordered
+    from odoo import Command, _, api, fields, models # alphabetically ordered
     from odoo.tools.safe_eval import safe_eval as eval
     # 3 : imports from odoo addons
     from odoo.addons.web.controllers.main import login_redirect


AntoineVDV (approved):
Hi, I agree with the change, especially since IDEs will also follow this order by default, but could you please rebase your branch onto 16.0 and change the PR's target branch to 16.0 too, so that we cover all versions of the documentation?

hamo-odoo:
@AntoineVDV Done, Re-targeting to `16.0`

AntoineVDV:
@robodoo r+

AntoineVDV:
> Hello @AntoineVDV and @hamo-odoo Not against the change itself, but shouldn't we be more specific then and comment something like `# alphabetically (ascii) ordered`?

I'm planning on rewriting the whole guidelines soon™, so until then, I merge all change requests that go in the right direction, and I will elaborate further during the rewrite.

hamo-odoo:
> Hello @AntoineVDV and @hamo-odoo Not against the change itself, but shouldn't we be more specific then and comment something like `# alphabetically (ascii) ordered`?

For the reference, I have created the PR-  https://github.com/odoo/documentation/pull/12951

PR #11985, opened 2025-02-07
[IMP] developer/../view_architectures: improve calendar doc

aab-odoo:
This commit reworks the documentation of the calendar view, by using a style which is more aligned with other views. We also reworded and re-organized it.

### DIFF
--- a/content/developer/reference/user_interface/view_architectures.rst
+++ b/content/developer/reference/user_interface/view_architectures.rst
@@ -3188,105 +3188,275 @@ calendar.
    order to set the initial focus of the calendar on the period (see `mode`) around
    this date (the context key to use being `initial_date`)
 
-Their root element is ``<calendar>``. Available attributes on the
-calendar view are:
-
-:string:
-  string (default: ``''``)
-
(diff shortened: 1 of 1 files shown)


Feyensv (approved):
@robodoo delegate+

content/developer/reference/user_interface/view_architectures.rst
```diff
@@ -3188,105 +3188,242 @@ calendar.
...
+.. include:: view_architectures/root_attribute_string.rst
+
+.. _reference/view_architectures/calendar/components:
+
+Components
+----------
+
+Calendar views accept a single type of child elements: ``<field>``. Those fields
+are displayed in a popover, in the given order, which opens when a record
+(a calendar event) is clicked. Field nodes can have the following attributes:
```
mcm-odoo, line 3340: We need to note that the fields displayed in the popover are read only. They are not editable.

aab-odoo (commented):
robodoo r+

aab-odoo:
robodoo rebase-ff

PR #12136, opened 2025-02-21
[IMP] contributing: documentation macOS make install

samueljlieber:
Task: [#4593945](https://www.odoo.com/odoo/project.task/4593945)

This PR adds macOS specific instructions to the contributing documentation for installing Make.

The package manager `APT` is not primarily used on macOS, rather [Homebrew](https://brew.sh/) is often used.

### DIFF
--- a/content/contributing/documentation.rst
+++ b/content/contributing/documentation.rst
@@ -132,11 +132,15 @@ to make changes from the GitHub interface.
 
    .. tabs::
 
-      .. group-tab:: Linux and macOS
+      .. group-tab:: Linux
 
          .. code-block:: console
 
             $ sudo apt install make -y
+
+      .. group-tab:: macOS
+
+         Follow the `guide to install Make on macOS <https://formulae.brew.sh/formula/make>`_
 
       .. group-tab:: Windows
 


content/contributing/documentation.rst
```diff
@@ -132,12 +132,16 @@ to make changes from the GitHub interface.
 
    .. tabs::
 
-      .. group-tab:: Linux and macOS
```
AntoineVDV, line 135: This is not the end of the world, but the tab sync between this group and the other "Linux and macOS" groups will stop working. We can merge like that, but it could be worth looking for a way to keep them in sync, like referencing a guide that covers both systems.

samueljlieber: > This is not the end of the world, but the tab sync between this group and the other "Linux and macOS" groups will stop working.

@AntoineVDV  I looked for an existing article or guide on installing Make on Linux and macOS, but I did not find anything that covered both systems effectively.

Maybe to keep the tab sync, we can list the OS like this?

Screenshot 2025-02-21 at 10 35 35 AM


```rst
      .. group-tab:: Linux and macOS

         **Linux**

         .. code-block:: console

            $ sudo apt install make -y

         **macOS**

         .. code-block:: console

            $ sudo apt install make -y
```

AntoineVDV: No, that would defeat the purpose of the tabs. Let's split this group then, and if we find that it's too weird, we can still split all the "Linux and macOS" groups so that they're synced again.

AntoineVDV (approved):
@robodoo r+

PR #11889, opened 2025-01-27
[IMP] Documents: updating current doc

Donapi:
taskid-4213978

PR #12164, opened 2025-02-24
Update documents.rst

GoloCatMash:
Added Info, that this Documentation is outdated regarding Workspace, as they now are folders.


Feyensv:
Thanks for the heads up, it takes some time for the documentation to be updated for each new version, but we don't introduce deprecation notices, unless we really don't know when an outdated page will be updated.

With #11889, the documentation of the documents app should be up to date now.

PR #12189, opened 2025-02-26
[IMP] developer/localization: adapt link to fiscal position file

jco-odoo:


jco-odoo:
@odoo/account-dev-doc-review  Could you check as the link branch is approved and I would like to get it merged to avoid conflicts, ...  The only reason we needed this PR is because for a link to the fiscal positions csv, its filename changes.  (because of the CoA structure changing, ...)

william-andre (approved):

william-andre:
@robodoo r+

PR #12197, opened 2025-02-26
📝 Use Bytes unit instead of Octets (GB instead of Go)

gbv-it:
Fixed a small typo.


xpl-odoo (changes requested):
Hello @gbv-it - thanks for your contribution!

Could you re-target the PR to target version 16.0, since your suggestion applies your suggested changes apply from this verion? Thanks!

xpl-odoo (approved):
Thank you for your contribution @gbv-it !

@robodoo r+

PR #12223, opened 2025-02-28
[IMP] howtos/website_themes: Rework of the Gradients page

bram-odoo:
This PR updates current "Gradients" page:

- Corrections
- New anchors

Task-4316611

### DIFF
--- a/content/developer/howtos/website_themes/gradients.rst
+++ b/content/developer/howtos/website_themes/gradients.rst
@@ -5,13 +5,15 @@ Gradients
 In this chapter, you will discover how to:
 
 - Add a gradient to a section or a title.
-- Add your own gradient to the Website Builder palette.
+- Add a custom gradient to the Website Builder palette.
+
+.. _website_themes/gradients/standard:
 
 Standard
 ========
 
-In standard, you can select several gradients directly from the Website Builder. However, for custom
-themes, you must add the gradients directly in the section tag with a style attribute.
+In standard, several gradients can be selected directly from the Website Builder. However, for
+custom themes, gradients have to be added directly in the `section` tag with a `style` attribute.
 
 **Use**
 
@@ -21,7 +23,7 @@ themes, you must add the gradients directly in the section tag with a style attr
         <!-- Content -->
     </section>
 
-To apply a gradient to text, use a font tag with the `text-gradient` class.
+To apply a gradient to text, use a `font` tag with the `text-gradient` class.
 
 .. code-block:: xml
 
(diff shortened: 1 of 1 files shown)


content/developer/howtos/website_themes/gradients.rst
```diff
@@ -5,13 +5,15 @@ Gradients
...
+
+.. _howtos/website_themes/gradients/standard:
 
 Standard
 ========
 
-In standard, you can select several gradients directly from the Website Builder. However, for custom
-themes, you must add the gradients directly in the section tag with a style attribute.
+In standard, several gradients can be selected directly from the Website Builder. However, for
+custom themes, gradients have to be added directly in the `<section>` tag with a `style` attribute.
```
AntoineVDV, line 16:
```suggestion
custom themes, gradients have to be added directly in the `section` tag with a `style` attribute.
```

content/developer/howtos/website_themes/gradients.rst
```diff
@@ -21,7 +23,7 @@ themes, you must add the gradients directly in the section tag with a style attr
         <!-- Content -->
     </section>
 
-To apply a gradient to text, use a font tag with the `text-gradient` class.
+To apply a gradient to text, use a `<font>` tag with the `text-gradient` class.
```
AntoineVDV, line 26: same

content/developer/howtos/website_themes/gradients.rst
```diff
@@ -32,21 +34,20 @@ To apply a gradient to text, use a font tag with the `text-gradient` class.
 Custom
 ======
 
-You can also add your own custom gradients to the Website Builder. This way, the user can easily
-use them without manually recreating them.
+Add a custom gradients to the Website Builder. This way, the user can easily use them without
```
AntoineVDV, line 37:
```suggestion
Add a custom gradient to the Website Builder. This way, the user can easily use them without
```

AntoineVDV (approved):
@robodoo delegate+

content/developer/howtos/website_themes/gradients.rst
bram-odoo: Hello @AntoineVDV, in this suggestion you had the brackets around `font` and in the previous one, you remove them around `section`. I would keep the brackets as recommended here and as we talk about an HTML tag. What do you think?

AntoineVDV: I didn't make a suggestion here; the code excerpt is from your diff :D 

It's not a big deal, but I'd keep it simple and reference tags with `tag`, not `<tag>`. Otherwise, you'd need to do the same for other technical concepts that might be less readable.

For example, should we follow this style in documentation about RST, it could feel natural to mention RST roles with `:ref:`, but we'd want to stick to `seealso` to avoid having to write `.. seelaso::`.

bram-odoo: Woops, I'm an idiot sandwich xD My bad! Your example speaks for itself, I understand. Thanks for your input!

bram-odoo:
@robodoo r+

PR #12234, opened 2025-03-03
[IMP] payment_xendit: update Xendit information

nni-odoo:
After odoo/odoo#158445 Xendit supports tokenization allowing payments to be done on Odoo and Xendit. We also add specific for recurriny payments to enable a recurring payments from card configuration

### DIFF
--- a/content/applications/finance/payment_providers.rst
+++ b/content/applications/finance/payment_providers.rst
@@ -138,8 +138,8 @@ Online payment providers
      -
      -
    * - :doc:`Xendit <payment_providers/xendit>`
-     - The provider's website
-     -
+     - Odoo or the provider's website
+     - |V|
      -
      -
      -
--- a/content/applications/finance/payment_providers/xendit.rst
+++ b/content/applications/finance/payment_providers/xendit.rst
@@ -6,6 +6,10 @@ Xendit
 several Southeast Asian countries. It allows businesses to accept credit cards as well as several
 local payment methods.
 
+.. note::
+    Credit card payments are processed through Odoo, while all other payment methods are handled via
+    Xendit's website.
+
 .. _payment_providers/xendit/configure_dashboard:
 
 Configuration on the Xendit Dashboard
@@ -33,6 +37,10 @@ Configuration on the Xendit Dashboard
 #. In the :guilabel:`Webhook URL` section, enter your Odoo database's URL, followed by
    `/payment/xendit/webhook` (e.g., `https://example.odoo.com/payment/xendit/webhook`) in the field
    :guilabel:`Invoices paid` and click the :guilabel:`Test and save` button next to it.
+#. To allow recurring payments for credit cards, go to :menuselection:`Configuration: Payment
+   Channels` in the left part of the application page. Then, hover your mouse over the
+   :guilabel:`Visa, Mastercard, JCB, Amex` channel, click :guilabel:`View Details`, and enable
+   :guilabel:`Recurring Payments` by toggling the related switch.
 
 Configuration on Odoo
 =====================


content/applications/finance/payment_providers.rst
```diff
@@ -138,8 +138,8 @@ Online payment providers
      -
      -
    * - :doc:`Xendit <payment_providers/xendit>`
-     - The provider's website
-     -
+     - Odoo or The provider's website
```
auva-odoo, line 141:
```suggestion
     - Odoo or the provider's website
```

content/applications/finance/payment_providers.rst
```diff
@@ -154,6 +154,8 @@ Online payment providers
      providers. Payment providers allow customers to pay online, and bank accounts are added and
      configured in the Accounting app to do a :doc:`bank reconciliation
      <accounting/bank/reconciliation>`.
+   - On Xendit, credit card payments are done via Odoo while other payment methods are done via
+     Xendit's website.
```
auva-odoo, line 158: I know this relates to the table's content so it kinda makes sense to have it here, but I think it gets a bit lost here and should therefore be moved to the Xendit-specific doc page. I would also rephrase it slightly:
```suggestion
  Credit card payments are processed through Odoo, while all other payment methods are handled via Xendit's 
  website.
```

content/applications/finance/payment_providers/xendit.rst
```diff
@@ -33,6 +33,9 @@ Configuration on the Xendit Dashboard
 #. In the :guilabel:`Webhook URL` section, enter your Odoo database's URL, followed by
    `/payment/xendit/webhook` (e.g., `https://example.odoo.com/payment/xendit/webhook`) in the field
    :guilabel:`Invoices paid` and click the :guilabel:`Test and save` button next to it.
+#. To allow recurring payments for credit card, please enable recurring payments by going to
+   :menuselection:`Payment Channels`, click :guilabel:`View Details` button for the cards channel,
+   and toggle :guilabel:`Recurring Payments` on.
```
auva-odoo, line 38: Should this be in a tip? since it's not really part of the "API/webshook" flow? or is this mandatory, for everyone?
```suggestion
   .. tip:: 
   To allow recurring payments for credit cards, go to :menuselection:`Configuration: Payment Channels`. Then, hover your mouse over the :guilabel:`Visa, Mastercard, JCB, Amex` channel, click :guilabel:`View Details`, and enable :guilabel:`Recurring Payments` by toggling the related switch.
```
- to avoid repeating "recurring payments" 3x
- since the button is not visible unless you hover your mouse over the line
- (the last part of the sentence might be a bit overkill, you can just write *enable :guilabel:`Recurring Payments`*, I think it would be fine)

auva-odoo (changes requested):
Hey @nni-odoo thank you! I added a few comments, let me know what you think.

content/applications/finance/payment_providers/xendit.rst
nni-odoo: It's indeed not mandatory but I think this configuration step is quite important especially for people who's going to utilise subscription. Not sure if tip or bullet point would be better in this case...
Noted on the change!

nni-odoo:
@auva-odoo Thank you for the suggestions!

auva-odoo (approved):
@nni-odoo thank you for the changes! I force-pushed to fix the check fails and do some fine-tuning. The list item about recurring payments can stay as such, it's fine by me :slightly_smiling_face:
You can r+ when you're ready!

@robodoo delegate+

nni-odoo:
@robodoo r+

nni-odoo:
@auva-odoo thanks for review!

PR #12246, opened 2025-03-04
[FIX] core/github_link: deprecated werkzeug method

Feyensv:
Latest versions of werkzeug do not provide url_unparse anymore, leading to an error during the build:

AttributeError: module 'werkzeug.urls' has no attribute 'url_unparse'

This commit replaces the use of url_unparse by the urllib counterpart.


AntoineVDV (approved):
@robodoo r+

PR #12350, opened 2025-03-10
[IMP] studio: updating datetime field info

mcsi-odoo:
Task-4366107


content/applications/studio/fields.rst
```diff
@@ -182,9 +182,60 @@ Date & Time (`datetime`)
...
+  #. Add a :guilabel:`Date & Time` field and select the :guilabel:`Daterange widget`. This field
+     stores the counterpart date (i.e., the start date if the field added in step 1 is the
+     end date, or vice versa) and will be visible on the UI.
+  #. Give the field an appropriate label.
+  #. Select the field you added in step 1 from either the :guilabel:`End date field` or
+     :guilabel:`Start date field` dropdown, as relevant.
+  #. Update any other properties as needed and click :guilabel:`Close` in the upper right corner of
+     the screen.
+
+  .. tip::
```
auva-odoo, line 222: I wonder if this should be moved above the Date range bullet item/heading since it also concerns the Date & Time widget. What do you think? I'm not sure, let's discuss this?

content/applications/studio/fields.rst
```diff
@@ -483,11 +534,13 @@ Properties
 - :guilabel:`Help Tooltip`: To explain the purpose of a field, write a description under
   :guilabel:`Help Tooltip`. It is displayed inside a tooltip box when hovering with your mouse over
   the field's label.
-- :guilabel:`Placeholder`: To provide an example of how a field should be completed, write it under
-  :guilabel:`Placeholder`. It appears in light gray as a placeholder until a value is entered.
 - :guilabel:`Widget`: To change the default appearance or functionality of a field, select one of
   the available widgets.
+- :guilabel:`Placeholder`: To provide an example of how a field should be completed, write it under
+  :guilabel:`Placeholder`. It appears in light gray as a placeholder until a value is entered.
```
auva-odoo, line 540: I know you didn't change this, but I don't think repeating the field's label is necessary?

auva-odoo (changes requested):
Hi @mcsi-odoo thank you for your work; this is definitely worth documenting! 
I think the screenshot you added is a great idea and really helps clarify the different options. One general note about screenshots (and I know you didn’t do this): would you consider removing the center alignment for all of them :see_no_evil:? (You don't *have* to :wink: )

auva-odoo (changes requested):
Thank you for the changes @mcsi-odoo! I only have a few minor comments :slightly_smiling_face: 

This did, however, open another can of worms for me :see_no_evil: but I think it's best to address it in a separate task and PR, especially since it's not urgent at all.
(Also see my comment on line 201) I actually think we could restructure/review the page a bit to:

- Explain how to add a field and mention the general properties at the top of the page (they are a bit lost at the bottom right now).
- Move the Date & Time properties to the relevant section (before the widget-specific sections).
- Look into documenting other field/widget-specific properties for other field types as well.

What do you think?

auva-odoo (approved):
Thanks again @mcsi-odoo ! Nice work on this!! 

@robodoo r+

PR #12362, opened 2025-03-11
[FIX] general: correcting explanation of name column in access rights tab

mcsi-odoo:
Task-4639525


mcsi-odoo:
@auva-odoo , CTH confirmed that this update is ok for him.

content/applications/general/users/access_rights.rst
```diff
@@ -101,35 +101,36 @@ icon to remove a row.
...
-  :guilabel:`Model` column represents the common name of the menu/model, and the :guilabel:`Name`
-  column represents the technical name given to the model. For each model, enable the following
-  options as appropriate:
+- :guilabel:`Access Rights` tab: lists the first level of rights (models) that this group has. The
+  :guilabel:`Name` column represents the name for the current group's access to the model
+  selected in the :guilabel:`Model` column.
+
+  To link a new access right to a group, click :guilabel:`Add a line`. Select the appropriate model
+  from the :guilabel:`Model` dropdown, then enter a name for the access right in the
+  :guilabel:`Name` column. For each model, enable the following options as appropriate:
```
auva-odoo, line 114: I know you didn't change the list below but since you changed the text around it, can we just capitalize the first letter after the colon? :pray: :see_no_evil:

content/applications/general/users/access_rights.rst
```diff
@@ -101,35 +101,36 @@ icon to remove a row.
...
   - :guilabel:`Write`: users can edit the object's existing values.
   - :guilabel:`Create`: users can create new values for the object.
   - :guilabel:`Delete`: users can delete values for the object.
 
   .. tip::
-     First try searching for the common name of the model in the drop-down menu of the
-     :guilabel:`Model` column. The :guilabel:`Model` technical name can be found by expanding the
-     model common name, which can be done by clicking the :guilabel:`(external link)` icon.
+     There are no conventions to follow for naming an access right. However, it is advisable
+     to choose a name that easily identifies the access right in question.
```
auva-odoo, line 123:
```suggestion
     There are no conventions to follow for naming an access right. However, it is advisable
     to choose a name that easily identifies its purpose, e.g., `res.partner.purchase.manager`.
```
- to avoid repetitions?
- I’d suggest including examples in the text as well while keeping the image, as it helps with clarity. We try to make sure the text is understandable without relying on the image (which it is, mostly, but I think an example would make it clearer :wink: ). EDIT: maybe we could even explain the example a tiny bit more (and briefly mention the purpose of that example acess right in the text? What do you think?)
- I also wonder if we could avoid the *There are ...* wording here?

auva-odoo (changes requested):
Thank you for this PR @mcsi-odoo ! I only have one suggestion (and a tiny nitpick :see_no_evil: ). Let me know what you think!

content/applications/general/users/access_rights.rst
mcsi-odoo: I like your suggestions. And thanks for the reminder that the text should be (more) understandable without the image. Will see what I can do with the 'There are ...'.

auva-odoo (approved):
Thank you for the changes @mcsi-odoo LGTM :ok_hand: 

@robodoo r+

PR #12381, opened 2025-03-12
[IMP] accounting: update vendor bills tutorial link

afma-odoo:
task-4207830


afma-odoo:
error in the branch name

PR #12389, opened 2025-03-12
[FIX] ecommerce: require module install for email queue optimization

AntoineVDV:
In version 17.0 to 18.2, it is required to install the `sale_async_emails` module to enable the email queue optimization.

This commit also rephrases slightly the step to enable the cron because it should already be enabled, unless the user disabled it manually. Indeed, the cron is created enabled in versions 17.0 to 18.2, while it is created disabled but automatically enabled when toggling the `sale.async_emails` system parameter in master.

task-3872792


auva-odoo (approved):
Thank you @AntoineVDV !

@robodoo r+ up to saas-18.2

AntoineVDV:
@robodoo up to master

I'll remove the module mention in master but keep the new phrasing since it applies in 18.3+ too.

AntoineVDV:
@robodoo fw=skipmerge

PR #12419, opened 2025-03-14
[REF] ecommerce: restructure doc

auva-odoo:
task-4645039

Content updates will be done in subsequent PRs


auva-odoo:
@robodoo r+

PR #12528, opened 2025-03-17
IMP] developer/reference/cli: add --pre-upgrade-scripts, improve --upgrade-path

aj-fuentes:
### [IMP] developer/reference/cli: --pre-upgrade-scripts

Document the new cli option.

### [IMP] developer/reference/cli: improve docs of --upgrade-path
Be explicit about what we expect as parameter.

### DIFF
--- a/content/developer/reference/cli.rst
+++ b/content/developer/reference/cli.rst
@@ -92,7 +92,8 @@ Running the server
 
 .. option:: --upgrade-path <upgrade_path>
 
-   specify an additional upgrade path.
+   comma-separated list of directories from which additional upgrade scripts
+   are loaded.
 
 .. option:: --pre-upgrade-scripts <pre_upgrade_scripts>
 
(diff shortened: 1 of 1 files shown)


content/developer/reference/cli.rst
```diff
@@ -94,6 +94,12 @@ Running the server
 
    specify an additional upgrade path.
 
+.. option:: --pre-upgrade-scripts <pre_upgrade_scripts>
+
+   specify upgrade scripts to be run before loading base module when an upgrade
```
KangOl, line 99:
```suggestion
   comma-separated list of path to upgrade scripts to be run before loading base module when an upgrade
```

AntoineVDV (approved):
@robodoo delegate+

AntoineVDV:
@aj-fuentes Just don't forget to adapt your commits because they don't make sense in this context and repo ;)

aj-fuentes:
Hi @AntoineVDV, I don't know what makes sense here :) Would you give a suggestion please?

AntoineVDV:
@aj-fuentes Same logic as in other repos, I'd say, so something like: "[IMP] developer/cli: document `--pre-upgrade-scripts` option"

aj-fuentes:
OK, done.

PR #12541, opened 2025-03-18
[FIX] contributing: dead idiomatic python link

Megaaaaaa:
The link replaced is dead and doesn't lead to anything anymore. The new link is the exact same page but hosted on another website.


AntoineVDV (approved):
@robodoo r+

PR #12617, opened 2025-03-24
[IMP] Settings: Users Images and Icons

justinmallette:
Replace images and icons to match style standards.


justinmallette:
@samueljlieber ready for TR

content/applications/general/users.rst
```diff
@@ -51,40 +48,17 @@ login.
...
-.. tip::
-   Users are considered internal database users. Portal users are external users, who only have
-   access to the database portal to view records. See the documentation on :doc:`users/portal`.
-
-   Public users are those visiting websites, via the website's frontend.
-
-The :guilabel:`Portal` and :guilabel:`Public` user options do **not** allow the administrator to
-choose access rights. These users have specific access rights pre-set (such as, record rules and
-restricted menus), and usually do not belong to the usual Odoo groups.
-
```
Felicious, line 76: I noticed that the section on internal, portal, and public users  is being removed, and I wanted to clarify its importance before proceeding. This concept is critical for understanding user roles in Odoo, so it would be great if we could keep and improve it rather than remove it entirely. 

image
image

The PR description mentions that certain content is being removed because it isn’t found in the settings. However, the Manage Users page  does include these options—they’re just not easily discoverable due to outdated navigational instructions. Instead of deleting this section, let’s do two things to make it more accurate and helpful: 

1. Fix the Navigational Instructions : Let’s update the steps to align with our latest standards for finding the Manage Users page . This will help readers locate it without confusion.
2. Elaborate on User Definitions : On the Manage Users page , if you remove the "Internal Users" filter, you’ll see the other types of users (portal and public). For this section, let’s either:
- Provide a brief explanation of what portal users  and public users  are, or  
- Link to existing documentation that covers these concepts in detail (if they’re already explained elsewhere).
         
This way, we’ll ensure the section remains valuable and accurate while improving its clarity and usability. 

Let me know if you’d like to discuss how to refine this further—I’m happy to help!

content/applications/general/users.rst
```diff
@@ -121,23 +94,6 @@ To resolve the issue, either:
...
-times. Odoo offers a few different methods to reset a user's password.
-
-.. tip::
-   Odoo has a setting to specify the length needed for a password. This setting can be accessed by
-   navigating to :menuselection:`Settings app --> Permissions` section, and entering the desired
-   password length in the :guilabel:`Minimum Password Length` field. By default the value is `8`.
-
-.. image:: users/minimum-password-length.png
-   :align: center
-   :alt: Minimum Password Length highlighted in the Permissions section of General Settings.
```
Felicious, line 139: I noticed that the tip and image related to the Permissions section  are being removed, and I wanted to better understand the thought process behind this change.  

I double-checked in the Settings app, and the Permissions section  is still available in Odoo 18. Additionally, developer mode doesn’t need to be enabled to access it. Since this section was originally added for a specific reason, I’m wondering if we might want to keep it and instead ensure the information is accurate and up-to-date. 

Let me know your thoughts—I’m happy to help refine this part to make sure it’s as clear and useful as possible! 

image

Felicious (changes requested):
Hi @justinmallette 

Appreciate your work on updating this doc with our latest team guidelines. In reviewing the changes, I noticed that some of the removed content aligns with functionality that’s still present in the system, like managing users and their permissions. I believe these sections still hold significant value, and I've detailed in the PR comments. 

I’d love to discuss this further to understand the intention behind the deletions and find a solution that balances simplification with maintaining the value these sections bring to the documentation. Let me know what you think!

content/applications/general/users.rst
justinmallette: Hi Felicia! I removed them since I didn't see them in the database I was using. I'll take another look and make those adjustments

content/applications/general/users.rst
justinmallette: I also didn't see this in the database I was looking in. Are you seeing these in a specific account or in the runbot?

content/applications/general/users.rst
```diff
@@ -54,23 +51,17 @@ login.
 User type
 ---------
 
-:guilabel:`User Type` can be selected from the :guilabel:`Access Rights` tab of the user form,
-accessible via :menuselection:`Settings app --> Users section --> Manage Users`.
+:guilabel:`User Type` can be chosen on the :guilabel:`Manage Users` page clicking on the
```
Felicious, line 54:
```suggestion
:guilabel:`User Type` can be chosen on the :guilabel:`Manage Users` page by clicking on the
```

Felicious (approved):

content/applications/general/users/invitation-email.png
samueljlieber, line 1: This image (`content/applications/general/users/invitation-email.png`) needs to be compressed with `pngquant`

content/applications/general/users.rst
```diff
@@ -32,7 +31,6 @@ Fill in the form with all the required information. Under the :doc:`Access Right
 The list of applications shown is based on the applications installed on the database.
 
 .. image:: users/new-user.png
-   :align: center
    :alt: View of a user's form emphasizing the access rights tab in Odoo.
 
 After filling out all the necessary fields on the page, manually :guilabel:`Save`. An invitation
```
samueljlieber, line 36: There is no button with "Save" text, rather the _Save manually_ cloud icon is used here
```suggestion
After filling out all the necessary fields on the page, :icon:`fa-cloud-upload` :guilabel:`(Save
manually)`. An invitation email is automatically sent to the user, using the email in the
:guilabel:`Email Address` field. The user must click on the link included in the email to accept the
invitation, and to create a database login.
```

content/applications/general/users.rst
```diff
@@ -54,37 +51,31 @@ login.
 User type
 ---------
 
-:guilabel:`User Type` can be selected from the :guilabel:`Access Rights` tab of the user form,
-accessible via :menuselection:`Settings app --> Users section --> Manage Users`.
+:guilabel:`User Type` can be chosen on the :guilabel:`Manage Users` page by clicking on the
+:guilabel:`search` bar, and then setting a filter for either :guilabel:`Internal User` or
+:guilabel:`Portal User`. Once that filter is set, click :guilabel:`New` in the upper-left corner to
+create a user account with that setting.
```
samueljlieber, line 57: No need to guilabel "search" bar here, and we can add a reference to the using predefined filters documentation.

Also, I did not notice any change in the configuration when creating a user when a filter is set on this view, I don't think it is correct to state that the user account will be created with the settings of the filter.
```suggestion
:guilabel:`User Type` can be chosen on the :guilabel:`Manage Users` page by clicking on the search
bar, and then :ref:`setting a filter <search/preconfigured-filters>` for either :guilabel:`Internal
User` or :guilabel:`Portal User`.
```

samueljlieber (approved):
Hi @justinmallette, thanks for your work on this PR! I am approving with a few technical corrections. 

Please remove the `/users/user-type.png` image, as it is no longer used. And be sure to compress the other image in my comment :)

Thank you for your work!
.
@robodoo delegate=justinmallette

justinmallette:
@robodoo r+

justinmallette:
@robodoo retry

PR #12714, opened 2025-03-31
[IMP] howto/website_themes: "Pages" page update

bram-odoo:
This PR updates: 

- Bootstrap documentation urls
- Github urls
- Odoo version number into the manifest example

Task-4316707

### DIFF
--- a/content/developer/howtos/website_themes/pages.rst
+++ b/content/developer/howtos/website_themes/pages.rst
@@ -25,7 +25,7 @@ the following way.
 Each default page is a template with its own content saved into a record. This is the reason why,
 :ref:`custom pages are created within a record <website_themes/pages/theme_pages>`.
 
-The `<t -call='website.layout'>` has some variables that can be set:
+The `<t-call='website.layout'>` has some variables that can be set:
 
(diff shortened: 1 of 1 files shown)


content/developer/howtos/website_themes/pages.rst
```diff
@@ -284,7 +284,7 @@ The page templates has to be defined into the :file:`__manifest__.py` of the mod
       'name': 'Airproof Theme',
       'description': '...',
       'category': 'Website/Theme',
-      'version': '17.0.0',
+      'version': '18.0.0',
```
AntoineVDV, line 287: I don't know if it will work in a codeblock, but you should try replacing `18.0` with a `{BRANCH}` placeholder to avoid having to maintain that after every version freeze.

AntoineVDV (approved):
@robodoo delegate+

content/developer/howtos/website_themes/pages.rst
bram-odoo: @AntoineVDV I tested it and it seems to work like a charm. Thanks !

PR #12721, opened 2025-03-31
[IMP] howto/website_themes: Building blocks page update

bram-odoo:
This PR applies the following changes: 
- Change the Github urls to the correct version of Odoo
- Adapt the way of creating a snippet template and how to add it to the builder
- Add explanation about how to make a custom snippet "inner content"

Task-4316707


bram-odoo:
@robodoo r+

PR #12753, opened 2025-04-02
[IMP] accounting: adding info about the follow-up report

afma-odoo:
task-4260085

### DIFF
--- a/content/applications/finance/accounting/payments/follow_up.rst
+++ b/content/applications/finance/accounting/payments/follow_up.rst
@@ -76,22 +76,23 @@ Invoices` to have a detailed list of the overdue invoices.
 
 Additional options can be set:
 
-- :guilabel:`Reminders`: :guilabel:`Automatic` or :guilabel:`Manual`.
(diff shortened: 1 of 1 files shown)


dade-odoo (approved):
Hey @afma-odoo , no notes on any of your changes here :)

I will say (and I'm sorry that I didn't catch this on the earlier PR), it would be nice to clean up the bulleted lists in the **Follow-ups for one customer** section. There's no need for capital letters at the beginning or periods at the end of each bullet if it's not a full sentence (under "Additional options can be set:"), and there's no need for the semicolons and periods when the bulleted items come after a complete sentence ("To manually send a payment...").

I'd be fine changing that here since it's the first major version with this text, but let me know if you or @auva-odoo disagree.

auva-odoo (approved):
@afma-odoo I agree with @dade-odoo regarding the bulleted list items in the *Follow-ups for one customer section*. Other than that, this PR looks great, nice work! :clap: 

@robodoo delegate+

afma-odoo:
@dade-odoo @auva-odoo Thank you for the review! ☺️

@robodoo r+

PR #12826, opened 2025-04-08
[IMP] ecommerce: delete duplicated text

lman-odoo:


auva-odoo (approved):
Good catch @lman-odoo! :ok_hand: 

@robodoo r+

PR #12852, opened 2025-04-10
[FIX] IAP: fix format type

igbe-odoo:
This commit fix a typo in the IAP doc page, in the user docs. One line in the IAP services section was bold and there was a :icon: visible. This commit fix this issue.

The formating error was introduced here: 21654b42e8b5593ff0e215dad0c8963f122c0015, where there was a missing "`"

no-task


xpl-odoo (approved):
Thanks for the fix @igbe-odoo

@robodoo r+

PR #12889, opened 2025-04-14
[IMP] reconciliation models: menu item removed

dade-odoo:
task-4652251

Replace instructions for accessing reconciliation models from the configuration menu with instructions to access them from the bank journal on the dashboard to match 18.2 UI.


auva-odoo (approved):
@robodoo r+

PR #13032, opened 2025-04-24
[IMP] accounting: SEPA pain version + fixes

auva-odoo:


auva-odoo:
@robodoo r+ up to saas-17.4

PR #13034, opened 2025-04-24
[IMP] administration: clarify supported browsers

aab-odoo:
This commit rewords a little bit the part of the doc about supported browsers, with an emphasis on the fact that the browser must always be up-to-date, which customers sometimes forget (and report issues because of this).

See https://github.com/odoo/odoo/pull/184894


xpl-odoo (approved):
@aab-odoo LGTM!

@robodoo delegate+

aab-odoo:
robodoo r+

PR #13047, opened 2025-04-24
[IMP] accounting: cheat sheet styles

samueljlieber:
Task: [#4743756](https://www.odoo.com/odoo/project/3835/tasks/4743756)
FWP up to `master`

This PR improves the styles on the [Accounting Cheat Sheet](https://www.odoo.com/documentation/16.0/applications/finance/accounting/get_started/cheat_sheet.html) by adding borders which adds more contrast to the highlights.

- [Current Accounting Cheat Sheet](https://www.odoo.com/documentation/16.0/applications/finance/accounting/get_started/cheat_sheet.html)
- [This PR Accounting Cheat Sheet](https://runbot233.odoo.com/runbot/static/build/79031168-16-0/logs/build/html/applications/finance/accounting/get_started/cheat_sheet.html)

Screenshot 2025-04-24 at 12 42 23 PM

As well as corrects the color of the Reconciliation button to use the Odoo primary and secondary colors:

Screenshot 2025-04-24 at 12 43 37 PM

### DIFF
--- a/static/css/accounting.css
+++ b/static/css/accounting.css
@@ -21,6 +21,7 @@ main.index .toctree-wrapper > .row:first-child > .col-md-3 {
 
 .related {
     background-color: hsl(317deg 16% 90%) !important;
+    border: 1px solid #000000 !important;
     transition: .3s;
 }
 .secondary {
@@ -46,13 +47,13 @@ label:hover,
 .accounts-table dl {
     margin: 0;
     padding: .5rem 0;
+    border: 1px solid transparent;
 }
 
 .accounts-table h4, .accounts-table h5 {
     font-weight: 700;
     text-transform: uppercase;
     padding: .5rem;
-
 }
 
 .accounts-table h4 {
@@ -61,11 +62,13 @@ label:hover,
 /* table root */
 .accounts-table > div {
     display: flex;
+    border: 1px solid transparent;
 }
 /* P&L & Balance Sheet columns */
 .accounts-table > div > div {
     flex: 1;
     padding: .5rem;
+    border: 1px solid transparent;
 }
 
 .accounts-table > div > div:first-child {
@@ -76,6 +79,7 @@ label:hover,
 .accounts-table > div > div div {
     display: flex;
     flex-direction: column;
+    border: 1px solid transparent;
 }
 
 .accounts-table > div > div div > h5 {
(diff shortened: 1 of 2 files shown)


static/css/accounting.css
```diff
@@ -11,15 +11,21 @@ main.index .toctree-wrapper > .row:first-child > .col-md-3 {
 .doc-aside p{
     padding: .5rem;
 }
+.doc-aside div {
+    border: 1px solid rgba(255, 255, 255, 0);
```
rugo-odoo, line 15: I'm not sure what is the point of this. This will put an **invisible** border (?) What are we trying to achieve ?

It happens a lot in the pr.

samueljlieber: Hi @rugo-odoo, the reason I added an 1px invisible border is to avoid the 1px layout shift that happens when the 1px black border appears on hover :)

AntoineVDV:
@robodoo delegate=rugo-odoo

static/css/accounting.css
rugo-odoo: I believe much of this is unnecessary.

1) For the one with `.highlight-op`, we can just change the border color. We already have a border so it won't shift.

2) If I'm not mistake, to prevent the shift, we can just use `box-sizing: border-box` for the elements we need. [border-box](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) will not add any width/height with the border.

It will be cleaner and easier to maintain.

rugo-odoo (changes requested):

samueljlieber:
Hi @rugo-odoo! Thank you for your helpful suggestions :) in https://github.com/odoo/documentation/commit/e92d9c0cc7bac737514cdddc29154a99733076e3 I was able to replace the border color for `.highlight-op` like you suggested. 

For your second point, about using `box-sizing: border-box`, I don't believe this will work since the `.related` elements don't already have a border, so an initial transparent border is still necessary (as far as I can tell from my testing). I did end up setting this transparent border in the `accounting.css` instead of in the [`accounting.js` (like I was before)](https://github.com/odoo/documentation/commit/c71c50f2bc14f28824d806047cc936d46b97e8ab#diff-033d8a280d3fe9c8f1ac6125ab1c9bed6ed76b762228ba929872cd02594b606f). 

With these changes, things should be cleaner and easier to maintain! 

Thank you for pushing for changes :) ready for another look! 🙏

rugo-odoo (approved):
@robodoo r+

PR #13134, opened 2025-04-29
[IMP] mail: from filter ICP missing part

Abridbus:
Functionals support noted that the from filter part without mail server was missing.
But was present on ancient documentation:
https://www.odoo.com/documentation/16.0/applications/general/email_communication/email_servers.html#use-a-default-from-email-address

Let's re-add it to clarify. And adding the anchoring

opw-4568395

### DIFF
--- a/content/applications/general/email_communication/email_servers_outbound.rst
+++ b/content/applications/general/email_communication/email_servers_outbound.rst
@@ -17,6 +17,16 @@ emails. The deliverability is optimized for this subdomain as it uses Odoo’s D
    If the database subdomain is `company-name.odoo.com` and all mailing configurations are the
    default ones, all emails will be sent from `notifications@company-name.odoo.com`.
 
+.. _email-outbound-default-from-filtering:
+
+This configuration is handled by the system parameter `mail.default.from_filter`.
+In case where the sender's domain do not match the value of this parameter, the notification address
+is used instead. Multiple values can be defined in this system parameter: comma-separated, domains
+or full email addresses are all allowed. Once an :ref:`outgoing mail server is configured
+<email-outbound-different-servers-personalized>`, the system parameter is no longer considered
+and the value used is the :ref:`FROM filtering
+<email-outbound-different-servers-personalized-from-filtering>` of the mail server.
+
 .. image:: email_servers_outbound/diagram-inbound-mailing-method.png
    :alt: Odoo’s default outbound messages configuration.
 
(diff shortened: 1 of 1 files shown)


content/applications/general/email_communication/email_servers_outbound.rst
```diff
@@ -17,6 +17,15 @@ emails. The deliverability is optimized for this subdomain as it uses Odoo’s D
...
 
+.. _email-outbound-default-from-filtering:
+
+This configuration is done by the system parameter `mail.default.from_filter`.
+In case the sender's domain do not match the value of this parameter, the notification address
+will be used.
+Multiple values can be defined, comma-separated, domains or full email addresses.
+Once an outgoing mail server is active, the parameter is is no longer considered the value used
+will be the :ref:`email-outbound-different-servers-personalized-from-filtering` of the mail server.
+
```
samueljlieber, line 28: Suggesting to use a note admonition here, along with some slight rewording for clarity. 

We can also use the existing `email-outbound-default` anchor for this section.
```suggestion
.. note::
   This configuration is handled by the `mail.default.from_filter` system parameter. In cases where
   the sender's domain does not match the value of this parameter, the notification address is used
   instead.

   Multiple values can be defined in this system parameter: comma-separated, domains, or full email
   addresses are all allowed.

   Once an :ref:`outgoing mail server is configured
   <email-outbound-different-servers-personalized>`, the system parameter is no longer considered
   and the value used is the :ref:`FROM filtering
   <email-outbound-different-servers-personalized-from-filtering>` of the mail server.
```

content/applications/general/email_communication/email_servers_outbound.rst
```diff
@@ -205,6 +214,8 @@ When an email is sent from Odoo, the following sequence is used to choose the ou
    higher the priority is). Failing to do so, the first server is determined by the servers' names,
    using alphabetical order.
 
+- If there is no mail server, Odoo will rely on the :ref:`system parameter <email-outbound-default-from-filtering>` value.
+
```
samueljlieber, line 218: Slight correction to avoid future tense, update the anchor that is used, and break the line before 100 chars :)
```suggestion
- If there is no mail server, Odoo relies on the :ref:`system parameter <email-outbound-default>`
  value.
  
```

samueljlieber (approved):
Hi @Abridbus! Thank you for this PR!

Im approving with a couple small adjustments that I'll push up after this review:
1. Commit message: [ADD] -> [IMP] (since this is an improvement to an existing doc)
2. Please see my suggestions below for specifics

I will delegate merge to you once checks pass, thank you!

samueljlieber:
@robodoo delegate=Abridbus

Abridbus:
Thanks for the review @samueljlieber 

I applied the change but I kept the anchors that I created.
I wanted them to send directly to the correct sentences! :) 

Can I ask you to double check again that everything is ok? 

Thanks!

samueljlieber:
Made a couple RST syntax adjustments so checks pass in https://github.com/odoo/documentation/commit/7901b6e1965569181790300e1747bae68c838a01 :)

samueljlieber:
@robodoo r+

samueljlieber:
@robodoo retry

PR #13151, opened 2025-04-30
[FIX] rpc: fix code examples for tutorial and service

leclerc-leo:
I am a newbie from April and was following the "Discover the web framework" guide (https://www.odoo.com/documentation/18.0/developer/tutorials/discover_js_framework/02_build_a_dashboard.html#call-the-server-add-some-statistics) and saw that the link and example provided were outdated.

For example, it was like this:
```js
setup() {
      this.rpc = useService("rpc");
      onWillStart(async () => {
         const result = await this.rpc("/my/controller", {a: 1, b: 2});
         // ...
      });
}
```
Which is now
```js
import { rpc } from "@web/core/network/rpc";
// ... 

setup() {
      onWillStart(async () => {
            const syncResult = await rpc("/my/controller", {a: 1, b: 2});
      })
      // ...
}
```

I also changed the code example inside the Reference for the RPC service as it was also out of date. There might be a need to move this to another category as it might not be a Service anymore, I wasn't able to determine is it would be best or not (or which one would be best).

### DIFF
--- a/content/developer/tutorials/discover_js_framework/02_build_a_dashboard.rst
+++ b/content/developer/tutorials/discover_js_framework/02_build_a_dashboard.rst
@@ -199,22 +199,23 @@ Let's improve the dashboard by adding a few dashboard items to display *real* bu
 The `awesome_dashboard` addon provides a `/awesome_dashboard/statistics` route that is meant
 to return some interesting information.
 
-To call a specific controller, we need to use the :ref:`rpc service <frontend/services/rpc>`.
+To call a specific controller, we need to use the :ref:`rpc <frontend/services/rpc>` function.
 It only exports a single function that perform the request: :code:`rpc(route, params, settings)`.
 A basic request could look like this:
 
 .. code-block:: js
 
+   import { rpc } from "@web/core/network/rpc";
+   // ...
+
    setup() {
-         this.rpc = useService("rpc");
-         onWillStart(async () => {
-            const result = await this.rpc("/my/controller", {a: 1, b: 2});
-            // ...
-         });
+      onWillStart(async () => {
+         const result = await rpc("/my/controller", {a: 1, b: 2});
+      })
+      // ...
    }
 
-#. Update `Dashboard` so that it uses the `rpc` service.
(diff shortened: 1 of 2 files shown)


content/developer/tutorials/discover_js_framework/02_build_a_dashboard.rst
```diff
@@ -199,22 +199,23 @@ Let's improve the dashboard by adding a few dashboard items to display *real* bu
 The `awesome_dashboard` addon provides a `/awesome_dashboard/statistics` route that is meant
 to return some interesting information.
 
-To call a specific controller, we need to use the :ref:`rpc service <frontend/services/rpc>`.
+To call a specific controller, we need to use the :ref:`rpc <frontend/services/rpc>`.
```
AntoineVDV, line 202:
```suggestion
To call a specific controller, we need to use the :ref:`rpc <frontend/services/rpc>` function.
```
or
```suggestion
To call a specific controller, we need to use :ref:`rpc <frontend/services/rpc>`.
```

content/developer/tutorials/discover_js_framework/02_build_a_dashboard.rst
```diff
@@ -199,22 +199,23 @@ Let's improve the dashboard by adding a few dashboard items to display *real* bu
...
-         });
+      onWillStart(async () => {
+         const result = await rpc("/my/controller", {a: 1, b: 2});
+      })
+      // ...
    }
 
-#. Update `Dashboard` so that it uses the `rpc` service.
-#. Call the statistics route `/awesome_dashboard/statistics` in the `onWillStart` hook.
+#. Update `Dashboard` so that it uses the `rpc` and call the statistics route `/awesome_dashboard/statistics`.
```
AntoineVDV, line 218: same

AntoineVDV (commented):
@robodoo delegate=aab-odoo delegate=fdardenne

fdardenne (approved):
Thank you for this update. It's still not perfect indeed since it's not a service anymore but let's merge it and improve it incrementally.
@robodoo r+

PR #13162, opened 2025-05-01
[IMP] Lunch: fix vendors error

larm-odoo:
Fixing an issue found by MFAR: "Hi, just noticed this, it should be Extra 2 quantity instead of Extra 1, as it's talking about the second extra and not the first one again https://www.odoo.com/documentation/18.0/applications/hr/lunch/vendors.html"

Just changing "1" to "2" in the very end of the doc.


larm-odoo:
Hi @slinkous - SUPER small change in this PR - just changing the error in the last note at the end of the doc!

slinkous (approved):

larm-odoo:
Hi @samueljlieber - since this is such a tiny change, I am moving right to tech review.

samueljlieber (approved):
Thanks @larm-odoo!
.
@robodoo r+

PR #13186, opened 2025-05-02
[IMP] Appraisals: Update main doc

larm-odoo:
Updating to new standards and updating images.


larm-odoo:
This is a duplicate PR

PR #13206, opened 2025-05-05
[IMP] Fleet: updating main doc larm

larm-odoo:
Updating to new standards


larm-odoo:
The info in this doc is being moved to the New Vehicles doc and the Models & Manufacturrs doc. Closing this PR as it is no longer needed due to reorganization.

PR #13222, opened 2025-05-06
[IMP] core: support reinit modules

HydrionBurst:
https://github.com/odoo/odoo/pull/206408

### DIFF
--- a/content/developer/reference/cli.rst
+++ b/content/developer/reference/cli.rst
@@ -82,6 +82,18 @@ Running the server
     comma-separated list of modules to update before running the server.
     Use ``all`` for all modules. (requires :option:`-d`).
 
+.. option:: --reinit <modules>
+
+    A comma-separated list of modules to reinitialize before starting the server.
+    (requires :option:`-d`).
+
+    The reinitialization is similar to a simple upgrade without executing any upgrade script.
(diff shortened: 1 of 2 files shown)


AntoineVDV (approved):
@robodoo delegate+

content/developer/reference/cli.rst
```diff
@@ -82,6 +82,11 @@ Running the server
     comma-separated list of modules to update before running the server.
     Use ``all`` for all modules. (requires :option:`-d`).
 
+.. option:: --reinit <modules>
+
+    comma-separated list of modules to reinitialize before running the server
+    (requires :option:`-d`).
+
```
Julien00859, line 89: I would add a paragraph to explain how this option is different than `-u`, and maybe update the entry for `-u` to explain the difference with `--reinit`. I would also emphasis that `--reinit` is a development option that should not be used in production, along with the risks of using it in production (like to change any manual configuration done in the ir.rules, if I did correctly understand the purpose of the command).

Julien00859 (approved):

Julien00859 (approved):

Julien00859:
@AntoineVDV we are ready, the community branch has been r+ :)

AntoineVDV:
It was delegated already^^

@robodoo r+

PR #13310, opened 2025-05-10
[IMP] purchase: THD feedback about UBL format & alternative upload

Felicious:
Implement THD's feedback that the schema is UBL, not Order-X, and add a simpler workflow

image


Felicious:
Hi @samueljlieber ! Sending this straight to tech review (: I found this feedback left by the PO while cleaning up my old tasks :sob:

samueljlieber (approved):
Nice one @Felicious, looks great to me!
.
@robodoo r+

PR #13324, opened 2025-05-12
[IMP] Live chat: reports

jero-odoo:
Updating Live chat reports doc for 18
Updating content and images to reflect UI changes and updated style guidelines
Removing duplicate "essentials" content


content/applications/websites/livechat/reports.rst
```diff
@@ -32,36 +32,37 @@ participant name and country, session duration, the number of messages, and the
 provides access to the complete transcripts of live chat sessions.
 
 To access this report, navigate to :menuselection:`Live Chat app --> Report --> Sessions History`.
+Each live chat session is represented by a Kanban card.
```
slinkous, line 35: Are these actually Kanban cards? I thought you needed the the moveable card-in-columns structure to call it that. Maybe just 'card'?

slinkous (approved):
Great PR, one tiny semantic thing. 

And if you wanted an entirely unnecessary aesthetic change, you could bump up all the next-level headings under "Available reports" to be the same level, just so folks can index what's on the page faster when they arrive.

content/applications/websites/livechat/reports.rst
jero-odoo: Yeah, they do not technically have the "stages/column" structure, but the view for the report is "Kanban" so I think that still makes them Kanban cards
image

content/applications/websites/livechat/reports.rst
```diff
@@ -25,47 +22,48 @@ To access a drop-down menu of all the available reports, navigate to :menuselect
...
+
+Export sessions history
+-----------------------
+
 The information in this report can be exported, or inserted into a spreadsheet.
 
-Click the :guilabel:`⚙️ (gear)` icon to the right of the :guilabel:`History` page title, in the
-top-left corner. Doing so reveals a drop-down menu.
+On the *Sessions History* report, click the :icon:`oi-view-list` :guilabel:`(List)` icon to switch
+to list view. Next, click the :icon:`fa-cog` :guilabel:`(cog)` icon to the right of the
```
samueljlieber, line 52: Nitpic- please use "gear" for the label
```suggestion
to list view. Next, click the :icon:`fa-cog` :guilabel:`(gear)` icon to the right of the
```

samueljlieber (approved):
Thanks @jero-odoo, approving with a nitpick!
..
@robodoo delegate=jero-odoo

jero-odoo:
@robodoo r+

PR #13360, opened 2025-05-13
[IMP] accounting: adding Liasse Fiscale section lost in last update

afma-odoo:
PR created to add a whole chapter lost in the last update. 
**Forward-port to be created up to saas-17.4 only** to avoid conflicts with 18.0 recent updates.
Thanks ☺️


auva-odoo (approved):
@robodoo r+ up to saas-17.4

PR #13361, opened 2025-05-13
[IMP] accounting: adding in 18.0 Liasse Fisc section lost in update

afma-odoo:
PR created to add a chapter lost in the last update. **Only for 18.0.**
**No forward-port to be created** to avoid conflicts with 18.1 updates.
Thanks ☺️


auva-odoo (approved):
@robodoo r+ up to 18.0

PR #13364, opened 2025-05-14
[FIX] accounting: fix financial report typo

dade-odoo:
Minor typo fix - "ES sales list" in the financial reports section of the main accounting page should be "EC sales list".


auva-odoo:
@robodoo r+

PR #13438, opened 2025-05-19
[ADD] rental: manage deposits

samueljlieber:
Task: [#4742978](https://www.odoo.com/odoo/project/3835/tasks/4742978)

This PR adds the **Manage deposits** document to the **Rental** scope.


content/applications/sales/rental/manage_deposits.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. example::
+   If this deposit service product is intended for the *Digital Camera* rental product, the name may
+   be entered as `Digital Camera Deposit`.
+
+Next, set the :guilabel:`Product Type` to :guilabel:`Service` and assign a :guilabel:`Sales Price`
+with the amount to collect for the deposit. Choose whether or not to include :guilabel:`Sales Taxes`
+for this deposit service product.
+
+Be sure that the :guilabel:`Rental` checkbox is ticked at the top of the product form.
```
samueljlieber, line 35: Potentially removing this line, however, by not configuring the deposit service product as **Rental**, it becomes unavailable for the combo product workflow. TBD.
```suggestion
```

StraubCreative (approved):
Hi @samueljlieber 

I think you captured the use case enough to publish and get more feedback from the community. My only suggestion is to be more concise in the writing in some spots. 

If you would, please check with an expert to get their thoughts before merge.

Thanks for this, looking forward to seeing it live 👍

samueljlieber:
Hi @StraubCreative thanks for your review! I did check with MOCH and the only change was to add the following notice to the important admonition:

```rst 
The :guilabel:`Configure your product` pop-up only appears if the optional product is set on the rental product.
```

Should be good to go!

samueljlieber:
@robodoo r+

PR #13453, opened 2025-05-21
[IMP] inventory: sendcloud pickup points

Felicious:
Include pickup point feature in Sendcloud


samueljlieber (approved):
Looks good to me @Felicious! Thanks!
.
@robodoo r+

PR #13500, opened 2025-05-22
[IMP] accounting: adding info on Teledec for multi-company

afma-odoo:
task-4802222


auva-odoo (approved):
Thank you @afma-odoo !

@robodoo r+

PR #13590, opened 2025-05-30
[IMP] Lunch: Update products

larm-odoo:
Updating to new standards.


larm-odoo:
Hi @jero-odoo - this is ready for a peer review (should be super short).

jero-odoo (approved):
Looks good to me @larm-odoo

PR #13620, opened 2025-06-06
[IMP] payment_providers: list of supported payment methods

auva-odoo:
task-4672735

### DIFF
--- a/content/applications/finance/payment_providers/nuvei.rst
+++ b/content/applications/finance/payment_providers/nuvei.rst
@@ -45,43 +45,52 @@ Configuration on Odoo
    You can also test Nuvei using the :ref:`test mode <payment_providers/test-mode>` and your Sandbox
    Dashboard values.
 
-.. _payment_providers/nuvei/services:
-
-Payment methods
-===============
-
-Most Nuvei payment methods are **region specific**. The supported payment methods and brands for
-each country are listed below:
-
-+---------------------------------+----------------------------------+
-| **Argentina**                   | **Ecuador**                      |
-|                                 |                                  |
-| - Boleto                        | - Card (AMEX, Mastercard, Visa)  |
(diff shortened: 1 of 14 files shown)


content/applications/finance/payment_providers/nuvei.rst
```diff
@@ -45,43 +45,52 @@ Configuration on Odoo
...
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :widths: auto
+
+   * - Payment method
+     - Currencies
+     - Countries
+   * - AstroPay TEF
```
masi-odoo, line 70: @auva-odoo is it possible instead of `AstroPay TEF` we can add the detail `AstroPay TEF (Banco Guayaquil, Banco Pichincha, Facilito)`?

We had it like this in the documentation because Ecuador clients do not have much knowledge of what AstroPay is compared to the local alternative methods it offers (local banks mentioned above).

CC @andg-odoo

content/applications/finance/payment_providers/nuvei.rst
```diff
@@ -45,43 +45,52 @@ Configuration on Odoo
...
+     - USD
+     - Ecuador
+   * - Boleto
+     - BRL
+     - Brazil
+   * - Card (American Express, Mastercard, Visa)
+     - All
+     - Argentina, Brazil, Canada, Chile, Colombia, Ecuador, Mexico, Peru, United States, Uruguay
+   * - Local Payments
```
masi-odoo, line 79: @auva-odoo same case here, `Pagos Locales` does not have much meaning for an Uruguayan client, so `Pagos locales (Abitab, RedPagos)` would be better.

CC @andg-odoo

auva-odoo:
@masi-odoo We had originally decided not to include brands in the list of supported payment methods, for any provider. However, I checked in with ANV, and given that “Local payments” is quite a generic name in this case, I’ll go ahead and add the relevant brands back for this one.

PR #13647, opened 2025-06-10
[IMP] l10n: FDM odoo certified versions

LoredanaLrpz:
task-4742924
Removed "only for major versions" + useless table


LoredanaLrpz:
@robodoo r+

PR #13669, opened 2025-06-12
[IMP] conf: remove pillow inventory, cache werkzeug

xmo-odoo:
The pallets website is regularly unreachable (due to DNS failures so it's not even that we get rate limited even though we probably hit it excessively), which causes a warning during documentation build, which in turn causes a completely unnecessary build failure.

Some werkzeug links seem to be generated (in the backend docs), so just commit the werkzeug inventory file as it's pretty small.

Also remove the pillow inventory entirely, it doesn't seem to cause issues but also seems completely unused.


xmo-odoo:
> No idea where Pillow might have been used some day in the docs, so 🟢

It was probably in a docstring: the only location I found the werkzeug intersphinx support to be used is in the http backend APIdoc e.g. https://www.odoo.com/documentation/18.0/developer/reference/backend/http.html#odoo.http.Response. Intersphinx is invoked by the 

    :class:`werkzeug.wrappers.Response`

and the

    :type result: Union[Response, werkzeug.wrappers.BaseResponse,
            werkzeug.exceptions.HTTPException, str, bytes, NoneType]

(possibly a few others), with it the objects are cross-linked, otherwise they're just written literally (in fact we can see this with `BaseResponse` in `load`, as that type does not exist anymore so intersphinx doesn't find it and can't link it).

xmo-odoo:
@robodoo r+

xmo-odoo:
@robodoo squash

(just noticed a typo in the commit message)

PR #13706, opened 2025-06-16
[FIX] iot: link to non-french url

andg-odoo:
A previous PR added a new link to the supported printers that work with the iot box, however, it has fr_FR in the URL so it will always redirect to the french webpage. The language has been removed from here to solve that.


Felicious (approved):
Thanks for making the fix, @andg-odoo !

@robodoo r+

PR #13752, opened 2025-06-20
[IMP] conf: vendor python inventories

Xavier-Do:
Reaching external ressources while running tests is an important source of random error and can even causes deterministic errors when a service is down for a while.

Runbot will soon block all external requests, this commit follows changes introduced in #13669 and vendors the python inventories.


AntoineVDV (approved):

Xavier-Do:
@robodoo r+

PR #13773, opened 2025-06-23
[FIX] spreadsheet: typo in chart example text

mcsi-odoo:


auva-odoo (approved):
@robodoo r+

xpl-odoo (approved):
Delegating just in case you found something else to address in the meantime
@robodoo delegate+

PR #13856, opened 2025-06-27
[IMP] partner_autocomplete: remove chatter section

RicGR98:
The chatter infos were removed in task 4416928

task-id none


RicGR98 (commented):
Up @odoo/crm-doc-review :)

PR #13929, opened 2025-07-07
[IMP] ecommerce: remove tip about starred products being shown first

xpl-odoo:
The information is wrong; starring a product in the backend does not make it appear first on the frontend.


xpl-odoo:
@robodoo r+

PR #13850, opened 2025-06-26
[ADD] accounting/l10n_co: Add doc on automatic currency Colombia

masi-odoo:
Add documentation on the automatic currency rate for CO which is an integration with the "Banco de la Republica" service.

Drive with photos: https://drive.google.com/drive/folders/1jdPv6PNF_ekTuNxqw9GXoIJgII9HN-Eh?usp=sharing

PR #13967, opened 2025-07-10
[IMP] l10n_co: Add automatic currency rate

StraubCreative:
Add documentation on the automatic currency rate for CO which is an integration with the "Banco de la Republica" service.
Continuation of #13850


StraubCreative:
Add missing line break on [b59233e](https://github.com/odoo/documentation/pull/13967/commits/b59233ed217dbc172f9e769983cc86e1a01eb32b)

All checks pass ✅ 
..
@robodoo r+

PR #13994, opened 2025-07-11
[FIX] l10n_co: Fix automatic currency rate

dade-odoo:
Align content with guidelines (menu navigation, guilabels, lists, and screenshot)


auva-odoo (approved):
Thank you @dade-odoo

@robodoo r+

PR #13998, opened 2025-07-11
[IMP] extensions: add copy to markdown button

mattismegevand:
Add a "Copy to Markdown" button using [Turndown](https://github.com/mixmark-io/turndown) to convert HTML docs to Markdown, with support for syntax-highlighted code blocks. This will allow an easier usage of LLM.

Here's how it look (copy button is next to version selector, right window show raw output and preview):
image

### DIFF
--- a/extensions/odoo_theme/__init__.py
+++ b/extensions/odoo_theme/__init__.py
@@ -16,6 +16,8 @@ def setup(app):
     app.add_js_file('js/menu.js')
     app.add_js_file('js/page_toc.js')
     app.add_js_file('js/switchers.js')
+    app.add_js_file('js/turndown.js')
+    app.add_js_file('js/copy_page_md.js')
 
     roles.register_canonical_role('icon', icon_role)
 
--- a/extensions/odoo_theme/layout_templates/header.html
+++ b/extensions/odoo_theme/layout_templates/header.html
@@ -19,5 +19,10 @@
     <div class="d-flex justify-content-end ms-auto">
         {%- include "layout_templates/language_switcher.html" %}
         {%- include "layout_templates/version_switcher.html" %}
+        <div class="ms-2">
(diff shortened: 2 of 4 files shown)


mattismegevand:
Hello! Yes, this is something I’ve seen in other documentation systems and found very practical.
It’s especially useful for dev docs, as it allows you to copy clean Markdown and ask questions to LLMs with better context and fewer hallucinations.
I chose a client-side approach to avoid any backend changes, but I’m happy to adjust if you think it would be better handled server-side let me know!

AntoineVDV:
Ok, I see. I thought it was meant for a specific project, like a RAG chatbot fetching information from the documentation, hence my suggestion to pre-process the information on the server.

For a general-purpose "copy to clipboard and paste in chatbot prompt" usage, I'm sure that the current approach fits very well. It might be perfect for power users of other technical documentation, but our readers are primarily functional people and not that tech-savvy.

WDYT @xpl-odoo @auva-odoo?

mattismegevand:
I understand your point, potentially we could limit this feature to the dev docs only?
So that we don't confuse non-tech people.

xpl-odoo:
Currently, I doubt it would be helpful for the average reader of the [user](https://www.odoo.com/documentation/18.0/applications.html#)'s part of the documentation; the [help chatbot](https://www.odoo.com/help) should be enough in most cases.

AntoineVDV:
> I understand your point, potentially we could limit this feature to the dev docs only? So that we don't confuse non-tech people.

To be honest, I doubt this feature's usefulness would outweigh its maintenance cost in our dev docs. The information is scattered across a bazillion pages that all large LLMs have already digested thousands of times anyway. The per-page export and manually feeding limitations make me believe she should not go forward with this.

mattismegevand:
Ok closing the PR then

PR #14070, opened 2025-07-17
[IMP] misc: magic sheets google drive links

xpl-odoo:
Magic sheets would be better managed using Google Drive instead of using the documentation repository.


xpl-odoo:
@robodoo r+

PR #14091, opened 2025-07-18
[ADD] reference/external_json_api

Julien00859:


Julien00859:
> I think it would be better to move the RPC-related content to a new file and reuse the very well-referenced external_api.rst file for the new external API instead. Then, in 19.1, the file about deprecated-now-unsupported content can be removed.

Done

> This is also because we don't document when this or that feature was added, deprecated, or removed. The documentation for version X.Y documents which features exist in Odoo X.Y, no more no less.

That we are dropping XML/JSON-RPC is serious enough to mandate the *deprecation* danger. A common question I've been asked is "is the JSON-2 API available in 18?", hence the "new in 19.0" passage.

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
+================
+External RPC API
+================
+
+.. danger::
+
+   .. deprecated:: 19.0
+
+   Both the XML-RPC and JSON-RPC APIs are scheduled for removal in Odoo 20.0 (fall 2026). The
+   :doc:`external_api` acts as replacement.
```
Flotchet: :doc:`external_api` acts as **a** replacement.

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
...
+this key, you will have to create a new one (and probably delete the one you
+lost).
+
+Once you have keys configured on your account, they will appear above the
+:guilabel:`New API Key` button, and you will be able to delete them:
+
+.. image:: external_api/delete-key.png
+   :align: center
+
+**A deleted API key can not be undeleted or re-set**. You will have to generate
```
Flotchet, line 138: A deleted API key **cannot** be undeleted or re-set
**nitpick:**  A deleted API key **cannot** be **recovered or reset**.

```suggestion
**A deleted API key cannot be recovered or reset**. You will have to generate
```

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
...
+
+Logging in
+----------
+
+Odoo requires users of the API to be authenticated before they can query most
+data.
+
+The ``xmlrpc/2/common`` endpoint provides meta-calls which don't require
+authentication, such as the authentication itself or fetching version
+information. To verify if the connection information is correct before trying
```
Flotchet, line 232: **nitpick:** To verify **whether** the connection information is correct
```suggestion
information. To verify whether the connection information is correct before trying
```

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
...
+         }
+
+   Result:
+
+   .. code-block:: json
+
+      78
+
+.. warning::
+   While most value types are what would expect (integer for
```
Flotchet, line 932: While most value types are what **you** would expect (integer for

```suggestion
   While most value types are what you would expect (integer for
```

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
...
+
+.. danger::
+
+   .. deprecated:: 19.0
+
+   Both the XML-RPC and JSON-RPC APIs are scheduled for removal in Odoo 20.0 (fall 2026). The
+   :doc:`external_api` acts as replacement.
+
+Odoo is usually extended internally via modules, but many of its features and
+all of its data are also available from the outside for external analysis or
```
Flotchet, line 16: maybe:  available externally for analysis or integration

```suggestion
all of its data are also available externally for analysis or
```

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
...
+integration with various tools. Part of the :ref:`reference/orm/model` API is
+easily available over XML-RPC_ and accessible from a variety of languages.
+
+Starting with PHP8, the XML-RPC extension may not be available by default.
+Check out the `manual <https://www.php.net/manual/en/xmlrpc.installation.php>`_
```
Flotchet, line 21: **nitpick**: I would more use something like "Refer to"

```suggestion
Refer to the `manual <https://www.php.net/manual/en/xmlrpc.installation.php>`_
```

content/developer/reference/external_rpc_api.rst
```diff
@@ -0,0 +1,1464 @@
...
+
+Odoo requires users of the API to be authenticated before they can query most
+data.
+
+The ``xmlrpc/2/common`` endpoint provides meta-calls which don't require
+authentication, such as the authentication itself or fetching version
+information. To verify if the connection information is correct before trying
+to authenticate, the simplest call is to ask for the server's version. The
+authentication itself is done through the ``authenticate`` function and
+returns a user identifier (``uid``) used in authenticated calls instead of
```
Flotchet, line 235: _**real big o nitpick here**_ : **which returns** instead of **and returns** to empathize the fact that the function is the one returning the uid

```suggestion
authentication itself is done through the ``authenticate`` function which
returns a user identifier (``uid``) used in authenticated calls instead of
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-                 asList("is_company", "=", true))),
-             new HashMap() {{ put("offset", 10); put("limit", 5); }}
-         )));
+     - .. image:: external_api/new-api-key.png
+          :align: center
 
-      .. code-tab:: go
+A description and a duration are needed to create a new api key. The description makes it possible
+to identify the key, and to determine later whether the key is still in use or should be removed.
+The duration determines the lifetime of the key after which the the key becomes invalid. It is
```
Flotchet: the the key

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
+   consulted on their ``/doc`` page.
 
-      .. note::
-         These examples use the `Apache XML-RPC library <https://ws.apache.org/xmlrpc/>`_.
+.. note::
 
-         The examples do not include imports as these imports couldn't be
-         pasted in the code.
+   Access to data via the external API is only available on *Custom* Odoo pricing plans. Access to
+   qthe external API is not available on *One App Free* or *Standard* plans. For more information
```
Flotchet: **qthe**

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-integration with various tools. Part of the :ref:`reference/orm/model` API is
-easily available over XML-RPC_ and accessible from a variety of languages.
+.. versionadded:: 19.0
 
-.. important::
-   Starting with PHP8, the XML-RPC extension may not be available by default.
-   Check out the `manual <https://www.php.net/manual/en/xmlrpc.installation.php>`_
-   for the installation steps.
+Odoo is usually extended internally via modules, but many of its features and all of its data are
+also available from the outside for external analysis or integration with various other softwares.
```
Flotchet: maybe: available externally for analysis or integration (same as the other file https://github.com/odoo/documentation/pull/14091/files#r2283296859)

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-         )));
+     - .. image:: external_api/new-api-key.png
+          :align: center
 
-      .. code-tab:: go
+A description and a duration are needed to create a new api key. The description makes it possible
+to identify the key, and to determine later whether the key is still in use or should be removed.
+The duration determines the lifetime of the key after which the the key becomes invalid. It is
+recommended to set a short duration (typically 1 day) for interactive usage. It is not possible to
+create keys that last for more than 3 months, it means that long lasting keys must be rotated at
```
Flotchet: nitpick **months, it means**  should be **months. It means** imo

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-             db, uid, password,
-             "res.partner", "search",
-             []any{[]any{
-                 []any{"is_company", "=", true},
-             }},
-             map[string]int64{"offset": 10, "limit":  5},
-         }, &records); err != nil {
-             log.Fatal(err)
-         }
+The :guilabel:`Generate Key` button creates a 160 bits strong random key. Its value appears on
```
Flotchet: **160 bits strong random key** -> **a strong 160-bit random key**

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-         }, &records); err != nil {
-             log.Fatal(err)
-         }
-         // count the number of fields fetched by default
-         count := len(records)
+The JSON-2 API uses the standard :ref:`security <reference/security>` model of Odoo. All operations
+are validated against the access rights, record rules and field accesses of the user.
 
-   Result:
+For **interfactive usage**, such as discovering the API or running one-time scripts, it is fine to
```
Flotchet: interfactive -> interactive

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
+create and use **dedicated bot users**.
 
-      121
+Using dedicated bot users has several benefits:
 
-   Conversely, picking only three fields deemed interesting.
+* The minimum required permissions can be granted to the bot, limiting the impact may the API key
+  gets compromised;
+* The password can be set empty to disable login/password authentication, limiting the likelihood
+  the account gets compromized;
```
Flotchet: compromized -> compromised

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
+* The :ref:`reference/fields/automatic/log_access` use the bot account. No user gets impersonalized.
```
Flotchet: gets impersonalized -> is impersonated

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
 
-   .. tabs::
 
-      .. code-tab:: python
+Database
+--------
 
-         models.execute_kw(db, uid, password, 'res.partner', 'read', [ids], {'fields': ['name', 'country_id', 'comment']})
+Depending on the deploiement, the ``Host`` and/or ``X-Odoo-Database`` request headers might be
```
Flotchet: deploiement -> deployment

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
+required. The ``Host`` header is required by HTTP/1.1 and is needed on servers where Odoo is
+installed next to other web applications, so a web-server/reverse-proxy is able to route the request
+to the Odoo server. The ``X-Odoo-Database`` header is required when a single Odoo server hosts
+multiple databases, and that :ref:`dbfilter` wasn't configured to use the ``Host`` header.
 
-      .. code-tab:: ruby
+Most HTTP client libraries automatically set the ``Host`` header using the connection url.
```
Flotchet: nitpick ad: url -> URL

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-             []any{[]any{
-                 []any{"is_company", "=", true},
-             }},
-             map[string]int64{"offset": 10, "limit":  5},
-         }, &records); err != nil {
-             log.Fatal(err)
-         }
+The :guilabel:`Generate Key` button creates a 160 bits strong random key. Its value appears on
+screen, this is the only time and place the key is visible on screen. It must be copied, kept secret
+and stored somewhere secure. If it ever gets compromized or lost, then it must be removed.
```
Flotchet: compromized -> compromised

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-   .. code-block:: json
+For **extended automated usage**, such as an integration with another software, it is recommended to
+create and use **dedicated bot users**.
 
-      121
+Using dedicated bot users has several benefits:
 
-   Conversely, picking only three fields deemed interesting.
+* The minimum required permissions can be granted to the bot, limiting the impact may the API key
+  gets compromised;
```
Flotchet: actually not 100% sure on this one
**may the API key gets compromised** should be **if the API key gets compromised**

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
-             asList(asList(
-                 asList("is_company", "=", true))),
-             new HashMap() {{
-                 put("fields", asList("name", "country_id", "comment"));
-                 put("limit", 5);
-             }}
-         )));
+The following examples showcase how to execute two of the :ref:`reference/orm/models/crud` on a fake
+database ``mycompany`` hosted on a fake website ``https://mycompany.example.com``. Its comprehensive
+documentation would be available at https://mycompany.example.com/doc
```
Flotchet: **would be available** is a bit speculative and makes it feel like we are not sure imo

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,348 @@
...
+
+Migrating from XML-RPC / JSON-RPC
+=================================
+
+.. note::
 
-.. _PostgreSQL: https://www.postgresql.org
-.. _XML-RPC: https://en.wikipedia.org/wiki/XML-RPC
-.. _base64: https://en.wikipedia.org/wiki/Base64
+   Under construction
```
Flotchet: ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣀⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⢀⣠⣴⠾⢻⣿⡟⠻⠶⢦⣤⣀⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⣀⣤⠾⠛⠉⠀⠀⣸⠛⣷⠀⠀⠀⠀⠉⠙⠻⠶⣦⣤⣀⠀⠀⠀⠀⠀
⠀⠀⠐⠛⠋⠀⠀⠀⠀⠀⠀⠛⠀⠛⠂⠀⠀⠀⠀⠀⠀⠀⠀⠈⠙⠛⠒⠂⠀⠀
⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡇⠀
⠀⠀⠀⢠⣤⣤⣤⠀⠀⠀⠀⢠⣤⡄⢠⣤⡄⠀⠀⠀⠀⠀⠀⠀⠀⠀⡄⠀⠀⠀
⠀⠀⠀⠈⠉⠉⠉⠀⠀⠀⠀⠸⠿⠇⠸⠿⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢰⣶⡆⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠃⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡀⠃⡀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⣿⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⠋⠈⠛⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⢸⣿⡇⢀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⡇⠀
⠀⠀⠀⠀⠀⠀⠀⢀⣴⠾⠋⢸⣿⡇⠈⠳⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠈⠁⠀⠀⠈⠛⠃⠀⠀⠀⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀

Flotchet (changes requested):

content/developer/reference/external_api.rst
Julien00859: I don't understand, can you do a \`\`\`suggestion instead?

content/developer/reference/external_api.rst
Julien00859: am not sure either, let's wait a 3rd opinion :)

content/developer/reference/external_api.rst
Julien00859: am not sure either, let's wait a 3rd opinion :)

content/developer/reference/external_api.rst
Flotchet: It's more about the flow here for me it feel like 2 separate sentences

```suggestion
create keys that last for more than 3 months. It means that long lasting keys must be rotated at
```

content/developer/reference/external_api.rst
rco-odoo: +1 on Florent's suggestion.
Julien's version is correct (except "may ... **gets** compromised" → "may ... **get** compromised") but Florent's is more straightforward to understand.

content/developer/reference/external_api.rst
rco-odoo: 3rd opinion says: "would" clearly suggests "no sure".
I suggest "**is** available at" or "**should be** available at".

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
+**Body JSON object**
 
-.. versionadded:: 14.0
+:ids: An array of record ids on which to execute the method.
+:context: Optional, an object of additional values. e.g. ``{"lang": "en_US"}``.
+:*param*: As many time as needed, a value for the method's *param* parameter.
 
-Odoo has support for **api keys** and (depending on modules or settings) may
-**require** these keys to perform webservice operations.
+The ``ids`` entry must be empty when calling a ``@api.model``-decorated method.
```
rco-odoo: This is not correct.  When calling an `@api.model` method, `ids` _should_ be empty or simply not given.  (In practice, it _may_ be nonempty.  By convention, the method shouldn't depend on `self`'s record ids.)

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
-.. _api/external_api/keys:
+:model: Required, the technical model name.
+:method: Required, the method to execute.
 
-API Keys
-~~~~~~~~
+**Body JSON object**
 
-.. versionadded:: 14.0
+:ids: An array of record ids on which to execute the method.
```
rco-odoo, line 46: Please add that `ids` is optional.  When not given, an empty list is used instead.

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
-         final XmlRpcClientConfigImpl start_config = new XmlRpcClientConfigImpl();
-         start_config.setServerURL(new URL("https://demo.odoo.com/start"));
-         final Map<String, String> info = (Map<String, String>)client.execute(
-             start_config, "start", emptyList());
+A description and a duration are needed to create a new api key. The description makes it possible
+to identify the key, and to determine later whether the key is still in use or should be removed.
+The duration determines the lifetime of the key after which the key becomes invalid. It is
+recommended to set a short duration (typically 1 day) for interactive usage. It is not possible to
+create keys that last for more than 3 months, it means that long lasting keys must be rotated at
+least once every 3 months.
```
rco-odoo:
```suggestion
Both a description and a duration are needed to create a new API key. The description makes it possible
to identify the key, and to determine later whether the key is still in use or should be removed.
The duration determines the lifetime of the key after, which the key becomes invalid. It is
recommended to set a short duration (typically 1 day) for interactive usage. For security reasons,
it is not possible to
create keys that last for more than 3 months. This means that long lasting keys must be rotated at
least once every 3 months.
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
 
-         final String url = info.get("host"),
-                       db = info.get("database"),
-                 username = info.get("user"),
-                 password = info.get("password");
+The :guilabel:`Generate Key` button creates a strong 160-bits random key. Its value appears on
+screen, this is the only time and place the key is visible on screen. It must be copied, kept secret
+and stored somewhere secure. If it ever gets compromised or lost, then it must be removed.
```
rco-odoo, line 163:
```suggestion
and stored in a secure place. If it ever gets compromised or lost, then it must be removed.
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
-Logging in
-----------
+Using dedicated bot users has several benefits:
 
-Odoo requires users of the API to be authenticated before they can query most
-data.
+* The minimum required permissions can be granted to the bot, limiting the impact may the API key
+  gets compromised;
+* The password can be set empty to disable login/password authentication, limiting the likelihood
+  the account gets compromised;
```
rco-odoo:
```suggestion
* The minimum required permissions can be granted to the bot, limiting the impact if the API key
  gets compromised.
* The password can be set empty to disable login/password authentication, limiting the likelihood
  of the account getting compromised.
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
+Transaction
+===========
 
-   .. code-tab:: php
+All calls to the JSON-2 endpoint run in their own SQL transactions. The transaction are committed in
+cases of success, and are rollbacked in case of error. Using the JSON-2 API, it is not possible to
+chain multiple calls inside a single transaction. It means that one must be cautious when doing
+multiple consecutive calls as the data might be modified/removed by other concurrent transactions
+and corrupt the data. This is especially dangerous when doing operations related to reservations,
+paiements, and the such.
```
rco-odoo:
```suggestion
All calls to the JSON-2 endpoint run in their own SQL transaction. The transaction is committed in
case of success, and discarded in case of error. Using the JSON-2 API, it is not possible to
chain multiple calls inside a single transaction. It means that one must be cautious when doing
multiple consecutive calls, as the database might be modified by concurrent transactions.
This is especially dangerous when doing operations related to reservations,
payments, and such.
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
 
-      $common = ripcord::client("$url/xmlrpc/2/common");
-      $common->version();
+The solution is to always call a single method that does all the related operations in a same
+transaction. This way the data is guaranteed to stays consistent: either everything is done
+(success, commit), either nothing is done (error, rollback).
```
rco-odoo:
```suggestion
The solution is to always call a single method that does all the related operations in a single
transaction. This way, the data is guaranteed to stay consistent: either everything is done
(success, commit), or nothing is done (error, rollback).
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
 
-   .. code-tab:: java
+In the ORM, the ``search_read`` method is one example of a single method that does multiple
+operations (``search`` then ``read``) in a single transaction. If another concurrent request is
+removing one of the records ``search`` is finding, then there’s a risk the subsequent call to
+``read`` fails for a missing record error. Such problem cannot occurs in ``search_read`` as that the
+system guarantees a proper isolation in between transactions.
```
rco-odoo:
```suggestion
In the ORM, the ``search_read`` method is an example of a single method that does multiple
operations (``search`` then ``read``) in a single transaction. If a concurrent request
removes one of the records ``search`` retrieves, then there is a risk that subsequent calls to
``read`` fail for a missing record error. Such problem cannot occur in ``search_read``, as the
system guarantees proper isolation between transactions.
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
 
-      final XmlRpcClientConfigImpl common_config = new XmlRpcClientConfigImpl();
-      common_config.setServerURL(new URL(String.format("%s/xmlrpc/2/common", url)));
-      client.execute(common_config, "version", emptyList());
+In business models those methods are often prefixed by ``action_``, such as
+``sale.order/action_confirm`` that verifies that a sale order is valid before confirming it.
```
rco-odoo:
```suggestion
In business models those methods are often prefixed by ``action_``, such as
``sale.order/action_confirm``, which verifies that a sale order is valid before confirming it.
```

content/developer/reference/external_api.rst
rco-odoo: Given the whole context, I suggest using **is**.

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
+          -X POST \
+          --oauth2-bearer $API_KEY \
+          -H "X-Odoo-Database: $DATABASE" \
+          -H "Content-Type: application/json" \
+          -d "{\"ids\": $ids, \"context\": {\"lang\": \"en_US\"}, \"fields\": [\"name\"]}" \
+          --silent \
+          --fail-with-body
+
+
+The above example would be equivalent to running::
```
rco-odoo:
```suggestion
The above example is equivalent to running::
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
+
+   Model = self.env["res.partner"].with_context({"lang": "en_US"})
+   records = Model.search([("name", "ilike", "%deco%"), ("is_company", "=", True)])
+   return json.dumps(records.ids)
+
+Then in a new transaction::
```
rco-odoo: It's the continuation of the sentence above:
```suggestion
then in a new transaction::
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
-          },
-          {
-              "comment": false,
-              "country_id": [ 76, "France" ],
-              "id": 17,
-              "name": "Camptocamp"
-          }
-      ]
+The remaining function are: ``server_version``, which exists under ``/web/version``; ``list_lang``
+and ``list_countries`` which exist via JSON-2 on the ``res.lang`` and ``res.country`` models; and
```
rco-odoo, line 463:
```suggestion
and ``list_countries`` which exists via JSON-2 on the ``res.lang`` and ``res.country`` models; and
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
-to initialize the record. For any field which has a default value and is not
-set through the mapping argument, the default value will be used.
+The object service defines 2 fonctions:
 
-.. example::
+#. ``execute(db, uid, passwd, model, method, *args)``
+#. ``execute_kw(db, uid, passwd, model, method, args, kw={})``
 
-   .. tabs::
+They both allow for access to all public model methods, including the generic ORM ones.
```
rco-odoo:
```suggestion
They both give access to all public model methods, including the generic ORM ones.
```

content/developer/reference/external_api.rst
```diff
@@ -1,1462 +1,541 @@
...
 
-      .. code-tab:: python
+Both functions are stateless. It means that the database, user id and user password are to be
+provided for each call. The model, method are arguments must likewas be provided. The ``execute``
```
rco-odoo:
```suggestion
Both functions are sessionless. It means that the database, user id and user password are to be
provided for each call. The model and method arguments must be provided, too. The ``execute``
```

content/developer/reference/external_api.rst
Julien00859: It is correct, json2 has the following code:

```py
if hasattr(func, '_api_model') and ids:
    e = f"cannot call {model}.{method} with ids"
    raise UnprocessableEntity(e)
```

content/developer/reference/external_api.rst
Julien00859: changed for:

> model, method and arguments must be provided, too.

rco-odoo (approved):
Good to me.

jakubsmolka:
Would it be possible NOT to rely on OEXP and OCA github communication and send this information directly as a newsletter/through account and partner managers to Odoo partners and customers?

It is big and important enough IMHO to be communicated directly to all necessary stakeholders. The fact that Fabien was considering to completely remove the old API for V19 with basically just 2 months in advance was scary enough, but I believe it is good not to underestimate the time it can take various parties to align themselves with this change - and the only reliable source of communication where Odoo can say "we did our best to reliable inform everyone in time" are the ones mentioned above

Julien00859:
Hello @jakubsmolka 

I like your idea of pro-actively reaching with the stakeholders, I'll try to push the idea at work.

Julien00859:
@jakubsmolka 

We met with Antony about this matter. Our communication will only take place during OXP, we won't use other communication channels, emails to stakeholders have been explicitly ruled out.

jakubsmolka:
> @jakubsmolka
> 
> We met with Antony about this matter. Our communication will only take place during OXP, we won't use other communication channels, emails to stakeholders have been explicitly ruled out.

OK, thasnk for the info and confirmation. I do wonder in a year when Odoo 20 comes out, how many parties will be surprised and how many external platforms with Odoo integrations will "suddenly" stop working for new customers or after upgrades...  but if that's a situation you are fine with occuring, I guess there is nothing else we can do. (and I know it will always happen to a degree that someone will be surprised, but it feels like not doing everything reasonably possible to communicate it properly.... is just inviting to make it worse)

Julien00859:
Hello @odoo/documents-doc-review we are ready

AntoineVDV:
@Julien00859 As discussed, we won't have time to review this until after the freeze. You said it only had to be merged for OXP but if you need the odoo/odoo PR to be merged ASAP, I recommend closing the documentation PR to let the other one go through ;)

rco-odoo:
Let the odoo/odoo PR go through.

Julien00859:
Reopening now that the community PR is merged

AntoineVDV:
Woops, that's new :open_mouth: 
image

You can open a new PR, I will check it now.

Or ask runbot if it's really necessary to prevent this.

PR #14153, opened 2025-07-25
[I18N] contributing, developer/tutorial: remove Transifex references

ticodoo:
Odoo is transitioning away from Transifex. Remove links/references to it and replace with new platform (internal Weblate server) links


ticodoo:
@robodoo r+

PR #14193, opened 2025-07-31
[FIX] developer: wrong import path for Form module in tests

Megaaaaaa:
Since https://github.com/odoo/odoo/pull/116779/commits/f5e7c67892cb7f096a08869e085ea1b5c546ed3d, Form is in odoo.tests and not in odoo.tests.common anymore

task-none


AntoineVDV (approved):
@robodoo r+

PR #14194, opened 2025-07-31
[FIX] database management: iOS app note

auva-odoo:
The iOS app can technically be udpdated, but might not be in the future.


auva-odoo:
@robodoo r+

PR #14223, opened 2025-08-04
Update testing.rst

xafr-odoo:
The current instructions generate errors when attempting to load the tour.

`module_loader.js:165 The following modules are needed by other modules but have not been defined, they may not be present in the correct asset bundle:  (3) ['web_tour.tour', 'web.core', 'web.utils']`

Replacing by `web.assets_frontend` fixes the issue.


xafr-odoo:
This seems that it was incorrect and more an issue on my end. I will update with a fix if I find one.

PR #14257, opened 2025-08-08
[REM] accounting: account_payment module install

auva-odoo:
Removing this doc as installing this module/patch is no longer necessary to disable online invoice payments


auva-odoo:
@robodoo r+

PR #14276, opened 2025-08-12
[IMP] website: change plausible warning

lman-odoo:
task-4853660

### DIFF
--- a/content/applications/websites/website/reporting/analytics.rst
+++ b/content/applications/websites/website/reporting/analytics.rst
@@ -45,9 +45,9 @@ Analytics`.
 
    .. warning::
       Deactivating the free Plausible.io account linked to your **Odoo Online** database
-      will also invalidate the existing keys, making them unusable. As a result, new
-      keys will be generated, and all historical data linked to the old keys will be lost. If you
-      plan to deactivate the account, it is recommended to back up your data locally beforehand.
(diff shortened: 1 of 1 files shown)


content/applications/websites/website/reporting/analytics.rst
```diff
@@ -45,9 +45,9 @@ Analytics`.
 
    .. warning::
       Deactivating the free Plausible.io account linked to your **Odoo Online** database
-      will also invalidate the existing keys, making them unusable. As a result, new
-      keys will be generated, and all historical data linked to the old keys will be lost. If you
-      plan to deactivate the account, it is recommended to back up your data locally beforehand.
+      will also remove the existing keys. As a result, new keys will be generated, and all
+      historical data is only stored in the old keys. If you plan to deactivate the account, it is
+      recommended to save the existing keys.
```
auva-odoo:
```suggestion
      will also remove the existing keys. As a result, new keys will be generated, while all
      historical data will remain associated with the old keys. If you plan to deactivate the account, it is
      recommended to save the existing keys to preserve access to that data.
```
- *while* instead of *and*: for contrasts (instead of joining these two facts)
- *associated* instead of *stored*: more accurate and less technical

auva-odoo (approved):
Thank you @lman-odoo ! 

@robodoo r+

PR #14359, opened 2025-08-20
[FIX] essentials: favorited search update correction

mcsi-odoo:
task-5005364


mcsi-odoo:
@robodoo fw=no

auva-odoo (approved):
@robodoo r+

PR #14396, opened 2025-08-25
[FIX] fiscal_localizations: change iot Mac to identifier

loouis-t:
The new identifier for IoT Boxes it the serial number (or motherboard uuid, for Virtual IoT Boxes), we then need to update the documentation to lead to IoT Boxes identifier instead of not existing anymore mac address.

Task: 5001274


xpl-odoo (approved):
@robodoo r+

PR #14403, opened 2025-08-25
[FIX] doc: remove extra "to" in "try adding"

ASPhilemon:


Feyensv (approved):
@robodoo r+ squash

Thanks for the rectification !

PR #14433, opened 2025-08-28
17.0 razorpay connect auva

auva-odoo:


content/applications/finance/payment_providers/razorpay.rst
```diff
@@ -2,53 +2,86 @@
 Razorpay
 ========
 
-`Razorpay <https://razorpay.com/>`_ is an online payments provider established in India and
-covering more than 100 payment methods.
+`Razorpay <https://razorpay.com/>`_ is an online payments provider established in India thatsupports
```
kcv-odoo:
```suggestion
`Razorpay <https://razorpay.com/>`_ is an online payments provider established in India that supports
```

content/applications/finance/payment_providers/razorpay.rst
```diff
@@ -2,53 +2,86 @@
...
+more than 100 payment methods.
+
+.. _payment_providers/razorpay/oauth-connection:
+
+Create a Razorpay account with Odoo (Indian companies only)
```
kcv-odoo, line 10: IMO Indian companies only is quite misleading here because any company can connect with Razorpay Oauth they just need to have Indian account not (Malaysia or Singapore) as there are currently only 3 country supported

So it's like US company doing business in India can use Oauth but they have to create Indian account with US details or something like that only Other counties account (Malasia and Singapore) will need use key/secret

Also we are promoting Oauth and prefer Oauth over key/secret and wanted to migrate old user to Oauth so I'll would be better to highlight that and having this `Indian companies only` might scare customer to use it 🤔 
(Reason we are promoting Oauth is key/sercet is old and might discontinue in future as Razorpay team was saying)

content/applications/finance/payment_providers/razorpay.rst
```diff
@@ -2,53 +2,86 @@
...
+===========================================================
+
+.. note::
+   - This method is only available for Indian companies.
+   - If you want to :ref:`try Razorpay as a test <payment_providers/test-mode>` first, use the
+     :ref:`manual credentials and webhook input method <payment_providers/razorpay/manual-connection>`.
```
kcv-odoo: I am not sure this note is quite useful instead we could just give warning `Connect flow does not support test mode` (or don't add anything at all) because we don't want user to goto lengthy onboarding process and for first time user this is not very useful and user can simply test it on like account by making 1INR payment

Still don't have very strong opinion on that. may be @AntoineVDV can help us here? 🙈

content/applications/finance/payment_providers/razorpay.rst
auva-odoo: Right, but when you create a razorpay account, you have to provide an address, correct? and that address would need to be an address in India for the account to be an Indian account, correct? and you also need an Indian phone number to create a Razorpay account. So how would a company outside India go about creating a Razorpay account?

content/applications/finance/payment_providers/razorpay.rst
auva-odoo: Yes @AntoineVDV I'd like your opinion on this when you have a moment please :pray:

AntoineVDV: Yeah, I agree with @kcv-odoo. I like the suggested formulation because you can still infer the information from the warning and achieve the configuration of your provider for test mode if you're tech-savvy and motivated.

content/applications/finance/payment_providers/razorpay.rst
kcv-odoo: Yes user will only able to connect through Oauth with Indian account only so I guess it's okay!

auva-odoo:
@robodoo r+

PR #14466, opened 2025-09-02
[IMP] windows_iot: link to nightly for installer

orma-odoo:
Old installer is not stable with 18.4+ databases. Changing the link to redirect to nightly build. For now to the one of 19-08 because latest versions are not working, will be changed when it's fixed or when Odoo 19 installer will be on official download page.

### DIFF
--- a/content/applications/general/iot/windows_iot.rst
+++ b/content/applications/general/iot/windows_iot.rst
@@ -34,14 +34,11 @@ Installation
 
 To install the Windows virtual IoT on a Windows computer:
 
-#. Access `Odoo's download page <https://odoo.com/download>`_ and download the Odoo installation
-   package for Windows **matching your database's version**.
+#. Download the `IoT installation package for Windows
+   <https://nightly.odoo.com/master/nightly/iot/odoo_iot_setup_18.5a1.20250819.exe>`_
 #. Open the downloaded :file:`.exe` file, allow the app to make changes to your device, select
    a language, and click :guilabel:`OK`.
 #. Click :guilabel:`Next`, then :guilabel:`I Agree` to accept the terms and conditions and continue.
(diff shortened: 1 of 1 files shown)


auva-odoo:
Hello @ZorbMax and thank you for this PR! 

- Sorry if this is a dumb question but the updates you made are valid for 18.4+ databases, but what about older version? Will that work for them too? Is there a more "universal" link we could add so this can be valid for everyone? (unless I'm mistaken and it *will* work for every Odoo version?)
- I suppose you removed steps because they aren't necessary with this setup? sorry I unfortunately can't test this for myself so I have to ask :see_no_evil:

orma-odoo:
Hello @auva-odoo :)

No problem, happy to explain the situation ;)

R&D recently changed how the windows IoT works in order to fix several compatibility issues we faced when trying to use it on intermediate versions (18.1, 18.2, etc). To explain in short, before windows IoT was in version 18 even if you paired it with 18.2 for example. Since 18.4, they changed this in order to switch the windows IoT code in the same version as the database connected to it (this is how the classic IoT box is doing since forever I think)

The new link I'm adding is to redirect the customers to this new version and improve their experience using Odoo compared to have to wait a reply from the Support when they face a bug and then we redirect them to this version.

I'm targeting saas-18.4 branch in order to redirect only recent databases as this installer won't work properly on older databases unfortunatly from what R&D told me. It's a temporary change because as soon as the installer for Odoo 19 will be on the download page we will be able to go back to the same documentation as before in theory (link to official download page). But I'm not sure when this installer will be available to be fair (I can ask if it's not really wanted to change documentation too often)

For the steps I removed/edited, yes it's because the installer on nightly is a bit different from the official one so I changed those to be more accurate to the real flow during installation.

Thank you for your feedback :D

PR #14471, opened 2025-09-02
[I18N] *: remove .tx/config

dylankiss:
After the switch to Weblate, we now only use the `.weblate.json` file instead of the previous `.tx/config` file.


ticodoo:
@robodoo r+

PR #14477, opened 2025-09-03
[FIX] developer/modules: init/uninstall hooks arguments

chienandalu:
Since https://github.com/odoo/odoo/pull/108254 an env is used directly in the init/uninstall hooks.

cc @moduon

fyi @Xino61122


chienandalu:
A simple doc fix. Could you take a look @ryv-odoo @rco-odoo ?

ryv-odoo:
Makes perfect sense. Thx @chienandalu 
@robodoo r+

AntoineVDV (approved):
@robodoo r+

PR #14501, opened 2025-09-04
[ADD] Payroll: Time Off To Report

larm-odoo:
Adding new doc for this section that was previously in the Work Entries doc.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5066970) for this PR.


larm-odoo:
Hi @jero-odoo - this is ready for a peer review. Thanks!

content/applications/hr/payroll/time_off_to_report.rst
```diff
@@ -0,0 +1,77 @@
+==================
+Time off to report
+==================
+
+It is not uncommon for payroll to be processed a few days before the end of a pay period, so that
```
jero-odoo:
```suggestion
Payroll is often processed a few days before the end of a pay period, so that
```

content/applications/hr/payroll/time_off_to_report.rst
```diff
@@ -0,0 +1,77 @@
...
+a :guilabel:`Payslip State` of :guilabel:`To defer to next payslip`. This is because the employee
+was *already paid* for that time as worked time, and it was logged as regular time spent at work.
+
+.. image:: time_off_to_report/time-off-to-report.png
+   :alt: A list of all time off requests that were not approved before payslips were generated.
+
+Defer multiple time off entries
+===============================
+
+In order to keep the employee's time off balances correct, the time off request **must** be applied
```
jero-odoo, line 45:
```suggestion
To keep the employee's time off balances correct, the time off request **must** be applied
```

larm-odoo:
Hi @Felicious - this is ready for a final review!

content/applications/hr/payroll/time_off_to_report.rst
```diff
@@ -0,0 +1,77 @@
+==================
+Time off to report
+==================
```
Felicious, line 3: Hi Lara! More lengthy of a comment here (: 

Since this is a shorter doc, I'm wondering if you'd like to experiment with our new action-focused document type and rewrite this in a new style. Currently, the doc is clear in explaining the dashboard, context, and examples, but if this document is meant to be read for only one reason (resolve payroll discrepancies), we can shift this doc to be more about what the user needs to do and why.

So instead of titling this "Tiome off to report", for instance, we can title it something like "Resolve time off conflicts". 

In addition, I have some questions for you to consider to help reframe this doc in a more action-oriented way:
- Who is the intended user for this doc (payroll administrators, HR officers, or general employees)?
- What is the specific goal this user is trying to accomplish when they open this page?
- At what point in their workflow would the payroll admin land here? Can we structure the doc so they can get from “problem identified” → “steps to fix it” quickly?
- Can we reframe the introduction: instead of explaining payroll theory, can we start with the problem the user faces (e.g., “An employee’s time off was approved after payroll was processed—what do you do?”)?
- What is the main decision point for the user—defer multiple entries at once, or handle them one by one?
- Would it help to start each section title with a verb (e.g., “Defer multiple time off entries” → “Defer multiple entries at once”)?

Felicious: To give you a sense of the vibe I'm going for, itl'l be something like this:

### Defer multiple time off entries

When employees request time off after payroll has already been processed, those entries must be deferred to the next pay period.
1. Open :menuselection:`Payroll app --> Work Entries --> Time Off to Report`.
2. Select the time off requests to defer:
  - To select individual requests, check the box next to each line.
  - To select all requests, check the box at the top of the :guilabel:`Employee` column.
3. At the top of the list, click :menuselection:`Actions --> Defer to Next Month`.
4. The selected entries are deferred to the following pay period.

Deferring multiple time off entries ensures time off balances remain accurate and eliminates the need to reissue paychecks.

Felicious (approved):
@robodoo r+

PR #14515, opened 2025-09-08
[IMP] Discuss: Canned responses update

seb-odoo:
Previously, the documentation described canned responses with a description field, no explicit shortcut indicator, and included default filters in the view.

Now, the description field has been removed, and the UI displays `::` as the shortcut prefix, and default filters have been removed from the view.


AntoineVDV (approved):
@robodoo r+

PR #14516, opened 2025-09-08
[IMP] Payroll: update work entries report

larm-odoo:
Updating to new standards


larm-odoo:
This doc is current and needs to updating.

PR #14567, opened 2025-09-15
[IMP]: Update doc to match current Odoo 18.4

Felicious:
A lot of updates happened during the life of 18x versions and the doc needs to be updated to reflect that.


Felicious:
@robodoo r+

PR #14572, opened 2025-09-16
[IMP] inventory: mto feedback

Felicious:
Address feedback that experts reported (:

- Fix inaccuracy that MOs and RFQs are triggered all the time --> they're triggered only when we don't have enough in stock
  - [Feedback](https://www.odoo.com/odoo/project/3835/tasks/4338750)
- Add warning that cancelling SOs do not automatically cancel linked RFQs and MOs
  - [Feedback](https://www.odoo.com/odoo/project/3835/tasks/5086397)

### DIFF
--- a/content/applications/inventory_and_mrp/inventory/warehouses_storage/replenishment/mto.rst
+++ b/content/applications/inventory_and_mrp/inventory/warehouses_storage/replenishment/mto.rst
@@ -11,13 +11,19 @@ Replenish on order (MTO)
 .. |BOM| replace:: :abbr:`BOM (bill of materials)`
 
 *Replenish on order*, also known as *MTO* (make to order), is a replenishment strategy that creates
-a draft order for a product every time it is required to fulfill a sales order (SO), or when it is
-needed as a component in a manufacturing order (MO).
+a draft order every time a product is needed to fulfill a sales order (SO) or as a component in a
+manufacturing order (MO).
 
-For products that are purchased from a vendor, a request for quotation (RFQ) is created to replenish
-the product, while an |MO| is created for products that are manufactured. The creation of an |RFQ|
-or |MO| occurs every time an |SO| or |MO| that requires the product is confirmed, regardless of the
-current stock level of the product being ordered.
+- For :doc:`purchased products <../../../purchase/manage_deals/rfq>`, Odoo creates a |RFQ|
+- For :doc:`manufactured products
+  <../../../manufacturing/basic_setup/configure_manufacturing_product>`, it creates a |MO|
+
+If stock is available, no |RFQ| or |MO| is generated and the sale proceeds normally. Otherwise, the
+|RFQ| or |MO| is generated and directly linked to the originating |SO| through a smart button.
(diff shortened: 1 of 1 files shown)


Felicious:
HI @larm-odoo ! can you help with this quick review? (:

content/applications/inventory_and_mrp/inventory/warehouses_storage/replenishment/mto.rst
```diff
@@ -11,13 +11,19 @@ Replenish on order (MTO)
...
-current stock level of the product being ordered.
+- For :doc:`purchased products <../../../purchase/manage_deals/rfq>`, Odoo creates a |RFQ|
+- For :doc:`manufactured products
+  <../../../manufacturing/basic_setup/configure_manufacturing_product>`, it creates a |MO|
+
+If stock is available, no |RFQ| or |MO| is generated and the sale proceeds normally. Otherwise, the
+|RFQ| or |MO| is generated and directly linked to the originating |SO| through a smart button.
+
+This approach offers clear traceability, since each |RFQ| or |MO| is tied back to its demand.
+However, it is also rigid. If the linked |RFQ| or |MO| is canceled, the |SO| cannot be fulfilled by
```
larm-odoo: I recommend having the sentence from "However" through the end of the paragraaph is an admonition, since this is super critical info. Maybe a warning? Important?

I'd reword it if it's in a block as well. Something like:
Incomplete or partially canceled MO's and RFQ's cause SO's to be incomplete, and unable to be fulfilled.  When a linked |RFQ| or |MO| is canceled, the |SO| cannot be fulfilled, and the order must be cancelled and redone.

I don't know if that is correct/the right way of saying it, but it sounds important enough to have a block of its own.
another replenishment document.

content/applications/inventory_and_mrp/inventory/warehouses_storage/replenishment/mto.rst
```diff
@@ -133,6 +139,14 @@ Finally, click the :guilabel:`Delivery` smart button at the top of the order to
 order. Once the products have been shipped to the customer, click :guilabel:`Validate` to confirm
 the delivery.
 
+Cancelling an SO with an MTO product
+------------------------------------
+
+When a |SO| is cancelled, and it had created an |RFQ| or |MO|, only the related delivery order is
+cancelled automatically. A warning appears in the |RFQ| or |MO| chatter noting the cancellation, but
```
larm-odoo: Only the delivery order is cancelled... as opposed to what else? It sounds like other things are not cancelled. I might mention what stays the same. Maybe it's just me who doesn't fully understand this without extra context, but you can add some more info here.

content/applications/inventory_and_mrp/inventory/warehouses_storage/replenishment/mto.rst
Felicious: ty for the feedback!!

Felicious:
@robodoo r+

PR #14583, opened 2025-09-17
[ADD] VoIP: DIDWW config

jero-odoo:
New doc outlining basic configuration for DIDWW voip provider (both in the DIDWW system and in Odoo).


jero-odoo:
Hey @Felicious this is ready for review. 
A few things to note:
-not sure about the "Enable phone.systems" heading. I do think this is worth its own section, but I am not sure the heading is the best.
-I added the screenshot with the plus sign because I wanted to make it clear, but I am ok with it being removed
- I was not sure about some of the parameters, so I based in on the screenshot. I do not know the why/how behind these options
Like we discussed, I would like to do a follow up once we get this published because I would like to add some additional context if I can.
Thanks!

content/applications/productivity/voip/didww.rst
```diff
@@ -0,0 +1,84 @@
...
+Enter the name, `DIDWW`, then update the :guilabel:`WebSocket` field with `wss://sip.phone.systems`.
+Under :guilabel:`PBX Server IP`, enter `sip.phone.systems`.
+
+.. image:: didww/new-provider.png
+   :alt: The VoIP providers page in Odoo.
+
+To configure a user's VoIP provider, click the user avatar in the top-right of the database, then
+click :guilabel:`My Profile`. Click into the :guilabel:`VoIP` tab, then under the :guilabel:`Voip
+Provider` field, select *DIDWW*. Finally, enter the :guilabel:`Voip Username` and :guilabel:`Voip
+Secret`, then save.
```
Felicious: I think this should be a guilabel? Pushing this quick fix up as a commit (:

```suggestion
Provider` field, select :guilabel:`DIDWW`. Finally, enter the :guilabel:`Voip Username` and :guilabel:`Voip
Secret`, then save.
```

content/applications/productivity/voip/didww.rst
```diff
@@ -0,0 +1,84 @@
...
+
+.. image:: didww/odoo-credentials.png
+   :alt: Add alt text.
```
Felicious:
```suggestion
   :alt: DIDWW provider, username, and secret credentials entered.
```

Felicious:
@robodoo r+

PR #14639, opened 2025-09-23
[IMP] base: remove t-cache directive from ir.qweb

Gorash:
see: https://github.com/odoo/odoo/pull/224487
see: https://github.com/odoo/enterprise/pull/93321


Gorash:
@robodoo r+

PR #14646, opened 2025-09-23
[IMP] tests/requirements.txt: bump Pillow version to match Debian's

AntoineVDV:
The test suite required Pillow==9.0.1 while runbot has Pillow==10.2.0 pre-installed with Debian 24.04, which triggered an unnecessary downgrade of the package.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,3 @@
-Pillow==10.2.0
+Pillow==9.0.1
 mock==5.0.1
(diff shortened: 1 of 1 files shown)


AntoineVDV:
Not needed in the end; the `ci/documentation_guidelines` build now runs on the `ci/documentation`'s docker image, which has been rebuilt with the `tests/requirements.txt` list of `18.0`.

PR #14649, opened 2025-09-23
[IMP] conf: strip `odoo.upgrade.testing`

aj-fuentes:
The prefix that includes the full module path is just noise. This patch follows the same strategy used to remove `odoo.upgrade.util` prefix.

To be merged after: odoo/upgrade-util#324


aj-fuentes:
@robodoo r+

PR #14650, opened 2025-09-23
[IMP] tests: enable additional resource file-related checkers

AntoineVDV:
The following custom checkers were never run by the `ci/documentation_guidelines` build and had to be run locally with `make review`:
- `check_image_size`: Check that images are not larger than the maximum file size allowed for their extension.
- `check_image_color_depth`: Check that PNG images are compressed to 8-bit color depth with PNGQuant.
- `check_resource_file_name`: Check that resource file names use hyphens rather than underscores.

Since reviewers systematically perform these checks manually, it makes sense to include them in the standard test suite. Should a check raise a false positive (e.g., an `example_db.zip` resource file is added and hyphens should not be used), the red CI can be safely ignored as it is not required for merging.

### DIFF
--- a/tests/main.py
+++ b/tests/main.py
@@ -18,20 +18,23 @@ CUSTOM_RST_DIRECTIVES = [
 ]
 
 ADDITIONAL_CHECKERS = [
-    checkers.resource_files.check_resource_file_referenced
+    checkers.resource_files.check_image_size,
+    checkers.resource_files.check_image_color_depth,
+    checkers.resource_files.check_resource_file_name,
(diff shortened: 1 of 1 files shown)


content/contributing.rst
Feyensv, line 1: it's the same number of chars before and after Contributing ? (github display is confusing :D)

AntoineVDV: No, I'm just adding temp diff to trigger the CI

Feyensv: Ok then :D

AntoineVDV:
@robodoo r+

PR #14668, opened 2025-09-24
[IMP] MRP: Subcontracting lead times

jero-odoo:
Docs task: https://www.odoo.com/mail/view?model=project.task&res_id=4882474&access_token=6d6b342b-72de-48c0-a5e7-842222976069

Adding content for Subcontracting lead times and subcontracting portal.


jero-odoo:
Hey @larm-odoo this is ready for review. Thanks!

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -20,8 +21,8 @@ contracting company to better predict the delivery dates of subcontracted produc
    Unforeseen circumstances can impact the completion of these actions, which means that lead times
    should not be viewed as guarantees.
 
-Configuration
-=============
+Lead times used in subcontracting
```
larm-odoo:
```suggestion
Lead times for subcontracting
```

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,98 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
    :alt: The Delivery Lead Time field for a subcontractor, on the Purchase tab of a product page.
 
 .. note::
    Multiple subcontractors can be added to the :guilabel:`Purchase` tab on a product's page, and a
    different :guilabel:`Delivery Lead Time` can be set for each.
 
+Customer lead time
+------------------
+
+The customer lead time for subcontracted products must always be longer than the combined total of:
```
larm-odoo:
```suggestion
The customer lead time for subcontracted products must **always** be longer than the combined total of:
```
just because it sounds important

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,98 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+   - The component with the longest lead time is *Drawer Black*: 5 days
+   - Subcontractor lead time: 3 days
+   - Purchase Security Lead time: 2 days
+
+   In this scenario:
+
+   - Odoo automatically computes the preperation lead time as **5 days** based on the longest
+     component lead time.
+   - The |PO| for the subcontractor is scheduled **2 days earlier** than usual, due to the Purchase
+     Security Lead Time.
```
larm-odoo: Does 'Purchase Security Lead Time' get a GUI because it's capitalized? If not, I suggest italics

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -85,17 +171,35 @@ contracting company.
...
+
+Subcontracting portal
+=====================
+
+The *Subcontracting Portal* feature allows subcontractors to record the components they have used
+for a subcontracted product. This provides greater visibility and accuracy when managing
+subcontracted |MOs|.
+
+To allow subcontractors access to the portal, navigate to the :menuselection:`Contacts` app, and
+open a contact record. Then click the :icon:`fa-cog` :guilabel:`(cog)` icon, and select
```
larm-odoo:
```suggestion
open a contact record. Then click the :icon:`fa-cog` :guilabel:`(Actions)` icon, and select
```
That's what it says when you rollover the icon

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -85,17 +171,35 @@ contracting company.
...
+:guilabel:`Grant portal access`.
```
larm-odoo, line 226: When I try this on Runbot, I get: 

"Access Error

You are not allowed to modify 'Grant Portal Access' (portal.wizard) records.

This operation is allowed for the following groups:
	- Extra Rights/Contact Creation

Contact your administrator to request access if necessary."

Not sure if this is a common scenario for certain types of users/accounts/etc, but if there is a logic for this error that a small admonition could explain, I recommend adding that

larm-odoo (approved):
Hi @jero-odoo - great doc, all optional comments. Approved!

jero-odoo:
Hey @Felicious this is ready for final/tech review. Thanks!

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,99 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+This ensures that the subcontractor has sufficient time to receive all the necessary materials, and
+complete the production before the final delivery date.
+
+Purchase security lead time
+---------------------------
+
+As subcontracting involves purchasing components, it is important to account for potential delays
+from vendors. The *Purchase Security Lead Time* adds a buffer to the subcontractor's delivery lead
+time. Odoo automatically schedules the |PO| to be created several days earlier, reducing the risk of
+late deliveries due to unforseen delays.
```
Felicious: Since the lead times doc already defines these terms in a neutral, generic way, I think it would strengthen this section to reframe each lead time definition **in the context of subcontracting**. That way, the reader sees not just the generic meaning of each lead time, but also how it applies when subcontractors are involved.

For example, here’s a rough draft for Customer Lead Time (feel free to rephrase or polish the language, I made it really rough :sob:)
> For subcontracted products, the customer lead time must cover more than just the subcontractor’s production. It must always be longer than the combined total of:
> - raw material procurement: time required for the subcontractor to purchase components if they do not have it in stock
> - subcontracting manufacturing: time required by the subcontractor to complete the production
> - delivery to business: time required to deliver the finished product to the business that requested the product
> - (optional) additional in-house manufacturing: if the subcontracted product is a component in another manufactured product, this also includes the time required for the business to complete in-house manufacturing
> - delivery to customer: time required for the business to deliver the finished product to the customer

> note:
> For flexibility on raw material procurement and more control on raw material quality, check out [link to resupply subcontractor]. To shorten the delivery to customer, check out [link to dropship subcontractor]

After you rewrite all of the lead times sections to be subcontractor-specific, let's also include links to the lead times doc for more context for the reader!

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -85,17 +172,39 @@ contracting company.
...
+.. warning::
+  Only users who are in the :ref:`group <access-rights/groups>` *Extra Rights/Contact Creation* can
+  grant portal access.
+
+From there, click :guilabel:`Grant Access` next to the name of the appropriate contact.
+
+.. image:: basic_subcontracting_lead_times/portal-access.png
+   :alt: The portal access management pop-up window.
+
+The contact recieves an email invitiation with instructions on how to access the portal and setup
```
Felicious:
```suggestion
The contact receives an email invitation with instructions on how to access the portal and setup
```

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,99 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+.. list-table::
+   :header-rows: 1
+   :widths: 25 55 20
+
+   * - Lead Time
+     - Description
+     - Applies To
+   * - Product delivery lead time
+     - Time required to prepare and deliver components to the subcontractor. Computed automatically
+       from the component with the longest lead time.
```
Felicious: This part doesn’t look accurate as written. In the basic subcontracting route, the business doesn’t provide components to the subcontractor. The Product delivery lead time seems more likely to refer to the time it takes for the finished subcontracted product to be delivered back to the business (tho im not sure which field on the product form this would map to)

Also, could you clarify what field the “computed automatically from the component with the longest lead time” piece is referring to?

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
Felicious: P.S.: I realize that many of the issues I had within this section come from following the SME’s wording closely. That said, part of our role as writers is to go beyond transcription. SMEs provide valuable input, but it’s on us to make sure the content is factually accurate, complete, and clear for the reader. That means validating details, filling gaps, and shaping the message so it teaches the right lesson. This step is critical to our value as a writing team, and I want to see even more effort here moving forward! (:

I’m happy to talk through any parts of lead times that aren’t clear, but I want to emphasize that it’s important you build this knowledge so you can handle tasks like this more independently within the next month or so! Or, make it a habit to regularly interact with Inventory SMEs who can clarify parts you might be unsure of before review.

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,99 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+
+Purchase security lead time
+---------------------------
+
+As subcontracting involves purchasing components, it is important to account for potential delays
+from vendors. The *Purchase Security Lead Time* adds a buffer to the subcontractor's delivery lead
+time. Odoo automatically schedules the |PO| to be created several days earlier, reducing the risk of
+late deliveries due to unforseen delays.
+
+This is a general setting, and applies to all products that are purchased or subcontracted.
```
Felicious: This section isn’t quite accurate. Procuring components on the subcontractor’s side isn’t tracked in Odoo, so the Purchase Security Lead Time isn’t about accounting for their procurement delays. Instead, this lead time is used to cover delays when receiving shipments from vendors (including subcontractors).

For example, it would apply if the subcontractor were using UPS when delivering the finished subcontracted product back to us, and UPS was backed up on orders and caused some delays. As the SME put it: “The Purchase Security Lead Time adds a margin of error to the subcontractor’s delivery lead time.”

Let’s reframe this section to reflect that

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,99 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+
+Days to purchase
+----------------
+
+This setting defines the number of days the purchasing department requires to validate and process a
+|PO|. This is a general setting, and applies to all products that are purchased or subcontracted.
```
Felicious: Could we reframe this? Right now it reads as if the delay is on our purchasing department, but in the subcontracting context it’s really about **how long it takes the subcontractor to confirm they can take on the work**, and according to the SME, also encompasses "represents the time required to prepare the components that will be sent to the subcontractor". (To clarify, this is like them ordering their own raw materials and the lead time for that, it's separate from the company making the subcontracting request.

In Odoo, that confirmation happens when the subcontractor approves the RFQ. On our company’s side, that’s the point where it’s safe to validate the RFQ and turn it into a PO.

Also by "general setting", do you mean that we can only set one days to purchase lead time for the entire company? Right now, I'm just a little confused why we're mentioning products that are purchased, since that's unrelated to subcontracting right now.

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +47,99 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+   - Subcontractor lead time: 3 days
+   - Purchase Security Lead time: 2 days
+
+   In this scenario:
+
+   - Odoo automatically computes the preperation lead time as **5 days** based on the longest
+     component lead time.
+   - The |PO| for the subcontractor is scheduled **2 days earlier** than usual, due to the Purchase
+     Security Lead Time.
+   - The total customer lead time for the product *Desk Combination* should be minimum **8 days**.
```
Felicious: This manufacturing section needs a more clarification on the context. Can you please address some clarifying questions I have, so we can get all the knowledge accurate in one place, before trying to change the draft? (: 
1. On MTO and subcontracting: we talk about MTO but the doc doesn’t explain what MTO looks like in a subcontracting flow. Could you clarify:
  a.  In an MTO subcontracting scenario, would we actually see a Manufacturing Order in Odoo?
  b. To my knowledge, the subcontractor’s manufacturing tracked entirely outside of Odoo. Would the “Manufacturing Lead Time” on the BoM still be used?
  c. If yes to b, does the Manufacturing Lead Time play a different role compared to its usual usage? 
2. On the definition of "Manufacturing Security Lead Time"
  a. The section describes this as a buffer for the subcontractor’s production, but my understanding is that Manufacturing Security Lead Time only applies to **in-house manufacturing**, not subcontractor production. Could you confirm if this is accurate?
  b. If that’s true, wouldn’t this lead time only apply if the subcontracted product is itself a component in another in-house manufacturing flow?
3.  On example accuracy
  a.  The example mixes subcontractor lead time (3 days), component lead time (5 days), and purchase security lead time (2 days), but then jumps straight to customer lead time = 8 days. Could you explain how that calculation was derived?
  b. How do we set the subcontractor lead time in Odoo? Actually, this doc doesn't detail how to set up any lead time, and I think this instruction would be very important for the user. 
4. On the role of lead times in scheduling
 a. Can you clarify how the PO’s deadlines (e.g., order deadline, expected arrival date for subcontracted products) are tied to each of the lead times mentioned here? Right now, the linkage between lead times and scheduling isn’t fully explained.

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -85,17 +172,39 @@ contracting company.
...
 
    .. image:: basic_subcontracting_lead_times/expected-arrival.png
-      :align: center
       :alt: The Expected Arrival date on a PO for a subcontracted product.
 
    Bike Friends begins manufacturing the tricycle on May 3rd — the day that the |PO| is confirmed —
    and finishes on May 6th, three days later.
 
    The tricycle is then shipped to Mike's Bikes the same day, and they receive it on May 8th, two
    days later.
```
Felicious, line 215: so glad we're getting rid of the bike friends example; it doesn't really have much relevance to subcontracting

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -85,17 +172,39 @@ contracting company.
...
+
+Subcontracting portal
+=====================
```
Felicious, line 218: overall a good addition!

Felicious (changes requested):
Hi @jero-odoo, thanks for tackling this very complex topic and incorporating the SME feedback! I especially appreciate the effort you put into summarizing all the lead times in a table at the end (: that really helps pull the concepts together

That said, there are still some significant areas to revisit. Lead times in subcontracting are tricky, and I think some of the misunderstandings here stem from gaps in the base knowledge of how manufacturing and purchase lead times actually work in this context. I completely understand that the original source material didn’t provide much clarity (and it’s possible even the original author didn’t fully grasp it), but this is where we add value as writers: giving readers the necessary context to understand why they’re configuring lead times a certain way, so they can map them to their own operations.

In my review, I’ve highlighted:
- Reframing opportunities: Some definitions are correct but too generic; reframing them in subcontracting terms would make them much more useful.
- Misunderstandings to correct: A few areas suggest a misinterpretation of how lead times behave. These will need rework, and I’m happy to revisit them together if helpful.
- Context gaps: Some workflows (e.g., around security lead times) can’t be explained without setting up more prerequisite information. Without that context, the reader won’t be able to follow.
- Missing angles: There are also some important features worth noting (like how PO deadlines are tied to lead times) that would strengthen this doc.

I’ve added detailed comments inline, but if you’d prefer, I’m happy to hop on a call to walk through the logic and examples together! My goal is to make sure the background around subcontracting and lead times feels clear and solid, so that future docs in this area are easier and more straightforward for you to tackle (:

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
jero-odoo: I actually think I am going to remove this section, I don't think it is adding anything and may in fact be more confusing than helpful

jero-odoo:
Hi @Felicious I've made siginificant updates based on the feedback provided, this is ready for re-review. I have also reached out to the SME for additional insight. 
Thank you

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +49,125 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+
+The Days to Purchase Lead Time defines the time between sending a |RfQ| and validating the |PO|.
+
+In a subcontracting context, this period represents:
+
+The time it takes for the subcontractor to confirm they can accept the work (i.e., when they approve
+the |RfQ| in Odoo).
+
+The time needed to prepare the components that will be sent to the subcontractor, so they are ready
+by the time production begins.
```
Felicious, line 109: kind of seems like these two items are bullet points?

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +49,125 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+
+Once the subcontractor approves the |RfQ|, it is safe for the company to validate the |RfQ| and
+convert it into a |PO|.
+
+This is a general setting in Odoo, meaning only one value can be configured for the entire company.
+It applies to all subcontracted products, but not to directly purchased products.
```
Felicious, line 115: I'm pretty sure it applies to both subcontracted and purchased products. Just curious whether Magdalena told you otherwise?

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +49,125 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+In a subcontracting flow, products configured with |MTO| do not generate a Manufacturing Order in
+Odoo. Instead, Odoo creates a |PO| to the subcontractor. The subcontractor's production is managed
+outside of Odoo, but lead times are still important for scheduling.
+
+- Manufacturing Lead Time: This field represents the subcontractor's production lead time (the time
+  the subcontractor needs to manufacture the product once they have all the necessary components).
+- Manufacturing Security Lead Time: This setting applies **only** to in-house manufacturing. It does
+  not buffer subcontractor production. However, if a subcontracted product is used as a component in
+  an in-house Manufacturing Order, then manufacturing security lead time applies to that downstream
+  in-house process.
```
Felicious, line 129: After reading this, I still have a lingering question

My main confusion is: if no MO is created in this flow, how do these manufacturing lead times actually get applied? In other words, if we’re asking users to set manufacturing lead times for subcontracted products, how does the MTO process use that data to notify them about planning deadlines or scheduling constraints?

Could you clarify this point more explicitly in the doc? I think it'll be helpful for the readers!

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
```diff
@@ -46,13 +49,125 @@ Once the subcontractor has been added, enter the number of days it takes them to
...
+  - Odoo schedules the PO for *Desk Combination* based on the 5-day component lead time.
+  - The |PO| is created **2 days earlier** than the calculated date due to the purchase security
+    lead time.
+  - The expected delivery date from the subcontractor is **3 days** after the |PO| date, reflecting
+    the subcontractor's vendor lead time.
+  - If *Desk Combination* is used in a downstream in-house |MO|, any additional manufacturing
+    security lead time would apply to that in-house operation.
+
+  This example shows how each lead time contributes to scheduling, rather than simply summing them
+  into a single customer lead time.
```
Felicious, line 169: Jess, I still need some clarification here.

1. On the BoM for Desk Combination:

- Is Drawer Black purchased directly from a vendor to the company? (5 days)
- Is [random component name] produced by a subcontractor and then sent to the company? (3 days)

Right now, the way the sentence is written—“how each lead time contributes to scheduling”—makes it sound like both items are components that need to be gathered at the company for in-house manufacturing. That isn’t clear in the text. Can you rewrite so the roles of each component are explicit?

2. if in-house manufacturing is part of this example, why isn’t the manufacturing lead time mentioned?

3. Finally, can you explain why we’re using this specific example to tie together all the lead times defined earlier? I’m not saying it’s the wrong choice, but I want to be sure the example actually verifies the reader’s understanding of all the lead times we’ve introduced.

content/applications/inventory_and_mrp/manufacturing/subcontracting/basic_subcontracting_lead_times.rst
Felicious: Can you elaborate why we're bringing back the Bike Friends example? As mentioned in the previous review, I liked that we were replacing it with the desk combination example, that's more specific to subcontracting

Felicious (changes requested):
Hi @jero-odoo,

This revision improves the accuracy and clarity of the concepts. Two areas still need clarification: I’ve left comments in the MTO and Desk Combination sections.

Once those are addressed, re-tag me in review. I also recommend involving an inventory/manufacturing expert to confirm the details. Lead times are already complex, and we risk misleading readers if anything here is incorrect.

PR #14670, opened 2025-09-24
[ADD] MRP: Delayed production filter

jero-odoo:
Docs task: https://www.odoo.com/mail/view?model=project.task&res_id=4873706&access_token=32da3bf0-be7b-4dea-84b7-
f392c3a59e10

Adding documentation about the "Delayed production" filters for manufacturing orders
Originally this was going to be added to one of the reporting docs, but since it is independent of the reports, it seemed like it deserved its own doc.


jero-odoo:
Hey @Felicious sending this to your for peer and tech review. Thanks!

content/applications/inventory_and_mrp/manufacturing/reporting/delayed.rst
```diff
@@ -0,0 +1,61 @@
...
+- :guilabel:`Late`: the |MO|'s *scheduled end date* exceeds the deadline, highlighted in red.
+- :guilabel:`Late Availability`: one or more required components are not available before the
+  deadline. For example, a confirmed purchase order or manufacturing order for components is
+  scheduled to end *after* the |MO| deadline.
+- :guilabel:`Components Available`: all components are available to begin production.
+
+Use case
+========
+
+Consider an |MO| with a deadline of **June 17th**:
```
Felicious: image

"finishes" nitpick but im wondering if it'd be good to be very specific with the technicals here. What does it mean for an MO to have a deadline, Odoo? What does it mean for the MO to finish? how about be as specific as 

"MO with a Scheduled End date of June 17", with a screenshot of it in Odoo? (:

Felicious (approved):
Clear and concise doc, @jero-odoo! 🎉

I have just one request: could you add an image at the end showing the specific fields in Odoo that are being referenced? That extra visual context would make the use case even clearer.

Overall, nice work! I appreciate how straightforward this was, and the use case came through clearly and was easy to understand.

@robodoo delegate+

jero-odoo:
@robodoo r+

PR #14677, opened 2025-09-25
18.0 website building blocks form masp

masp-odoo:


masp-odoo:
Close because I did another one on 18.0 and don't know how to cancel it.

PR #14692, opened 2025-09-27
[FIX] Incorrect example in Webhook doc

AcroVenom:
Updates documentation to clearly distinguish between user-testing errors and errors stemming from the Odoo documentation itself. Lack of clarity, particularly with complex topics like Odoo ORM and Webhooks, can discourage users and erode credibility when the perceived error is not their fault.


mcsi-odoo:
Hi @AcroVenom :slightly_smiling_face: Thanks for your contribution. We're currently preparing an update of the webhooks page as part of PR [#14540](https://github.com/odoo/documentation/pull/14540/), so I'll include your fix in there.

PR #14754, opened 2025-10-01
[IMP] payment: remove _should_build_inline_form references from docs

csan-odoo:
Docs updated to reflect current payment form behavior.

Ticket-4911075


locale/sources/developer.pot
AntoineVDV, line 1: You don't need to change POT files, as it's automatically done in general, but also, we don't translate the developer documentation 😁

AntoineVDV: Also, "[IMP] reference/payment_provider: remove _should_build_inline_form method" for the commit message.

csan-odoo: changes made :)

PR #14764, opened 2025-10-02
[FIX] Journla typo

theRealThagomizer:
Fixed a typo in US payroll localization page around journal/journla


theRealThagomizer:
@robodoo r+

PR #14765, opened 2025-10-02
[IMP] Subscriptions: Align to period start improvement

theRealThagomizer:
Updated language on the subscriptions page around "align to period start"

https://www.odoo.com/odoo/all-tasks/4610815

This 18.0 can FWP up to master.


content/applications/sales/subscriptions.rst
```diff
@@ -91,8 +91,18 @@ configuration fields:
...
+  day of the next :guilabel:`Billing Period`. When a subscription plan is purchased in the middle
+  of a billing period, the invoice shows a reduced cost.
+
+  .. example::
+     On the 15th of July, a customer purchases a monthly subscription for a streaming service.
+     Selecting :guilabel:`Align to Period Start` for this recurring plan ensures that all
+     subscriptions are billed on the 1st of each month. Their subscription begins as soon as 
+     payment is confirmed. The quotations shows the full cost of the subscription, but the invoice
+     shows an adjusted cost and has a note explaining the time period that is covered. On the 1st of
+     August, the customer is charged the full price for the month's subscription.
```
Felicious: I think we can be even more specific with the billing period here in the example, to make sure the reader really understands how this works! (:
```suggestion
     On the 15th of July, a customer purchases a monthly subscription for a streaming service.
     Selecting :guilabel:`Align to Period Start` for this recurring plan ensures that all
     subscriptions are billed on the 1st of each month. Their subscription begins as soon as 
     payment is confirmed. The quotation shows the full cost of the subscription for July 1 - 31, but the invoice
     shows an adjusted cost for July 15 - 31. On the 1st of
     August, the customer is charged the full price for the month's subscription.
```

Optional: I took out "has a note explaining the time period that is covered" because I thought that if we were specific about naming the billing period July 15-31, we didn't need this phrase anymore. What do you think?

theRealThagomizer:
> This looks good to me, @theRealThagomizer ! Nice job digging into the billing period behavior and double-checking it with the SME! (:
> 
> I left one optional note for added specificity, since I wasn’t completely sure about the billing period detail.
> 
> I also noticed the automatic checks failed because of a trailing whitespace. You can catch this in the future with this [VScode extension](https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces). (I’ve added it to the onboarding setup doc too!)
> 
> For next time, just make sure the CI checks pass before tagging the reviewer.

Thanks for the feedback, Felicia! I implemented your commit. I also downloaded the extension you suggested and will start using it going forward.

theRealThagomizer:
@robodoo r+

PR #14822, opened 2025-10-09
[ADD] New tiktok connector marketplace

jbelloc2:
Adding this because we are about to release a new marketplace: Tiktok. this documentation aim to guide users on how to best start sycnronizing their shops in Odoo

task-3690836


kidoodoo:
Doing it in a separate PR #16833

PR #14831, opened 2025-10-10
[IMP] ecommerce: payments

auva-odoo:
task-4294617

used the opportunity to update the checkout page (but not add new features; this will be done in a separate PR)


auva-odoo:
@robodoo r+

PR #14851, opened 2025-10-13
[IMP] Subscriptions: General style alignment

theRealThagomizer:
This PR updates 5 files, making the following changes:

- removing all instances of future tense from subscriptions.rst and automatic_alerts.rst
- performing optimizations suggested by `make review` on subscriptions, automatic_alerts, upselling, reports, and eccomerce.rst
- revising language in subscriptions.rst for readability

This 18.0 PR can FWP up to master.

### DIFF
--- a/content/applications/sales/subscriptions.rst
+++ b/content/applications/sales/subscriptions.rst
@@ -48,48 +48,48 @@ Set up recurring plans
 To get started with subscription products in Odoo, *recurring plans* (previously known as
 *recurrence periods*) must first be configured.
 
-Recurring plans are the time windows in which subscriptions are active before they renew again.
-While a subscription is active, customers receive products or services, and may also have access to
-additional benefits such as support desk triage. In terms of payment, these recurring plans
-designate how often the customer is charged in order to maintain the benefits of their subscription.
+Recurring plans are the time periods when subscriptions are active before they need to be renewed
+again. While a subscription is active, customers receive products or services, and may also have
+access to additional benefits like support desk triage. Recurring plans also set the interval when
+the customer is charged in order to maintain their subscription.
 
 To configure recurring plans, go to :menuselection:`Subscriptions app --> Configuration -->
 Recurring Plans`.
 
-By default, the **Subscriptions** app includes a number of common recurring plans already available,
-such as :guilabel:`Monthly` and :guilabel:`Yearly`.
+The **Subscriptions** app includes :guilabel:`Monthly`, :guilabel:`6 Months`, and :guilabel:`Yearly`
+as default recurring plan options.
 
 Create a new recurring plan by clicking :guilabel:`New` on the :guilabel:`Recurring Plans`
-dashboard, to reveal a blank form where the plan :guilabel:`Name`, :guilabel:`DETAILS`,
-:guilabel:`SELF-SERVICE` and :guilabel:`Pricing` field values are specified.
+dashboard. A new recurring plan form is generated with the :guilabel:`Name` and some
+:guilabel:`DETAILS` filled out by default. These can all be changed freely.
 
 .. image:: subscriptions/recurring-plan-blank-form.png
    :alt: A blank recurring plan form in the Odoo Subscriptions application.
 
 .. important::
-   The `Days` unit of measure *cannot* be used as a :guilabel:`Billing Period` for subscription
-   products. The daily recurrence period in Odoo is designated for rentals, and **cannot** be added
-   to subscription-based sales orders.
-
-   This limitation is there to avoid sales orders that would generate daily invoices.
+   `Days` *is not* available as an option in the :guilabel:`Billing Period` drop-down menu.
+   Recurring plans cannot be configured to bill on a daily basis.
 
 DETAILS section
 ---------------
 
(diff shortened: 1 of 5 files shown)


content/applications/sales/subscriptions/reports.rst
```diff
@@ -208,8 +208,8 @@ Reporting pages
 In the Odoo **Subscriptions** app, there are four different reporting pages available.
 
 To access, analyze, and customize various reports related to subscriptions, navigate to
-:menuselection:`Subscriptions app`, and click the :guilabel:`Reporting` drop-down menu in the
-header to reveal the following reporting pages:
+:menuselection:`Subscriptions app`, and click the :guilabel:`Reporting` drop-down menu in the header
+to reveal the following reporting pages:
```
Felicious: This is more the tutorial writing style! For docs, you can simply write:


```suggestion
:menuselection:`Subscriptions app --> Reporting` to reveal the following reporting pages:
```

content/applications/sales/subscriptions.rst
```diff
@@ -177,120 +175,108 @@ navigating the recurring revenue sources attached to the plan:
...
-Manually create a new customer subscription by navigating to either the :menuselection:`Sales` or
-:menuselection:`Subscriptions` app dashboards, and then clicking :guilabel:`New`.
+Manually creating a subscriptions quotation
+===========================================
 
-.. note::
-   Products that have been marked as :guilabel:`Subscriptions` on their product forms, and are also
-   sold on the **eCommerce** website will *automatically* create and confirm subscription quotations
-   in the backend of Odoo.
+A new customer subscription can be created by going to either the **Sales** or **Subscriptions** app
```
Felicious: If we're navigating somewhere, use menuselection! Can you change the bold for both sales and subscription to menuselection?

content/applications/sales/subscriptions.rst
```diff
@@ -177,120 +175,108 @@ navigating the recurring revenue sources attached to the plan:
...
 
 .. _subscriptions/quotations:
 
-Create a subscriptions quotation
-================================
-
-Manually create a new customer subscription by navigating to either the :menuselection:`Sales` or
-:menuselection:`Subscriptions` app dashboards, and then clicking :guilabel:`New`.
+Manually creating a subscriptions quotation
+===========================================
```
Felicious, line 230: I noticed that other instances of bullet points, we follow the format:

`- thingy: definition`, but in this section, you do `thingy, definition`. To maintain consistency with the rest of the document, can you change the items in the list in this section to be formatted the same as the rest of the article?

content/applications/sales/subscriptions.rst
```diff
@@ -177,120 +175,108 @@ navigating the recurring revenue sources attached to the plan:
...
+     into a sales order within the specified date range.
 
-- :doc:`Pricelist <sales/products_prices/prices/pricing>`, if one is available and appropriate to
-  use (i.e., summer sale discount, VIP customer, etc.).
-- :guilabel:`Payment Terms`, to set a specified time window for when the subscription must be paid.
-  This is not to be confused for when the quotation is *confirmed* and becomes a sales order, to
-  where, payment may then be obtained immediately or within a certain amount of days, weeks, months,
-  etc.
+- :doc:`Pricelist <sales/products_prices/prices/pricing>`, to provide a unique discount, seasonal
+  pricing, etc. (if one has been created.)
```
Felicious: the quotation template and pricelist fields seem to be optional. Based on my what I remember, our team typically doesn't explicitly point out optional fields, but we point out required ones, for brevity. Unless you think our users are creating a lot of unnecessary quotation templates or getting confused about whether they're required, I don't think it's necessary to point out that it's optional! Just an opinion (:

Felicious: On a separate note, our team typically puts the period after/outside the parentheses, with no period before the parentheses 
- so I'd write: `Sentence (blabla).` 
- and wouldn't write `sentence. (blabla.)`

content/applications/sales/subscriptions.rst
```diff
@@ -177,120 +175,108 @@ navigating the recurring revenue sources attached to the plan:
...
-:guilabel:`Recurring Plan`, as well as the :guilabel:`Order Lines` tab.
-
-Optionally, specify a:
-
-- :doc:`Quotation Template <sales/sales_quotations/quote_template>`, if one is readily available to
-  help populate the form fields.
+- :guilabel:`Invoice Address` and :guilabel:`Delivery Address`, to differentiate between the address
+  where a customer is invoiced and the address where a subscription that includes a physical good is
+  sent. This option must be enabled by configuring the :doc:`Customer Addresses
+  </applications/finance/accounting/customer_invoices/customer_addresses>` feature.
```
Felicious: So can you help me clarify: Is Delivery Address a field that only appears if the Customer Addresses feature is enabled? If so, let's format this page this way:

```
- Invoice Address: [insert definition]
- Delivery Address: visible only when the Customer Addresses feature is enabled. [definition: something like 
"used to set a delivery address that differs from the billing address, which is set in the Invoice address field]

theRealThagomizer: I'm actually going to ax this for now since it's an optional setting customers can enable. Maybe in the future, we'll create an "Optional subscription configurations" page.

theRealThagomizer:
@robodoo r+

PR #14882, opened 2025-10-16
[I18N] *: re-export pot files + remove obsolete pot

ticodoo:
PR does 2 things:

- Re-export pots, including if no terms have been updated due to some languages not being in sync with the pot file in Weblate (i.e. po files already existed and msgmerge is not run until Weblate detects a change in the pot file)
- remove `settings.pot` + its translations/references since it appears to be leftover from v14 (i.e. was probably deleted in a refactoring). We remove it in stable so translators don't waste time translating things that are never seen by users.


oomsveta (approved):
robodoo r+

oomsveta:
robodoo rebase-ff

PR #14910, opened 2025-10-17
[IMP] Taxes: Python taxes update

antoine162:
task-4681891

In 18.0, Python taxes were revamped (for mostly technical reasons), but some changes were also made to the UI.

Specifically, the tax computation option was renamed from `Python Code` to `Custom Formula`, the `Python Code` field was renamed to `Formula` and changed to accept a Python expression rather than a block of code, and the `Applicable Code` field was removed.


antoine162:
Done as part of https://github.com/odoo/documentation/pull/15059

PR #14958, opened 2025-10-21
[IMP] Contacts: Address type correction

jero-odoo:
Updating to remove references to "Private Address" options, as well as updating images to reflect UI updates
Docs task:https://www.odoo.com/mail/view?model=project.task&res_id=5171761&access_token=f520dcd1-d8ff-4153-a41a-aa06d857d8fd


jero-odoo:
@Felicious sending this straight to you for final/tech. Thanks!

content/applications/essentials/contacts.rst
```diff
@@ -17,41 +17,39 @@ To create a new contact, navigate to the :menuselection:`Contacts app`, and clic
 Contact type
 ------------
 
-Odoo allows for both :guilabel:`Individual` and :guilabel:`Company` contacts. Select either
-:guilabel:`Individual` or :guilabel:`Company`, depending on the type of contact that is being added.
+Odoo allows for both :guilabel:`Person` and :guilabel:`Company` contacts. Select either
+:guilabel:`Person` or :guilabel:`Company`, depending on the type of contact that is being added.
```
Felicious: Feels like this line could be more concise? Optional suggestion!


```suggestion
Odoo supports both :guilabel:`Person` and :guilabel:`Company` contacts. Select the
appropriate type for the new contact.
```

content/applications/essentials/contacts.rst
```diff
@@ -17,41 +17,39 @@ To create a new contact, navigate to the :menuselection:`Contacts app`, and clic
...
-First, fill in the name of the :guilabel:`Individual` or :guilabel:`Company`. This is how the name
+First, fill in the name of the :guilabel:`Person` or :guilabel:`Company`. This is how the name
 appears throughout the database. This field is **mandatory**.
 
 .. tip::
-   :guilabel:`Individual` contacts can have a :guilabel:`Company` contact linked to it. After
-   selecting :guilabel:`Individual`, a new :guilabel:`Company Name...` field appears below the
-   first name field.
+   :guilabel:`Person` contacts can have a :guilabel:`Company` contact linked to it. After selecting
+   :guilabel:`Person`, a new :guilabel:`Company Name...` field appears below the first name field.
```
Felicious: I'm a bit confused about the behavior described here. When I selected the **Person** radio button, I didn't see a company contact appear. Instead, I saw **Company** field. Could you clarify what you meant by “company contact” here?

Image

Also, in my view, the field label was **Company**, with the placeholder text “Company Name…”, and it appeared below the image or **Phone** field, not under the first name. In fact, there was only one field for the full name. 

Maybe we’re referring to different things?

For this tip, I recommend simplifying it for brevity, something like this:

```suggestion
   When :guilabel:`Person` is selected, the :guilabel:`Company  field becomes visible.
```

content/applications/essentials/contacts.rst
```diff
@@ -17,41 +17,39 @@ To create a new contact, navigate to the :menuselection:`Contacts app`, and clic
...
-Next, enter the :guilabel:`Address` of the :guilabel:`Company` or :guilabel:`Individual`.
+Next, enter the :guilabel:`Address` of the :guilabel:`Company` or :guilabel:`Person`.
 
 .. tip::
-   If the :guilabel:`Individual` option is chosen, then the *type of address* can be chosen from a
+   If the :guilabel:`Person` option is chosen, then the *type of address* can be chosen from a
    drop-down menu. Options for this drop-down menu include: :guilabel:`Contact`, :guilabel:`Invoice
-   Address`, :guilabel:`Delivery Address`, :guilabel:`Other Address`, and :guilabel:`Private
-   Address`.
+   Address`, :guilabel:`Delivery Address`, and :guilabel:`Other Address`.
```
Felicious: I don't believe this is accurate to the current version. From the way I'm reading this doc, it sounds like the Address type selection is on the contact form, but I couldn't find it.

Image

Instead, you have to click the "Add Contact" button and that's where the Contact, Invoice, and Delivery options appear. I'm slightly confused by this new layout, though, and I think we might have to make more changes to this page to lead up to these instructions. 

Image

content/applications/essentials/contacts.rst
```diff
@@ -71,27 +69,21 @@ At the bottom of the contact form are several tabs. On the :guilabel:`Contacts &
...
-On the :guilabel:`Create Contact` pop-up form, start by clicking the default :guilabel:`Other
-Address` field at the top to reveal a drop-down menu of address-related options.
-
-Select any of the following options:
+On the :guilabel:`Create Contact` pop-up form,  select one of the following options:
 
 - :guilabel:`Contact`: adds another contact to the existing contact form.
 - :guilabel:`Invoice Address`: adds a specific invoice address to the existing contact form.
 - :guilabel:`Delivery Address`: adds a specific delivery address to the existing contact form.
 - :guilabel:`Other Address`: adds an alternate address to the existing contact form.
```
Felicious: It seems like they're no longer displaying "Address"

Image

content/applications/essentials/contacts.rst
Felicious: Now that I've read the document to the end, I think the only place we should talk about the address type is in the "Create Contact" pop-up and we should just remove this. 

What I'd need your help determining: what should the Address in the contact form always be? For example, if we're creating an invoice contact type, what should they put in the Address field of the contact form? Is it different from the one they'd enter on the "Create Contact" pop-up?

Felicious (changes requested):
Thanks for starting to address Ray’s feedback, @jero-odoo!

It seems this doc is more outdated than we expected and may require broader structural updates to align with the current version. So, I'm bumping this to at least 2 points.

I’ll leave the call to you: if you think a major restructuring is needed and the total changes exceed ~50%, we can increase the PR value to 3 pts.

That said, we should definitely revisit the “Contact form” section, as its structure likely needs adjustment. Also, I couldn’t find the Address Type drop-down in the 19.0 runbot (I might’ve missed it ! feel free to point it out!). If the Address Type can now only be set in the Create Contact form, we’ll need to update the doc accordingly.

content/applications/essentials/contacts.rst
jero-odoo: "I'm a bit confused about the behavior described here. When I selected the Person radio button, I didn't see a company contact appear. Instead, I saw Company field. Could you clarify what you meant by “company contact” here?"
This means that "person" contacts can be linked to existing or new company contacts. You need to link them in the company field. Does that make sense?

content/applications/essentials/contacts.rst
jero-odoo: correct, this is on the "add a contact" tab, not the address tab. It needs some additional rewriting to reflect the changes

jero-odoo: I am a little confused by what you are asking here. You cannot select the address type on the MAIN contact form (directly under the contact name). You select it on the "Add contact" tab, which is where you add "alternate" contacts.

jero-odoo:
Hey @Felicious this is ready for another review. It involved some pretty significant changes (more than 50%).
Something to note, the UBO tab is related to the Equity app, and there is no docs on that yet. I included information on how to request the contact to fill out the form for that, and how the tab reflects information from the app, but I did not dig too deep into editing the tab. It is possible to make changes to the holdings on the tab within the contact record, but I do not know if it is necessarily recommended (since it involves stocks and voting shares, I didn't want to overstep), so I kept the information minimal.
Let me know if you have any questions. Thanks!

Felicious (approved):
Thanks for your efforts on updating the doc in its entirety, and looking into the UBO tab, @jero-odoo (: 

@robodoo r+

PR #14979, opened 2025-10-22
[IMP] orm: add warning about model extension

emi-odoo:
When a model is defined with `_inherit` using a list, Odoo will do a check with the class name to set the `_name` of the model being extended. This is different than the previous versions and can be misleading

### DIFF
--- a/content/developer/reference/backend/orm.rst
+++ b/content/developer/reference/backend/orm.rst
@@ -1227,18 +1227,20 @@ When using :attr:`~odoo.models.Model._inherit` but leaving out
 :attr:`~odoo.models.Model._name`, the new model replaces the existing one,
 essentially extending it in-place. This is useful to add new fields or methods
 to existing models (created in other modules), or to customize or reconfigure
-them (e.g. to change their default sort order)::
+them (e.g. to change their default sort order)
+
+.. code-block:: python
 
     class Extension0(models.Model):
-    _name = 'extension.0'
-    _description = 'Extension zero'
+        _name = 'extension.0'
+        _description = 'Extension zero'
 
-    name = fields.Char(default="A")
+        name = fields.Char(default="A")
 
     class Extension0(models.Model):
-    _inherit = ['extension.0']
+        _inherit = 'extension.0'
 
-    description = fields.Char(default="Extended")
+        description = fields.Char(default="Extended")
 
 .. code-block:: python3
 
@@ -1270,7 +1272,9 @@ model.
 
 The main difference is in the meaning. When using Delegation, the model
 **has one** instead of **is one**, turning the relationship in a composition
-instead of inheritance::
+instead of inheritance
+
+.. code-block:: python
 
     class Screen(models.Model):
         _name = 'delegation.screen'
@@ -1338,7 +1342,9 @@ In that case, the attributes of the field are taken from the parent class
 and overridden by the ones given in subclasses.
 
 For instance, the second class below only adds a tooltip on the field
-``state``::
+``state``
+
+.. code-block:: python
 
     class FirstFoo(models.Model):
         state = fields.Selection([...], required=True)


kmagusiak (approved):

emi-odoo:
@robodoo r+

emi-odoo:
@kmagusiak I `r+` too soon lol
Are you the one supposed to `r+` or to delegate it?

kmagusiak:
> @kmagusiak I `r+` too soon lol Are you the one supposed to `r+` or to delegate it?

I cannot. @Julien00859 or @HydrionBurst

Julien00859 (changes requested):
Can you make another commit to fix the indentation, so we can backport that commit? Also we can't r+ it, we need someone from the documentation team

emi-odoo:
@Julien00859 you mean the indentation for the code examples of inheritance? If yes: will do 🫡

content/developer/reference/backend/orm.rst
```diff
@@ -1229,16 +1229,18 @@ essentially extending it in-place. This is useful to add new fields or methods
 to existing models (created in other modules), or to customize or reconfigure
 them (e.g. to change their default sort order)::
```
Julien00859: `::` makes what follow a code-block, if you change the thing for a `.. code-block` make sure to remove `::`
```suggestion
them (e.g. to change their default sort order)
```

content/developer/reference/backend/orm.rst
```diff
@@ -1275,6 +1277,8 @@ The main difference is in the meaning. When using Delegation, the model
 **has one** instead of **is one**, turning the relationship in a composition
 instead of inheritance::
```
Julien00859: here too

content/developer/reference/backend/orm.rst
```diff
@@ -1343,6 +1347,7 @@ and overridden by the ones given in subclasses.
 For instance, the second class below only adds a tooltip on the field
 ``state``::
```
Julien00859: and again

content/developer/reference/backend/orm.rst
```diff
@@ -1343,6 +1347,7 @@ and overridden by the ones given in subclasses.
...
 
+.. code-block:: python
```
Julien00859, line 1347: and empty line
```suggestion
.. code-block:: python

```

content/developer/reference/backend/orm.rst
```diff
@@ -1249,6 +1251,12 @@ will yield::
 
     {'name': "A", 'description': "Extended"}
 
+.. warning:: if you use :attr:`~odoo.models.Model._inherit` with a list
+    but leaving out :attr:`~odoo.models.Model._name`, the class
+    name will be used to infer the model you're extending.
+    This is not the case when :attr:`~odoo.models.Model._inherit`
+    is a ``string``
```
HydrionBurst: From the code I feel we should claim
`_inherit = 'str'` will set the default `Model._name` and is the only case we don't need to explicitly specify `Model._name`

So we can infer
```py
Class MyModel:
    # _name = 'my.model'
```
will automatically use `_name = 'my.model'` and log a warning, no matter if it has `_inherit = [...]` or not

emi-odoo: @HydrionBurst -> good point! will add it

emi-odoo:
@Julien00859 -> I force pushed as I needed to resplit the work on the two commits (one for the warning and one for the code examples)

content/developer/reference/backend/orm.rst
```diff
@@ -1249,6 +1251,11 @@ will yield::
 
     {'name': "A", 'description': "Extended"}
 
+.. warning:: When :attr:`~odoo.models.Model._inherit` is set using a string,
+    this will also set the value of :attr:`~odoo.models.Model._name`.
+    In all the other cases the :attr:`~odoo.models.Model._name` is set by
+    the attribute itself or by the class name (i.e: ``class MyModel(models.Model)`` will be
+    translated into ``_name = 'my.model'``, even if not explicitly set)
```
HydrionBurst: we raise a warning `_logger.warning("Class %s has no _name, please make it explicit: _name = %r", name, attrs['_name'])` when infer model name from class name. So I think developers shouldn't use this feature. And we don't have to document it here which increase the complexity.

```suggestion
    In all the other cases you need to explicity set the :attr:`~odoo.models.Model._name`.
```

emi-odoo: I had developers being confused anyways because in the code example we're using `_inherit = [...]` (at least before this PR we were). And they didn't notice the class name. For new devs I think it's easy to get confused of the relationship between the class name and the model name and the table name. People will try to do it even if they don't realize what they're doing, so, from my point of view, it's better to explicitly explain what's going on in case they make a mistake :)

Julien00859:
```suggestion
.. warning:: When :attr:`~odoo.models.Model._inherit` is set to a string,
    then :attr:`~odoo.models.Model._name` is set to the same value,
    unless `_name` is explicitly set.
```

emi-odoo:
@Julien00859 @HydrionBurst -> adapted using JUC's suggestion! I feel like we still have the chance to confuse devs but better than before :D

content/developer/reference/backend/orm.rst
HydrionBurst: I would simplify everything, the description of `inhertance` is also not accurate.

Inheritence
`_name` is not any `_inherit`

```
_name = 'a'
_inherit = 'b'
```
```
_name = 'a'
_inherit = ['b']
```
```
_name = 'a'
_inherit = ['b', 'c']
```
Extension
`_name` is one of `_inherit` and doesn't have to be specified only when `_inherit` is a str
```
_inherit = 'a'
```
```
_name = 'a'
_inherit = 'a'
```
```
_name = 'a'
_inherit = ['a']
```

Delegation
`_inherits`

emi-odoo: I agree, but maybe this could be for another task/PR, what do you think?

HydrionBurst: No problem, it is already much better than before.

BTW, I think the message shouldn't be a `warning` 🤔 

And the `Note` in the next line is for the content `{'name': "A", 'description': "Extended"}` in the previous line.
So we shouldn't add this content in between.

I would do
```
{'name': "A", 'description': "Extended"}. # It will also yield .....

Note: your new content
```

emi-odoo: @HydrionBurst a note would suffice! After the rephrasing especially :D

will check the format this evening :D

emi-odoo:
@robodoo r+

emi-odoo:
robodoo rebase-ff

emi-odoo:
@robodoo r+

PR #14992, opened 2025-10-23
[ADD] Fleet: Add reporting doc

larm-odoo:
No doc currently exists for this.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/4928588) for this PR.


larm-odoo:
Hi @Felicious - this is the LAST Fleet doc that needed to be created =)

content/applications/hr/fleet/cost_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+To view the *Cost Analysis Report*, navigate to :menuselection:`Fleet app --> Reporting --> Costs`.
+The default report shows a bar chart of service and contract costs for the current year, organized
+by month.
+
+.. image:: cost_analysis/cost-analysis.png
+   :alt: The default cost analysis report showing contract and service costs.
+
+In this report, it is possible to determine that June had the highest total costs, with almost
+$14,000 dollars in contract and repair costs, while January had the lowest costs, with less than
+$2,000 of contract costs and no repairs.
```
Felicious: what do you think about putting this in an example admonition block, since the user probably wouldn't see the same figures? I won't force you to do it, since i think your paragraph is written in a tone that sounds theoretical, so it's fine out here on its own

content/applications/hr/fleet/cost_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+To view these metrics, open the default :ref:`cost analysis report <fleet/total-costs>`. Next, click
+into the search bar, and click :guilabel:`Vehicle` under the :icon:`oi-group` :guilabel:`Group By`
+column.
+
+.. image:: cost_analysis/by-vehicle.png
+   :alt: The cost analysis report showing annual costs by vehicle.
+
+From this report, it can be determined the `Toyota Corolla` with the license plate `STU-1718` cost
+the most, with almost $14,000 in costs for the year. Additionally, the both `Nissan Micro` vehicles
+and the `Ford Focus` cost the least, with under $5,000 in total costs each.
```
Felicious: on the other hand, the content here sounds more like it's referring to the image so I think we should put the example block here

content/applications/hr/fleet/cost_analysis.rst
```diff
@@ -0,0 +1,89 @@
...
+:guilabel:`Variation` column, showing increases or decreases in costs from the third and fourth
+quarter of the current year.
+
+.. image:: cost_analysis/detailed-costs.png
+   :alt: A pivot table with a detailed cost comparison between the current and previous year.
+
+In this example, it can be determined that the `Ford Focus` had the greatest change in terms of
+lowered total costs, with a 75.25% reduction in costs. Additionally, the `Nissan Micro` with the
+license plate of `DEF-456` was the only vehicle with an overall increase in total costs, with an
+increase of 41.78%.
```
Felicious: example? (:

content/applications/hr/fleet/cost_analysis.rst
larm-odoo: I was struggling with this doc, because the info you get by default is very bare-bones, so I had to think of something to write! These examples I thought were good, but since the bulk of it was "an example" I kept it in the "main body" of the section so it wasn't too small! But I do agree, they ARE exaples- but the whoel sections seem to be =D

larm-odoo:
@robodoo r+

larm-odoo:
@robodoo r+

Felicious:
@robodoo r+

Felicious:
@robodoo delegate=larm-odoo

PR #14993, opened 2025-10-23
[FIX] Payroll: Updated image with real email address

larm-odoo:
In the current version of the US Localization payroll configuration doc, the email address visible on the employee form is a REAL email address used for test databases, and needs to be changed with a fake email.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5193782) for this PR.


larm-odoo:
Hi @Felicious - this is a no-point PR just to update an image. I did not realize I left my REAL email address in the contact! It was just my test email for testing databases, but still! The fake email was in the personal tab below, which is why I did not realize I left it in the above image.

Felicious:
@robodoo r+

PR #15031, opened 2025-10-27
[I18N] fix Korean po file

ticodoo:
Weblate sometimes incorrectly msgmerges when a line ends with `/""` and wraps one of the `"`'s to the next line causing an invalid po file since the closing `"` is now escaped so the line never correctly ends.

This unfortunately has to be manually fixed directly in the code for now.


ticodoo:
@robodoo r+

Issue #15042, opened 2025-10-27
Example in Chapter 14 is throwing Exception


eximius313:
Chapter 14 gives such example:

<img width="1000" height="456" alt="Image" src="https://github.com/user-attachments/assets/fd0634b2-815a-4bca-af1c-1dd37a735e41" />

But having:
```
    <record id="estate_property_action" model="ir.actions.act_window">
        <field name="name">Properties</field>
        <field name="res_model">estate_property</field>
        <field name="view_mode">kanban,form</field>
    </record>

    <record id="estate_property_view_kanban" model="ir.ui.view">
        <field name="name">estate_property.kanban</field>
        <field name="model">estate_property</field>
        <field name="arch" type="xml">
            <kanban>
                <templates>
                    <t t-name="kanban-box">
                        <div>
                            <field name="name" />
                        </div>
                    </t>
                </templates>
            </kanban>
        </field>
    </record>
```

Throws:
```
Odoo Client Error

UncaughtPromiseError > OwlError
Uncaught Promise > An error occured in the owl lifecycle (see this Error's "cause" property)

Occured on odoo.local on 2025-10-27 19:42:02 GMT

OwlError: An error occured in the owl lifecycle (see this Error's "cause" property)
    Error: An error occured in the owl lifecycle (see this Error's "cause" property)
        at handleError (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:762:101)
        at App.handleError (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:1409:29)
        at ComponentNode.initiateRender (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:854:19)

Caused by: Error: Missing 'card' template.
    at KanbanArchParser.parse (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:9301:70)
    at Object.props (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:9574:1064)
    at View.loadView (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:10079:382)
    at async Promise.all (index 0)
    at async ComponentNode.initiateRender (http://odoo.local/web/assets/f0bc70e/web.assets_web.min.js:853:36)
```

eximius313:
Thank you @AntoineVDV

THDES-odoo:
"kanban-card" seems to be outdated as well, OWL requires "card" now

PR #15050, opened 2025-10-28
[FIX] tutorials,howtos: change all instances of `kanban-box` to `kanban-card`

Mathilde411:
Solves #15042


AntoineVDV (approved):
@robodoo r+

PR #15057, opened 2025-10-30
[IMP] Maintenance: Access rights clarification

jero-odoo:
Docs task:https://www.odoo.com/mail/view?model=project.task&res_id=5208833&access_token=80d65af7-caf8-405a-b2a9-518488ddaf1c

"When a user is not set as an Equipment Manager in their Maintenance access rights, they must be a follower of the equipment to view/access it thanks to the record rule "Users are allowed to access equipment they follow" which makes it so that if a user isn't an Equipment Manager in access rights they can still use the equipment so long as they are a follower of it. To fix this, you need to manually add the user as a follower of the specific equipment they need to maintain."

Updated doc to include section on maintenance access rights and following equipment records.


jero-odoo:
@Felicious this is ready for review. Thanks!

PR #15071, opened 2025-10-31
Update cheat_sheet.rst

fpodoo:
claraification of where to cofigure Invenotry valuation method


fpodoo:
@robodoo r+

Felicious:
We fixed the issues you pointed out, @fpodoo, in the following PR: https://github.com/odoo/documentation/pull/15083

We have to close this one because the branch name didn't start with `19.0-` so the checks will always fail.

Thank you for your input! (:

PR #15073, opened 2025-10-31
[IMP] developer: deprecated attributes in groupby buttons

william-andre:
In order to be more consistent, the `type="edit"` attributes have been removed and the button is forced to be "View" in `.button-secondary`


content/developer/reference/user_interface/view_architectures.rst
aab-odoo, line 1558: I would not link list/button here has our button doesn't support `icon`/`string` attribute. So I would simply write `A special button element...` (no link)

aab-odoo (approved):
Thanks @william-andre

william-andre:
@robodoo r+

PR #15083, opened 2025-10-31
[IMP] update cheat sheet with accounting methods location

erjer-odoo:
clarification of where to configure Inventory valuation method


erjer-odoo:
@robodoo r+

PR #15090, opened 2025-10-31
[IMP] Inventory: make Storage Categories current with 18.0

erjer-odoo:
Making storage categories documentation current with 18.0, requested here:: https://www.odoo.com/odoo/my-tasks/4681301


content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
Felicious: Let's give a better definition of storage categories, something like:


```suggestion
A *storage category* is used with :doc:`putaway rules <putaway>` to assign a storage location to incoming products that considers capacity.
```

(Ugh the wording can be so much better! feel free to refine)

Felicious: since storage categories is available by default
```suggestion
#. :ref:`Enable features in the settings <inventory/routes/enable-storage-categories>`
```

Felicious:
```suggestion
#. :ref:`Define capacity limitations <inventory/routes/define-storage>`
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -27,13 +27,10 @@ Configuration
...
 Then, in the :guilabel:`Warehouse` section, ensure the :guilabel:`Storage Locations` and
-:guilabel:`Multi-Step Routes` features are enabled.
-
-Next, activate the :guilabel:`Storage Categories` feature. Finally, click :guilabel:`Save`.
+:guilabel:`Multi-Step Routes` features are enabled. Click :guilabel:`Save`.
 
 .. image:: storage_category/enable-categories.png
-   :align: center
-   :alt: Show the Storage Categories feature.
+   :alt: Enable "Storage Locations" and "Multi-Step Routes" to enable storage categories.
```
Felicious: no need to format alt tags!
```suggestion
   :alt: Enable Storage Locations and Multi-Step Routes to enable storage categories.
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -27,13 +27,10 @@ Configuration
 
 To enable storage categories, go to :menuselection:`Inventory app --> Configuration --> Settings`.
 Then, in the :guilabel:`Warehouse` section, ensure the :guilabel:`Storage Locations` and
-:guilabel:`Multi-Step Routes` features are enabled.
-
-Next, activate the :guilabel:`Storage Categories` feature. Finally, click :guilabel:`Save`.
+:guilabel:`Multi-Step Routes` features are enabled. Click :guilabel:`Save`.
```
Felicious:
```suggestion
:guilabel:`Multi-Step Routes` features are enabled. 

If intending to set capacities by :ref:`package type <inventory/routes/set-capacity-package>`, also make sure :guilabel:`Packages` is enabled. Click :guilabel:`Save`.
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -69,7 +66,7 @@ product:
   the same time.
 
 .. tip::
-   When clicked, the :guilabel:`Location` smart button shows which storage locations the category
+   When clicked, the :guilabel:`Locations` smart button shows which storage locations the category
```
Felicious: Missing icon! 

Image


```suggestion
   When clicked, the :icon:`oi-arrows-v` :guilabel:`Locations` smart button shows which storage locations the category
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -103,6 +99,11 @@ For companies using :doc:`packages <../../product_management/configure/package>`
 possible to ensure real-time storage capacity checks, based on package types (e.g., crates, bins,
 boxes, etc.).
 
+Create the package type before you can assign it to a storage category. Create it on the
```
Felicious: It seems like you're not walking through the reader how to create a package type. I think that's the right move, to keep this doc focused on storage categories. 

To make sure the reader knows what to do, I recommend adding a link to the doc that goes into how to create them. 

 
```suggestion
Create the :ref:`package type <inventory/warehouses_storage/package-type>` before assigning it to a storage category. Create it on the
```
we generally avoid "you" unless it's necessary, like when there are multiple roles being described (our company setup and a customer's setup). In this case, I think we can achieve brevity without losing clarity when dropping "you"!

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -103,6 +99,11 @@ For companies using :doc:`packages <../../product_management/configure/package>`
...
+:guilabel:`Inventory` tab of the product form (in the :guilabel:`Packaging` section), or create it
+from the :guilabel:`Product Packagings` page. Be sure to set the :guilabel:`Package Type`.
+
+
```
Felicious:
```suggestion
```

extra line

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
Felicious: line ended early when I wrote it before!


```suggestion
   Assigning categories to storage locations tells Odoo these locations meet specific requirements,
   such as temperature or accessibility. Odoo then evaluates these locations, based on defined
   capacity, and recommends the best one on the warehouse transfer form.
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -69,7 +66,7 @@ product:
   the same time.
 
 .. tip::
-   When clicked, the :guilabel:`Location` smart button shows which storage locations the category
+   When clicked, the :guilabel:`Locations` smart button shows which storage locations the category
    has been assigned to.
```
Felicious, line 84: Add an important block explaining the all or nothing rule


```suggestion
.. important::
   Odoo does **not** automatically split quantities across multiple storage locations. If an
   incoming receipt contains several units or packages and the first recommended location exceeds
   its capacity, Odoo still routes all items to that same location instead of selecting another one
   with available space.

   *(Example: If a location can hold 10 units and 12 units arrive, all 12 are still assigned to that
   location.)*
```

Note: typically we put examples in an example admonition block but i think nested admonitions look atrocious. I'm not sure whether parentheses + italics is the best way to format. I'll leave the decision up to you (but I would be willing to die on the hill of no nested admonitions, so if you want to do that, be sure to come with a good reasoning 😂 )

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -103,6 +99,11 @@ For companies using :doc:`packages <../../product_management/configure/package>`
...
 boxes, etc.).
 
+Create the package type before you can assign it to a storage category. Create it on the
+:guilabel:`Inventory` tab of the product form (in the :guilabel:`Packaging` section), or create it
+from the :guilabel:`Product Packagings` page. Be sure to set the :guilabel:`Package Type`.
+
+
 .. important::
    Enable the :guilabel:`Packages` feature in :menuselection:`Inventory app --> Configuration -->
    Settings` to show the :guilabel:`Capacity by Package` tab.
```
Felicious: since we're bringing this up to the enable settings section
```suggestion
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -115,9 +116,21 @@ boxes, etc.).
...
+   To ensure that the storage category works as intended, click the :guilabel:`Put in Pack` button
+   on the receipt form when receiving products into inventory. However, Odoo will treat multiple
+   received packages as a single package if you do this, so you must manually separate the packages
+   into separate storage locations. Click the :guilabel:`Open Move` button to the right of the
+   :guilabel:`Units` field, then click :guilabel:`Add a line`. Finally, split the package into
+   separate locations, then click :guilabel:`Save`. Finally, click :guilabel:`Put in Pack` on the
+   receipt.
+
+   .. image:: storage_category/package-stock-move.png
+      :alt: Complete the "Stock move" box to route packages.
```
Felicious: As discussed, this is procedural instructions that shouldn't be inside an admonition block, which is for conceptual claims. Let's:

1. Move the procedural instructions [down to the "Use case: package type"](https://github.com/odoo/documentation/pull/15090/files#r2491418458)
2. Rewrite the important block to summarize the "all or nothing rule". [link](https://github.com/odoo/documentation/pull/15090/files#r2491365264)

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -166,7 +192,8 @@ Continuing the example from above, the `High Frequency Pallets` storage category
 `PAL1` and `PAL2` locations.
 
 Then, :ref:`putaway rules <inventory/routes/putaway-rule>` are set, so that any pallets received in
-the warehouse are directed to be stored in `PAL1` and `PAL2` locations.
+the warehouse are directed to be stored in `PAL1` and `PAL2` locations. When products are received,
+they are put into a pack before validating the receipt.
 
 Depending on the number of pallets on-hand at each of the storage locations, when one pallet of
 lemonade cans is received, the following scenarios happen:
```
Felicious, line 200: Let's replace this with procedural information, starting on the receipt, to walk the reader through how the pallets are routed PAL1 or PAL2, maybe with screenshots? 

Add the "Put in Pack" step you described above in the [important block(https://github.com/odoo/documentation/pull/15090/files#r2491403933).

Let's also change the pallet of lemonade (u don't have to use the same product if you don't want to), to two or three pallets (scenario continued below).

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
Felicious: It's my belief that these if statements are missing the exception case where we have to split the receipt if the locations are partially full. 

Is there a different way we can write these scenarios that account for that edge case?

And when handling the partially full situation, let's go into the detail you outlined in the [important block](https://github.com/odoo/documentation/pull/15090/files#r2491403933) where we manually click "Put in Pack".

Felicious (changes requested):
Good job testing doing this deep dive, @erjer-odoo ! 

Appreciate the strenuous testing you did and challenging my misunderstandings of the putaway behavior to result in a more accurate document! To make sure your new findings are properly highlighted, I'd like to suggest we move some things around and map out the workflows in even more detail

Let me know if I can clarify any of my comments!

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -144,18 +156,32 @@ With the :ref:`storage category <inventory/routes/define-storage>` and :ref:`loc
...
+Use the :guilabel:`Sublocation` field to specify whether a sublocation should be used and how to
+find that sublocation:
+
+- :guilabel:`No`: No sublocation is used. The destination is whatever is listed in the
+  :guilabel:`Store to` field.
+- :guilabel:`Last Used`: The last location that had a move associated with it for that product or
+  product category is used. If there is no last location used, the destination is whatever is
+  specified in the :guilabel:`Store to` field.
+- :guilabel:`Closest Location`: The locations specified as part of the storage category are used. A
+  storage category is mandatory in the :guilabel:`Having Category` field.
```
Felicious: Is there a Sublocation we MUST set the putaway rule for the storage category to work? This Sublocation definition here seems more like an aspect of a putaway rule rule.

For the purposes of this PR, let's tell the reader if there's a Sublocation they must set to, and in a future PR, we'll move this sublocation definition part out to a future PR. (we wouldn't want to do that right now because your storage categories PR would be merged first, and this crucial info about sublocation doesn't exist anywhere, and we'd like it to exist even if there might be a better place for it.)

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -144,18 +156,32 @@ With the :ref:`storage category <inventory/routes/define-storage>` and :ref:`loc
...
+   Continuing the example from above, the `High frequency pallets` storage category is assigned to
+   the putaway rule directing pallets of lemonade to locations with the `High frequency pallets`
    storage category :ref:`assigned to them <inventory/routes/assign-location>`.
 
    .. image:: storage_category/smart-putaways.png
-      :align: center
       :alt: Storage Categories used in a variety of putaway rules.
 
+If using multiple storage locations for a single category, create putaway rules for each location to
+ensure that if one storage location is in use, the secondary locations must be used.
```
Felicious: A nitpick but I was thinking that this paragraph is a little easy to miss since it's being sandwiched by an image and a heading. Do you think you can move this up to above the image?

erjer-odoo:
Updated with comments from @Felicious .

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -69,8 +70,17 @@ product:
   the same time.
```
Felicious: I ran make review and noticed that this line ends too early. Catching my mistake from 17 months ago 😅 

```suggestion
- :guilabel:`Allow mixed products`: several different products can be stored in this location at the
  same time.
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -162,16 +190,24 @@ Use case: limit capacity by package
...
+- If `PAL1` is partially full (for example, with one pallet), Odoo treats more than one received
+  pallet as a single pallet on the receipt. You must manually separate the two pallets into separate
+  storage locations. Click the :guilabel:`Open Move` icon to the right of the :guilabel:`Units`
+  field, and then in the :guilabel:`Open: Stock move` box, click :guilabel:`Add a line`. Finally,
+  split the receipt by quantity into separate locations, then click :guilabel:`Save`.
+
+     .. image:: storage_category/package-stock-move.png
+        :alt: Update the Stock move box to route pallets to the correct locations before validating.
 
+- If `PAL1` and `PAL2` are full, the pallet is redirected to `WH/Stock/Pallets`.
```
Felicious: (opinion) the last bullet point gets buried and looks easy to miss below the image. It might be a little bit out of order, but what do you think about listing the partially full bullet last, and moving the "full" bullet point above the image?

Image

Felicious (approved):
Great work testing this feature meticulously, @erjer-odoo !

Had super minor things I found and this PR is ready to merge

@robodoo delegate+

Felicious:
I'd typically rate this as a 2pt update since it mainly brings the doc from version 18 to 19 with under 50% of the lines changed. That said, given the depth of research and testing you invested (especially around a complex feature that even the functional expert found challenging :heart_hands: ) I’m increasing the value to 3pts. This reflects the extra effort you put into ensuring accuracy and completeness in the original content!

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
erjer-odoo: What I ended up doing was moving that hidden bullet point above the previous bullet point, so it comes last in the list now.

erjer-odoo:
@robodoo r+

PR #15157, opened 2025-11-06
[IMP] Inventory: update putaway rules documentation

erjer-odoo:
Updating putaway rules documentation to explain "sublocation" and its link to storage categories. Also updated some screenshots to be current with 18.0.


erjer-odoo:
Updated putaway rules documentation with extra information about empty putaway rules and their use, including a best practice.

Felicious (approved):
@robodoo r+

PR #15172, opened 2025-11-07
[IMP] inventory: re-add sendcloud links

Felicious:
Update the Sendcloud integration link to align with Odoo’s native workflow (instead of a 3rd-party app).

Incorporates [feedback from Sendcloud](https://www.odoo.com/odoo/project/3835/tasks/5247599) to include Odoo UTMs in the link to Sendcloud's page, per our partnership agreement to direct Odoo customers to Sendcloud.

Also includes minor formatting updates to meet the 100-character-per-line limit.

### DIFF
--- a/content/applications/inventory_and_mrp/inventory/shipping_receiving/setup_configuration/sendcloud_shipping.rst
+++ b/content/applications/inventory_and_mrp/inventory/shipping_receiving/setup_configuration/sendcloud_shipping.rst
@@ -7,8 +7,8 @@ carriers with Odoo. Once integrated, users can select shipping carriers on inven
 their Odoo database.
 
 .. seealso::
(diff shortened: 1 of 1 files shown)


Felicious:
Hi @larm-odoo ! Could I get your help double checking that all the links I've added work? :D

larm-odoo:
Hi @Felicious - all the links work and take me to the right places, either internally or to the external Sendcloud docs. I did see 2 things I thought I should point out. I see this sentence near the bottom :
image

The parenthesis isn't "attached" to the first part of the sentence, and it appears in a smaller blue ink- I have never seen that kind of format before, so I would just check that.

Second, I noticed all the images are centered- they can be justified left to go with the new standard, but not necessary. 

But everything works, so I am approving - and if you want to change those few things, cool. =)

Felicious:
@robodoo r+

PR #15234, opened 2025-11-11
[IMP] AI: Ai fields edit

jero-odoo:
Update to AI fields doc
The server action that computes AI field only updates empty fields, it does not overwrite existing data. Added a note to correct.


jero-odoo:
@Felicious is making a quick update based on feedback from an expert. Thanks!

PR #15326, opened 2025-11-17
[IMP] essentials/properties: update Property fields doc 19

emmi-odoo:
task-5173906

### DIFF
--- a/content/applications/essentials/property_fields.rst
+++ b/content/applications/essentials/property_fields.rst
@@ -8,12 +8,12 @@ fields allow information storage and management by adding values.
 
 .. admonition:: Property vs. regular fields
 
-   Properties act as pseudo-fields; they behave like regular fields but are not saved as columns in
-   the database. They also rely on a defined :ref:`parent record <property-fields/properties-apps>`.
+  Properties act as pseudo-fields; they behave like regular fields but are not saved as columns in
+  the database. They also rely on a defined :ref:`parent record <property-fields/properties-apps>`.
 
-   .. example::
-      Adding a property to a *task* inserts a field in *all tasks* within the *same
-      project* while other projects' tasks remain unaffected.
+  .. example::
+     Adding a property to a *task* inserts a field in *all tasks* within the *same
+     project* while other projects' tasks remain unaffected.
 
 .. _property_field/add:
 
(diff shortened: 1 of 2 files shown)


emmi-odoo:
Hello @xpl-odoo, here is an update on the Property field documentation. I tried to define the "Suffix" field as thoroughly as possible. I'm not sure if it needs that much explanation. Regarding the "Model" column, the Contract form has recently been merged with the Employee form; that's why I wrote "Employee/Contract" for Payroll. Let me know if I simply need to write "Employee" instead. 

Thanks for the review ;)

content/applications/essentials/property_fields.rst
```diff
@@ -27,87 +27,96 @@ In the popover, enter the property's :guilabel:`Label`, choose a :guilabel:`Fiel
...
+      Select a :guilabel:`Default Value` if desired.
+  * - :guilabel:`Separator`
+    - Group several properties under a foldable label
+    -
 
 .. note::
-   - Select whether to display the property in the Kanban, List, or Calendar views cards for every
-     field with the :guilabel:`Display in Cards` option.
+   - Enter a :guilabel:`Suffix` to specify the contextual description that is automatically
+     appended to the displayed value of the field type. For example, to indicate cost per
```
xpl-odoo:
```suggestion
   - Enter a :guilabel:`Suffix` to add a contextual description after a
   field's value. For example, to indicate the cost per
```
Could be simplified IMO

content/applications/essentials/property_fields.rst
```diff
@@ -27,87 +27,96 @@ In the popover, enter the property's :guilabel:`Label`, choose a :guilabel:`Fiel
...
+     kilometer, select the :guilabel:`Monetary` field type an name it **Cost**, then enter **per
+     km** in the :guilabel:`Suffix` field. The property displays the following: **Cost** [added
+     value] **per km**.
```
xpl-odoo:
```suggestion
   - Enter a :guilabel:`Suffix` to add a contextual description after a
     field's value. For example, to indicate the cost per
     kilometer, you could use a :guilabel:`Monetary` field and name it **Cost**, then enter **per
     km** as the :guilabel:`Suffix`. The property will display the following: **Cost** [
     value] **per km**.
```

content/applications/essentials/property_fields.rst
```diff
@@ -27,87 +27,96 @@ In the popover, enter the property's :guilabel:`Label`, choose a :guilabel:`Fiel
...
+   - Enable :guilabel:`Display in Cards` to select whether to display the property in the Kanban,
+     List, or Calendar views cards for every field.
```
xpl-odoo:
```suggestion
     List, or Calendar views' cards for every field.
```

xpl-odoo (approved):
@robodoo r+

PR #15346, opened 2025-11-18
[IMP] crm: revise automatic rule based lead assignment

meval1006:
Revised the Configure rule-based assignment section to include time period intervals. This is due to a certification question for Odoo 19.

Project task link: https://www.odoo.com/odoo/action-4043/5243891


Felicious:
@robodoo r+

PR #15351, opened 2025-11-18
[IMP] Sales: adding/deleting variants clarification

theRealThagomizer:
Hiya, Felicia! Got a 2 pt. PR related to the certification questions. This PR adds a new section to the Product Variants page that explains what happens when attributes are added or deleted to existing variants.

### DIFF
--- a/content/applications/sales/sales/products_prices/products/variants.rst
+++ b/content/applications/sales/sales/products_prices/products/variants.rst
@@ -300,5 +300,39 @@ impacts that can be taken advantage of throughout the Odoo database.
 .. note::
    Changes to the product template automatically apply to every variant of that product.
 
+Adding and deleting attributes from product variants
+====================================================
+
+When attributes are added or deleted from products, variants that have already been created are
+affected. Depending on whether or not these variants have been used in orders, they are archived or
+deleted and recreated.
+
+- Variants that have not been used in an order are deleted and recreated with the new attributes.
+  This may result in the deletion of additional customizations, depending on how the variants were
+  configured.
+- Variants that have been previously used in one or more orders are archived and removed from the
+  product catalog instead of being deleted.
+
+.. example::
+   A business offers leather motorcycle jackets with the following attributes and values for a total
+   of 27 product variants:
+
+    - Size: Small, Medium, Large
+    - Color: Black, Brown, Natural
+    - Material: Leather, Vegan Leather, Premium Leather
+
+   Due to rising costs, the business makes the decision to stop offering different leathers and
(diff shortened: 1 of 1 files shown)


content/applications/sales/sales/products_prices/products/variants.rst
```diff
@@ -299,5 +299,18 @@ impacts that can be taken advantage of throughout the Odoo database.
 .. note::
    Changes to the product template automatically apply to every variant of that product.
 
+Adding and deleting attributes from product variants
+====================================================
```
Felicious, line 304: Really cool troubleshooting section and documenting an edge case! To better conceptualize the impact, let's use an example. Start with some attributes with multiple values, add 1 attribute value to an order, then delete the attribute, and show that the attribute value that was never used gets deleted and the value that was used gets archived (and specify where)

content/applications/sales/sales/products_prices/products/variants.rst
```diff
@@ -300,5 +300,34 @@ impacts that can be taken advantage of throughout the Odoo database.
...
+
+- Variants that have not been used in an order are deleted and recreated with the new attributes.
+  This may result in the deletion of additional customizations, depending on how the variants were
+  configured.
+- Variants that have been previously used in one or more orders are archived and removed from the
+  product catalog instead of being deleted.
+
+.. example::
+   A business offers leather jackets with a variety of customization options: size, color, and
+   material. They sell standard leather jackets and limited edition leather jackets as two different
```
Felicious: Nitpicking here so we're really precise with the language about what an attribute vs value is


```suggestion
   A business offers leather jackets with a variety of customization attributes: size, color, and
   material. Each attribute has multiple values: for size, there is S, M, L. They sell standard leather jackets and limited edition leather jackets as two different
```

content/applications/sales/sales/products_prices/products/variants.rst
```diff
@@ -300,5 +300,34 @@ impacts that can be taken advantage of throughout the Odoo database.
...
+   products. The standard leather jackets are a best-selling item, but the limited edition jackets
```
Felicious: What do you mean that "They sell standard leather jackets and limited edition leather jackets as two different products"? In Odoo, we can set different prices for product variants, but this sentence makes it sound like there are two separate products, which would be a concept different from variants. Could you clarify what you mean here?

Felicious (approved):
@robodoo delegate=theRealThagomizer

theRealThagomizer:
@robodoo r+

PR #15370, opened 2025-11-19
[IMP] Purchase: remove blanket order field

jero-odoo:
docs task:https://www.odoo.com/odoo/project.task/5266493

Removing section about blanket order field (can no longer edit blanket order options from the "blanket order field"
also updated images/UI from the rest of the doc to match v18


jero-odoo:
@Felicious this is a quick update, sending to you for final/tech and merge. Thanks!

content/applications/inventory_and_mrp/purchase/manage_deals/blanket_orders.rst
Felicious, line 96: Maybe add here that total quantities ordered through linked RFQs are automatically updated in this Ordered field?

Image

jero-odoo:
@robodoo r+

PR #15374, opened 2025-11-20
[IMP] reference/cli: db-filter for Web UI

iCart:
The odoo-bin help for the db-filter says:
`Regular expressions for filtering available databases for Web UI.`

However, the current documentation does not mention that the filters
only affect the Web UI, and not the cron workers.


Feyensv (commented):
The 16 doc is not built anymore (https://www.odoo.com/documentation/16.0/last_build.html)
Please target 17.0 instead :eyes:

PR #15464, opened 2025-11-26
[FIX] Payroll: Missing attachment info

larm-odoo:
MUBA discovered missing info in the salary attachment doc. Adding the info for the bullet point:

- :guilabel:`Negative Amount`: Tick this checkbox if the salary attachment is a negative value.

This sentence was incomplete/cut off in the original doc.

[Task card](https://www.odoo.com/odoo/project/3835/tasks/5344269) for this PR. This is ONLY for version 18 - there is a 19 update already in Github, being reviewed.


larm-odoo:
Hi @StraubCreative - I am tagging you for a one-and-done review, since this is only to add the missing 4 words in a sentence, and basically nothing else changed! Only 1 point, so it's a fix more or less.

larm-odoo:
@robodoo r+

PR #15530, opened 2025-12-01
[IMP] db: odoo accounts update

xpl-odoo:


content/administration/odoo_accounts.rst
```diff
@@ -1,80 +1,42 @@
...
 
-To delete an Odoo.com account, start by clicking the profile icon in the upper-right corner
-(represented by the username and icon) to reveal a drop-down menu. From the drop-down menu, select
-:guilabel:`My Odoo.com account`, which reveals the user portal.
+  .. note::
+     - Changing your Odoo.com account's password also modifies the password used to sign in on the
+       database(s) you created with this account.
+     - You can set a specific password for your user to access a database by going to the
+       :ref:`database's settings <users/change-password>`; however, you will still be able to sign
+       in with your Odoo.com account password.
```
auva-odoo: I'm not sure I understand this: is "your user" the user that's automatically created for the person who creates a db? (sorry not sure about this!). And if so, would that mean that that person would be able to sign into a db using two different passwords?

content/administration/odoo_accounts.rst
```diff
@@ -1,80 +1,42 @@
...
 
-From the user portal, the delete option can be accessed by going to :menuselection:`My Account -->
-Edit Security Settings --> Delete Account`. It can also be accessed by going to
-`https://www.odoo.com/my/home <https://www.odoo.com/my/home>`_.
+- Enable **two-factor authentication** using an authenticator app.
```
auva-odoo: Just my personal opinion but I think we could give a bit more information about the steps, like we did in the old version of this doc.

Also, should we mention Trusted devices (and how you can remove them) and the "Revoke all" button --> what does that do exactly? does it remove all trusted devices?

content/administration/odoo_accounts.rst
```diff
@@ -1,80 +1,42 @@
...
+- **Log out** from all devices.
+- **Delete** your account.
```
auva-odoo: I personnally don't see this option; why? when would someone not be able to delete their account exactly?

content/administration/odoo_accounts.rst
```diff
@@ -1,80 +1,42 @@
...
 
-.. danger::
-   Deleting an Odoo account is irreversible. Be careful performing this action, as the Odoo.com
-   account is **not** retrievable once deleted.
+  .. danger::
+     Deleting your Odoo.com account is **irreversible**. Once deleted, your account **cannot be
+     retrieved**.
```
auva-odoo, line 40: and what about the Developer API key section? Should we at least refer to the documentation link provided on the page?

content/administration/odoo_accounts.rst
```diff
@@ -1,80 +1,42 @@
...
+     in to the database(s) you created with this account.
+   - You can set a specific email address for your user in a database by clicking your avatar in the
+     top-right corner and selecting :guilabel:`My Preferences`; however, you will still need to
+     sign in to the database using the email address used at the time of the database
+     creation.
+   - To change the email address used to sign in to a database, `contact Odoo Support
+     <https://www.odoo.com/help>`_.
+
+.. tip::
+   Creating an account on a database does not create an Odoo.com account. If the email address of
```
auva-odoo: By "account on a database", do you mean a user? sorry i'm confused! 🙈

content/administration/odoo_accounts.rst
xpl-odoo: "your user" indeed means the user that's automatically created for the person who creates the DB (and that user is linked to the person's Odoo.com account). I tried to clarify.
Yes, you can then log in to a database using two different passwords (with your Odoo.com password or with the one set on the DB, if any). It's weird, and I'm not sure I should mention it, but it's the (confusing) way it works.

content/administration/odoo_accounts.rst
xpl-odoo: As we have a rather in-depth documentation on 2FA (at database-level), I think a seealso could work here.
I don't see anything about Trusted devices here. Mind sharing where/how you find that? (maybe I'm just blind :sweat_smile: )

content/administration/odoo_accounts.rst
xpl-odoo: I believe it's only disabled for Odoo employees.

content/administration/odoo_accounts.rst
xpl-odoo: I didn't think dev mode would impact that page - thanks for pointing it out! After investigating, it's a relatively niche use (to download DB dumps) according to Antoine, hence why it's "hidden" under dev mode.
Still, I'll mention it and use the same link provided on the page, but I don't think it's necessary to explain further.

content/administration/odoo_accounts.rst
xpl-odoo: You can create a (customer/portal user) "account" on any DB (i.e., https://dbname.odoo.com/web/signup) by default. However, adding that information here may be superfluous/confusing, so I'll rephrase here :)

content/administration/odoo_accounts.rst
auva-odoo: Yeah you're right, a seealso is sufficient! As far as the trusted devices go, I think it's just because I have enabled one for my account (probably the "don't ask again on this device" option or something similar).

content/administration/odoo_accounts.rst
auva-odoo: That makes sense :sweat_smile:

content/administration/odoo_accounts.rst
auva-odoo: I'm not sure I understand because I don't think I enabled dev mode and I could still see this section? (I just tried again and was able to see it)

xpl-odoo: I believe it's visible by default for Odoo employees, but that's not the case for non-Odoo employees :)

xpl-odoo:
@robodoo r+

auva-odoo:
@robodoo r+

PR #15571, opened 2025-12-04
[IMP] admin: missing header with X-Accel-Redirect

Julien00859:
Nginx doesn't set the Content-Security-Policy and X-Content-Type-Options headers on the response it sends to the browser even though they were present on the response from the Odoo server.


Julien00859:
@robodoo r+

PR #15573, opened 2025-12-04
[IMP] developer: new schedule attribute for calendar views

juliusc2066:
This commit adds documentation support for the new calendar view attribute introduced in https://github.com/odoo/odoo/pull/238557

task-5364034


AntoineVDV (approved):
The diff LGTM, but the commit title doesn't make sense with regard to the documentation repo. Something like "[IMP] developer/view_architectures: new `schedule` attribute for calendar views" would be more appropriate.

@robodoo delegate+

juliusc2066:
@robodoo r+

PR #15580, opened 2025-12-05
[IMP] Belgium: Explain Peppol EAS requirements

antoine162:
task-none

Starting from 1 January 2026, companies in Belgium need to be registered on the Peppol network. Specifically, even though there are two EASs for Belgium, which are:
  - 0208 - company registry number;
  - 9925 - Belgian VAT;

all Belgian companies must be registered under scheme 0208 with their Belgian company number.

Scheme 9925 is optional which means companies can be registered both on 0208 and 9925. But in any case, they need to be registered on 0208.

We add an `important` block in the Belgium page to explain this requirement.

### DIFF
--- a/content/applications/finance/fiscal_localizations/belgium.rst
+++ b/content/applications/finance/fiscal_localizations/belgium.rst
@@ -497,15 +497,27 @@ via Codaclean are checked every twelve hours.
    To manually check for new CODA files, go to the Accounting Dashboard and click :guilabel:`Fetch
    from Codaclean` on the relevant :ref:`journal <belgium/codaclean-configuration-journal>` .
 
-Electronic invoicing
-====================
+.. _belgium/peppol:
 
-Odoo supports the **Peppol BIS Billing 3.0 (UBL)** electronic invoicing format. To enable it for a
-customer, go to :menuselection:`Accounting --> Customers --> Customers`, open their contact form,
-and under the :guilabel:`Accounting` tab, select the :guilabel:`Peppol BIS Billing 3.0` format.
+Electronic invoicing with Peppol
+================================
+
+As of 1 January 2026, all Belgian companies must be registered on the Peppol network to send and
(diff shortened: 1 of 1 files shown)


antoine162:
(sending this straight to be-doc-review as Dallas is off next week :smile:)

content/applications/finance/fiscal_localizations/belgium.rst
```diff
@@ -500,10 +500,27 @@ via Codaclean are checked every twelve hours.
 Electronic invoicing
 ====================
 
-Odoo supports the **Peppol BIS Billing 3.0 (UBL)** electronic invoicing format. To enable it for a
+As of 1 January 2026, all Belgian companies must be registered on the :ref:`Peppol
+<accounting/e-invoicing/peppol>` network to send and receive electronic invoices.
+
```
xpl-odoo, line 507: I would make it clearer that the info presented below is provided in addition to the instructions of the Peppol documentation. Something like:

```restructuredtext
To activate Peppol, refer to the :ref:`Peppol documentation <accounting/e-invoicing/peppol>`, 
ensuring during :ref:`registration <accounting/e-invoicing/peppol-registration>` that: 
- the :guilabel:`Peppol EAS` setting is set to :guilabel:`0208 - Numero d'entreprise /
  ondernemingsnummber / Unternehmensnummer`
- the :guilabel:`Peppol Endpoint` setting is set to the company registry number

In addition, when :ref:`enabling Peppol for a customer <accounting/e-invoicing/contact-verification>`,
ensure the :guilabel:`Peppol BIS Billing 3.0` format is selected.
```

What do you think?

antoine162: That's a lot simpler and clearer, thanks so much! ❤️

xpl-odoo:
@robodoo r+

PR #15598, opened 2025-12-08
[IMP] Appointments: resource capacity note

jero-odoo:
Docs task: https://www.odoo.com/odoo/project.task/5347837

Adding a note regarding Resource capacity

Essentially, no matter what the capacity is on a resource, customers can only book the resource for a max of 12 people on the website. (This will be raised to 20, but is crrently 12) The workaround is to add a new system parameter to raise the limit.


jero-odoo:
@erjer-odoo this is ready for review. Thanks!

jero-odoo:
@Felicious this is ready for review. Thanks!

Felicious:
@robodoo r+

PR #15601, opened 2025-12-09
[I18N] *: re-export pot files

ticodoo:


dylankiss:
@robodoo r+

AntoineVDV:
@robodoo delegate+

Maybe a stupid question, but why don't we automate this?

dylankiss:
@robodoo r+

ticodoo:
@AntoineVDV we've talked about it, but it requires modifications/custom code of what we already use for the pot file export automation + we'll need to coordinate with the runbot team about having a (nightly?) build that also exports the pot files (maybe it could be part of the build that also builds all the languages? :wink:)

So basically, it's on the list, but hasn't been a high priority since it's pretty fast to re-export the pot files. It's just slightly less fast now that Dylan needs to r+ the PRs

PR #15609, opened 2025-12-09
[IMP] Payroll: Adding 8 new states to US Payroll user doc

hun-odoo:
Adding 8 new states to the US Payroll user documentation. We are expanding the scope for US Payroll and will be working on additional states, and the documentation needs to be updated as we merge new states.


hun-odoo:
Closing this since my branch name does not follow the required rules. Will create another PR

PR #15654, opened 2025-12-11
[IMP] general: network configuration recommendation

loouis-t:
We add a network security recommendation to tell clients not to expose their IoT Box on the internet.

### DIFF
--- a/content/applications/general/iot/iot_box.rst
+++ b/content/applications/general/iot/iot_box.rst
@@ -14,6 +14,11 @@ To start using an IoT box:
    Devices can also be connected after the IoT box is added to the network and/or connected to the
    database; however, a reboot of the IoT box might be required.
 
+.. important::
+   Never make the IoT box accessible from the public Internet.
+   IoT systems are designed to provide your database with access to devices on your local network.
+   Exposing them to the public Internet would create a **security risk**.
+
 .. _iot/iot_box/network:
 
 Network connection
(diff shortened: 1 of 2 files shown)


content/applications/general/iot.rst
```diff
@@ -23,6 +23,12 @@ The following IoT systems are supported:
    - It is also possible to create a Windows Virtual Machine on a MacOS/Linux computer. However,
      this option is not supported by Odoo, and no troubleshooting assistance will be provided.
 
+.. important::
+   Never expose your IoT system directly to the Internet.
+
+   IoT Systems are designed to provide your database an access to your local network,
+   exposing them to the Internet would create a security risk.
```
auva-odoo:
```suggestion
   Never make the IoT system accessible from the public Internet.

   IoT Systems are designed to provide your database with access to devices on your local network. Exposing them to the public Internet would create a security risk.
```

loouis-t:
> Thank you for this PR @loouis-t! I think the important note would have more impact in the iot_box.rst (l28) and windows_iot.rst (l12 maybe?) files individually, even if that means duplicating that information, and change _IoT system_ accordingly. I would also rephrase slightly to clarify what we mean exactly as the phrasing could be a bit obscure to some people? (see my suggestion). What do you think?

Hello, thanks for you feedback, it makes sense, I updated my PR!

PR #15674, opened 2025-12-11
[IMP] Payroll: Adding 8 new states to US Payroll user doc

hun-odoo:
Adding 8 new states to the US Payroll user documentation. We are expanding the scope for US Payroll and will be working on additional states, and the documentation needs to be updated as we merge new states.


Felicious:
The changes look good to me, @hun-odoo !  Merging now :) 

@robodoo r+

Congrats on your first, direct PR contribution to the documentation!! You've collaborated a lot with us in the past on Manufacturing docs, but I hope it's been just as fun authoring something!! :blush:

PR #15698, opened 2025-12-12
[MOV] Barcode: Move Barcode Lookup to setup section

erjer-odoo:
After a brief discussion with Felicia, we have decided to move the Barcode Lookup documentation from the Integrations section to the Supply Chain > Barcode > Setup section of our documentation. I updated our Supply Chain "Barcode lookup" section to link to the new location. I've added a redirect for this in the 18.0 redirects file. Finally, I updated one other document that used to link to this in its former location to use the new location, ensuring it builds correctly.


Felicious:
Hi @odoo/ecommerce-doc-review and @odoo/general-doc-review, we determined that it didn't quite make sense for this Barcode Lookup doc to exist in the general scope. As written, Barcode Lookup fills in product form details, which theoretically means it could be applied in other contexts like Events and PoS. However, as written, this document doesn't contain that context to fill in attendee info, so we're having trouble envisioning how it can be used in non-supply chain contexts.

For the sake of putting this document near similar docs, we chose to move it. If PoS writers determine that they need this document to exist here in the general scope for a reason outside of what we currently know, let's discuss and see whether we can revise this document to include other app contexts.

PR #15713, opened 2025-12-13
Add GitHub Actions workflow for Markdown conversion

UCNorg:
This workflow automates the conversion of various document formats to Markdown using Pandoc. It includes steps for checking out the repository, installing Pandoc, running a conversion script, and committing the converted files back to a branch.


UCNorg:
1

PR #15741, opened 2025-12-16
[FIX] accounting: remove LIFO as inventory valuation option

dade-odoo:
quick removal as LIFO is not a valid inventory valuation option


auva-odoo (approved):
@robodoo r+

PR #15800, opened 2025-12-19
[FIX] *: make several untranslatable parts translatable

dylankiss:
Some parts of the documention were currently not translatable:
- the custom `Card` directives
- the strings in the layout HTML pages
- the placeholder text in the search box
- the "Edit on GitHub" text on each page

This commit makes sure that all of these are translatable now.


ticodoo:
LGTM, Dylan tested that the translation correctly loads and the exported terms match what's missing. Also, TIL, sphinx doesn't export the msgids in alphabetical order, but in order of appearance in the code.

@robodoo rebase-ff r+

PR #15879, opened 2026-01-05
[IMP] howtos/website_themes: adapt doc to new website builder API

duau-odoo:
Following the website refactoring [1], this documentation is obsolete.

task-5258016

[1]: https://github.com/odoo/odoo/commit/9fe45e2b7ddbbfd0445ffe25a859e67a316d02b2


content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,27 +20,28 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
-   │       └── s_snippet_name
-   │           └── 000.js
-   │           └── 000.scss
-   │           └── 000.xml
-   │           └── options.js
+   └── src
+       ├── snippets
+       │   └── s_snippet_name
+       │       ├── 000.scss
+       │       └── 000.xml
```
robinlej, line 42: `000.js` files were replaced by `snippet_name.js` (and possibly `snippet_name.edit.js`), they didn't disappear

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,27 +20,28 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
+       └── builder
+           ├── s_snippet_name_option.js
+           └── s_snippet_name_option.xml
```
robinlej: That's not entirely true :sweat_smile: 
In `/website`, it's `/builder/plugins/options/*`, with 
- `snippet_name_option.js`
- `snippet_name_option.xml`
- `snippet_name_option_plugin.js`
- and sometimes `snippet_name_option.edit.scss`

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +562,439 @@ instead:
...
 
-.. _website_themes/building_blocks/custom/options/binding/data_selector:
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :widths: 30 70
 
-data-selector
-*************
+   * - Member
```
robinlej: `Property` rather

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +562,439 @@ instead:
...
-interact with the user interface.
+.. _website_themes/building_blocks/custom/options/binding:
 
-`data-selector` binds all the options included in the group to a particular element matching the
-selector value (CSS class, ID, etc). The option will appear when the matching selector is selected.
+Binding
+~~~~~~~
 
-.. code-block:: xml
+Binding relies on CSS selectors (class, tag, ID, etc.) to decide when an option appears.
```
robinlej: I find this sentence confusing, the previous sentences were clearer
Edit: it's a bit repetitive with what comes just after

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +562,439 @@ instead:
...
-   <div data-selector="section, h1, .custom_class, #custom_id">
+.. _website_themes/building_blocks/custom/options/binding/selector:
 
-It can be used in combination with other attributes like `data-target`, `data-exclude` or
-`data-apply-to`.
+selector
+********
 
-.. _website_themes/building_blocks/custom/options/binding/data_target:
+Option components declare selector metadata that controls when they appear.
```
robinlej: > declare selector metadata

"declare a selector property"?

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +562,439 @@ instead:
...
-   <we-range
-      string="Images Spacing"
-      data-select-class="o_spc-none|o_spc-small|o_spc-medium|o_spc-big" />
+   <BuilderRange
+      action="'setClassRange'"
+      actionParam="['o_spc-none', 'o_spc-small', 'o_spc-medium', 'o_spc-big']"
+      max="3" />
 
-Each step of the range is separated by a `|`. Here, each class name corresponds to a step.
+Each step of the range maps to one entry in the `actionParam` array.
```
robinlej: That's only true in the specific context of non-numbered actions. Most `BuilderRange` take a min and a max. Shouldn't the basic implementation be shown first?

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -925,160 +1078,130 @@ Formats the option as a color/gradient to choose from.
...
-         //...
-      }
-   });
-
-   options.registry.AirproofSnippet = AirproofSnippet;
-
-   export default AirproofSnippet;
-
-Finally, the custom method can be called on your custom option through the XML template:
+   :caption: ``/website_airproof/static/src/builder/airproof_snippet_option.js``
```
robinlej: `airproof_snippet_option_plugin.js`

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +562,439 @@ instead:
...
+
+Inner content selectors
+***********************
+
+To make a custom snippet "inner content" (droppable in another building block), add its selector
+to the `so_content_addition_selector` resource, which contains all selectors for existing inner
+content building blocks.
+
+.. code-block:: javascript
+   :caption: ``/website_airproof/static/src/builder/airproof_snippet_option.js``
```
robinlej: `airproof_snippet_option_plugin.js`

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +562,439 @@ instead:
...
+   import { registry } from "@web/core/registry";
 
-Then insert the different available options:
+   export class AirproofSnippetOption extends BaseOptionComponent {
+       static template = "website_airproof.AirproofSnippetOption";
+       static selector = ".s_airproof_snippet";
+       static applyTo = ":scope > .row";
+   }
+
+   class AirproofSnippetOptionPlugin extends Plugin {
```
robinlej: We mostly use 2 different files for those, don't we?

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: In most of the `website_*` modules, it's usually under `/website_builder/`, should I put that instead? (I checked on `website_forum`, `website_event` and `website_sale`)

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: Before the [refactoring of builder options](https://github.com/odoo/odoo/commit/36f741745b927d7cbac9b32042bd53b47711da6f) yes, now I checked, it's 50-50 : 118 `BaseOptionComponent` are in a `*plugin.js` file, and 100 aren't. Maybe we need a clear "policy" ?

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,27 +20,30 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
+       │       ├── snippet_name.js
+       │       ├── 000.scss
+       │       └── 000.xml
+       └── website_builder
+           ├── snippet_name_option.js
+           ├── snippet_name_option.xml
+           └── snippet_name_option_plugin.js
 
 .. seealso::
    `XML templates of the different snippets
```
robinlej: It links to an old commit. The most recent one is https://github.com/odoo/odoo/blob/8fe9109b8915c940998108ded22960eb699c1f4b/addons/website/views/snippets/snippets.xml

But it's also not the file with "templates of the different snippets" anymore, it's just the XML registry of existing snippets and groups.

I'm also under the impression that it would be nice to add a small info box explaining in a few words that, even if we're not going to explain how it works on this page, the JS files under `/snippets` are interactions

> see also an [implementation example with the s_floating_blocks snippet](https://github.com/odoo/odoo/blob/dd148d22a03e27c0687f106aeb5179741e8f1e5e/addons/website/static/src/snippets/s_floating_blocks/floating_blocks.js) and the [base Interaction class](https://github.com/odoo/odoo/blob/78fdef11cb79a40222e26c3a1f9971c0b45cb000/addons/web/static/src/public/interaction.js)

bram-odoo:
Hey there, thanks for this hard work :-) Just to set the context, we did not write the 19.0 theming documentation already. All parts that hadn't be modified by @duau-odoo are related to the 18.0. We are currently working on the new version and this PR will obviously help us a lot :-)

content/developer/howtos/website_themes/gradients.rst
bram-odoo, line 1: I can't seem to find it but I remember reading something in the official documentation mentioned that `patch`should be used with cautious as a final option (meaning: if you can extend it, prefer that way). But I guess things evolved since then. 

When I upgraded our `website_airproof` module, I tried to avoid using patch as it was mentioned that could be risky. Don't know if my code is ok or if we should go for yours. Let me know 😄

https://github.com/odoo/tutorials/blob/992dbdfc097afd80b9c63a903f45214468542028/website_airproof/static/src/js/html_editor/main/font/color_picker_gradient_tab.js

content/developer/howtos/website_themes/shapes.rst
```diff
@@ -270,26 +270,44 @@ Define the styles of your shape.
...
+`background_shape_groups_providers` resource.
+
+.. code-block:: javascript
+   :caption: ``/website_airproof/static/src/builder/airproof_background_shapes.js``
+
+   import { Plugin } from "@html_editor/plugin";
+   import { _t } from "@web/core/l10n/translation";
+   import { registry } from "@web/core/registry";
+
+   export class AirproofBackgroundShapesPlugin extends Plugin {
```
bram-odoo, line 283: Initially I was just adding a subgroup to the "Basic" one. So I adapted my code to create a separate main group + a subgroup inside of it. I did it a bit more directly: 

```javascript
import { backgroundShapesDefinition } from "@html_builder/plugins/background_option/background_shapes_definition";
import { _t } from "@web/core/l10n/translation";

backgroundShapesDefinition.airproof = {
    label: _t("Airproof"),
    subgroups: {
        aiproof: {
            label: _t("Subgroup #1"),
            shapes: {
                "website_airproof/waves/01": {
                    selectLabel: _t("Waves 01")
                },
            },
        }
    }
};
```
What do you think about it?

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,31 +20,38 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
+The view templates live under `views/snippets/`:
 
 ::
 
    views
-   ├── snippets
-   │   └── options.xml
-   │   └── s_snippet_name.xml
+   └── snippets
+       ├── snippets.xml
```
bram-odoo, line 29: Initially we named this file "options.xml" as it contains all our "global" options not only code related to the building blocks (header, footer, etc) (+ the redundancy induced by the parent folder itself: `snippets/snippets/xml`)

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,31 +20,38 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
-   │   └── snippets
-   │       └── s_snippet_name
-   │           └── 000.js
-   │           └── 000.scss
-   │           └── 000.xml
-   │           └── options.js
+   └── src
+       ├── snippets
+       │   └── s_snippet_name
+       │       ├── snippet_name.js
```
bram-odoo, line 40: Compared to the older naming convention, does it mean that the "versioning" system has changed also?

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,31 +20,38 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
+       │       ├── 000.scss
+       │       └── 000.xml
+       └── website_builder
```
bram-odoo, line 43: We have many discussions about the correct name to use to be... consistent with the standard. In the previous versions we use to "group" the "front" files (scss, js and XML) and the options in the `/snippets/s_snippet_name` folder. In 19, the structure is a bit different (kinda confusing sometimes tbh, it's sometimes really difficult to locate the right file), sometimes we override things from the `/html_builder` sometimes the `/html_editor` and both modules have their own structure that are "same but different" if you know what I mean ^^". 

So here, if I take the "Badge" standard snippet as an example, the options are located in : 
`/html_builder/static/src/plugins/badge_option_plugin.js` 
so the override in our custom module would be : 
`/static/src/html_builder/plugins/badge_option_plugin.js`

If we use the convention of the Website application, we should use then: 
`/static/src/builder/plugins/my_file.js` or `/static/src/builder/plugins/options/my_file.js` (why are they some in `plugins/` and others in `/plugins/options/`? No idea at all

😅

So back to your proposition, I would stick to the Website convention and use "builder" as name folder (I would also use a subfolder per snippet but I could live with a flat folder also.

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +568,448 @@ instead:
...
-   <template id="s_airproof_snippet_options" inherit_id="website.snippet_options" name="Airproof - Snippets Options">
-      <xpath expr="." position="inside">
-         <!-- Options -->
-      </xpath>
-   </template>
+   import { BaseOptionComponent } from "@html_builder/core/utils";
+   import { Plugin } from "@html_editor/plugin";
+   import { registry } from "@web/core/registry";
+
+   export class AirproofSnippetOption extends BaseOptionComponent {
```
bram-odoo, line 627: I did understand how Plugin and BuilderAction works (at least partially) but I couldn't figure out what was the point with the `BaseOptionComponent` as I saw we could specify the template, selector and other things directly into the `Plugin` : 

https://github.com/odoo/tutorials/blob/19.0-5237455-website_airproof-bram/website_airproof/static/src/js/html_builder/plugins/options/s_airproof_carousel_option_plugin.js

(My apologizes for this kind of questions but we were thrown in the pit of this new JS environment and had to understand how is works by ourselves 😅)

content/developer/howtos/website_themes/shapes.rst
duau-odoo: We looked into this as well, but mutating `backgroundShapesDefinition` directly is risky.
It’s a shared object, so when you open the website builder it modifies the definition in memory, and those changes persist for the whole JS session. If you then open mass mailing without a full reload, it reuses that same state and can see website-specific shapes even though they were never meant to be available there.

But to be honest, we stumbled onto this use case while preparing this PR, so the resource was [added very recently](https://github.com/odoo/odoo/pull/243121)

content/developer/howtos/website_themes/gradients.rst
duau-odoo: Indeed, we should still be cautious with `patch` ! I agree with your solution, only nitpick is that I find it cleaner to fully remove the object from the registry and add it back, rather than replacing it in place:
```diff js
const colorPickerTabs = registry.category("color_picker_tabs");
-const gradientEntry = colorPickerTabs.get("html_editor.gradient");

-gradientEntry.component = AirproofColorPickerGradientTab;
-colorPickerTabs.category["html_editor.gradient"] = gradientEntry;

+colorPickerTabs.remove("html_editor.gradient");
+colorPickerTabs.add(
+    "html_editor.gradient",
+    {
+        id: "gradient",
+        name: _t("Gradient"),
+        component: AirproofColorPickerGradientTab,
+    },
+    { sequence: 60 }
+);
```

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -559,330 +568,448 @@ instead:
...
-data-drop-in
-************
+`exclude` removes specific selectors from the match.
 
-`data-drop-in` defines the list of elements where the snippet can be dropped into.
+.. code-block:: javascript
+   :caption: The option appears if an `<ul>` tag (without `.navbar-nav` class) is selected
 
-.. code-block:: xml
+   export class AirproofSnippetOption extends BaseOptionComponent {
```
bram-odoo, line 815: As we use CSS selectors, could this notation be correct? 

```javascript
static selector = "ul:not(.navbar-nav)";
```

Or does it impact the priority of the rule?

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -891,30 +1018,71 @@ Formats the option as a text field.
...
+`<BuilderList>`
+***************
+
+Render and edit a list of items (add, remove, reorder) in the options panel. Useful for prefilled
+values, custom options, or settings that map to a list of objects.
+
+.. code-block:: xml
+   :caption: Simple list with two fields
+
+   <BuilderList
```
bram-odoo, line 1043: I'm curious about this one. Do you have any concrete example?

content/developer/howtos/website_themes/shapes.rst
bram-odoo: Thanks for the explanations, it's 100% clear :-)

content/developer/howtos/website_themes/shapes.rst
```diff
@@ -270,26 +270,44 @@ Define the styles of your shape.
...
+       static id = "airproofBackgroundShapes";
+       resources = {
+           background_shape_groups_providers: () => ({
+               airproof: {
+                   label: _t("Airproof"),
+                   subgroups: {
+                       airproof: {
+                           label: _t("Airproof"),
+                           shapes: {
+                               "illustration/airproof/01": {
```
bram-odoo, line 293: Is it mandatory to use `illustration/` ? 
On my side I tried my chance with `website_airproof/waves/01` to use a folder based on the project/module then a sub category that is more descriptive.

content/developer/howtos/website_themes/theming.rst
```diff
@@ -219,11 +219,21 @@ By reading the source code, variables related to options are easily noticeable.
...
-      data-name="..."
-      data-customize-website-views="..."
-      data-customize-website-variable="'Sidebar'"
-      data-img="..."/>
+   <BuilderSelectItem
+      id="'...'"
+      title.translate="..."
+      actionParam="[
+         {
+            action: 'websiteConfig',
```
bram-odoo, line 227: Where could we find all the parameters available in this `websiteConfig`? ATM I think this is the point that I understand the less.

content/developer/howtos/website_themes/gradients.rst
bram-odoo: Just to be sure: is the result the same? My point here was to keep the standard ones and add our custom gradient to the existing list, not replace them completely by our custom gradients.

content/developer/howtos/website_themes/gradients.rst
```diff
@@ -39,17 +39,36 @@ Custom
...
+            id: "gradient",
+            name: _t("Gradient"),
+            component: AirproofColorPickerGradientTab,
+        },
+        { sequence: 60 }
+    );
+
+
+.. note::
+   Add the JavaScript file to the `website.website_builder_assets` bundle so the editor loads it.
```
bram-odoo, line 74: We initially loaded it to the `html_editor.assets_editor` as we did an override related to the `html_editor` module

```yaml
        "html_editor.assets_editor": [
            "website_airproof/static/src/js/html_editor/main/font/color_picker_gradient_tab.js",
        ],
```

content/developer/howtos/website_themes/shapes.rst
```diff
@@ -270,26 +270,44 @@ Define the styles of your shape.
...
-         </div>
-      </xpath>
-   </template>
+   registry.category("website-plugins").add(
+       AirproofBackgroundShapesPlugin.id,
+       AirproofBackgroundShapesPlugin
+   );
+
+.. note::
+   Add the JavaScript file to the `website.website_builder_assets` bundle so the editor loads it.
```
bram-odoo, line 310: We loaded it through the `html_builder.assets` as we override a file from this module: 
https://github.com/odoo/odoo/blob/d0f9862a1f378919880c7aab84544a46612f18fe/addons/html_builder/__manifest__.py#L37

I guess it's preferable to use the `website_builder_assets` to bundle the files only for the Website context (we don't need this shape in the Mailing Editor).

content/developer/howtos/website_themes/gradients.rst
duau-odoo: https://github.com/odoo/documentation/blob/970b5f8f556d8be4e3dbcacfc453e3b93f91edf0/content/developer/howtos/website_themes/gradients.rst?plain=1#L45-L70

Yes it's the same as your solution, we extend the current component `ColorPickerGradientTab`, so we still have the original ones, only difference is how we change it in the registry

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: I thought you were using this file to register the snippets in `snippet_groups`, or are you doing this directly in the snippet files themselves?

I just based myself on how we do in website, and bridge module:
https://github.com/odoo/odoo/blob/19.0/addons/website/views/snippets/snippets.xml
https://github.com/odoo/odoo/blob/19.0/addons/website_event/views/snippets/snippets.xml

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: Versioning system is still the same, only difference is that 000.js files became Interactions, and are not public widget anymore.
00X.xml and 00X.scss are still the same

The first PR refactoring publicwidget if you want to have a look: https://github.com/odoo/odoo/pull/185998

robinlej: By the way, and just FYI, we removed the JS versioning system because we never had to use it in over 10 years, so we thought that it wasn't necessary and used the excuse of a refactoring to remove it.
(Which is obviously why a proper case turned out 2 months ago, and we now have a `GallerySlider001.js`.)

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: Fully agreed that it's confusing, and tbh we don't really have a convention between files in `plugins/` and other in `plugins/options/`.
Original goal I think was to put in `plugins/` files who are not core plugins and not options, and obviously options plugins in the `options/` folder, but everything has been a bit mixed up, so now there's no clear delimitation between the two.
We need to put some order in it but obviously it can't be done in stable.

To get back to the point, I think it would be easier for you to take inspiration on website bridge modules, rather than website directly, as the new file structure is shallower in those modules.
And for example in those modules the folder is called `website_builder` and not `builder` like in website,but I don't really have a preference so it's really as you want, as long as it's clear for people reading the doc !

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: After the refactoring, both were available, `BaseOptionComponent` were for more complex options and then we [refactored ](https://github.com/odoo/odoo/pull/230283) them all into `BaseOptionComponent` in stable... 🫣
Obviously it broke some clients customisation and we brought back the object way of creating options for compatibility.

That said, even though it works without the class, the "proper" way of doing it is with `BaseOptionComponent` because otherwise you can't extend it, patch it
There's also a context we are building around the options, you can access plugins shared methods through dependencies

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: Yes I think, it should work, we have an example in website here:
https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/layout_option/grid_column_option.js

Technically, we have only two occurences of using `:not` in a selector in html_builder and website.
So if you need it for a complex selector, it will work, but otherwise I think it's better to follow the `selector` and `exclude` logic

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: There are only two use cases in the codebase:
[DonationOption](https://github.com/odoo/odoo/blob/0d7e3d4c0ea976e37871ca44a10a442cce7caa85/addons/website_payment/static/src/website_builder/donation_option.xml#L17-L24)

image

So this ones are just values you're defining, you can order them as you want etc.

and 
[FormOption](https://github.com/odoo/odoo/blob/0d7e3d4c0ea976e37871ca44a10a442cce7caa85/addons/website/static/src/builder/plugins/form/form_option.xml#L222-L230)

image


There it will depends of the type of field, it can be some custom values as well, or depends of records like shown on the screenshot. And users are able to select which values are available for visitors to pick from

content/developer/howtos/website_themes/shapes.rst
duau-odoo: It's not mandatory, it depends of your record like shown there :
https://github.com/odoo/documentation/blob/970b5f8f556d8be4e3dbcacfc453e3b93f91edf0/content/developer/howtos/website_themes/shapes.rst?plain=1#L200-L205


So if I'm not mistaken, as long as you respect the url, it should work fine

content/developer/howtos/website_themes/gradients.rst
duau-odoo: It depends mostly of your goal, we try to keep everything scoped, if you load it into `html_editor.assets_editor` it means that in every module you have the colorpicker, you will have your custom gradient available.
Otherwise if it's in `website.website_builder_assets`, you will only have it into the website builder colorpicker

content/developer/howtos/website_themes/shapes.rst
duau-odoo: Indeed, same as my gradients comment, if you load it in `html_builder.assets` it means that your customization will also end up in `mass_mailing`

content/developer/howtos/website_themes/theming.rst
duau-odoo: That's a very good question, technically [`websiteConfig`](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/customize_website_plugin.js#L555-L735) is an action like any other.
We use it to toggle views, assets and set CSS vars
Also, those changes are made in DB, so it is saved right away


We don't really have a list of parameters, but I tried to list them:
- `views` [HeaderTemplateOption](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/options/header/header_template_option.xml#L12-L16), used to enable/disable a view
- `assets` [ThemeButtonOption](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/theme/theme_button_option.xml#L79), it's used to enable/disable assets
- `vars` [HeaderTemplateOption](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/options/header/header_template_option.xml#L12-L16), used to set a CSS variable 
- `varsOnClean` [FooterOption](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/options/footer_option.xml#L85) (it's the only usecase in the codebase), it's used to define the value to use when you clear the option instead of `null`
- `checkVars` [HeaderTemplateOption](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/options/header/header_template_option.xml#L12-L16)
- `resetViewArch` [HeaderElementsOption](https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/options/header/header_elements_option.xml#L21), it will reset the view arch to its original template when the view is being disabled

(Not sure if the examples are really helping understand the logic :grimacing:)

duau-odoo:
@bram-odoo thanks for the in-depth review ! Don't hesitate if you have additional questions, feedback, or changes you would like !

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: Yep, we register the snippets in their groups with this file and additionnaly we used it also to declare the custom templates used in the dynamic snippets (blogs, products, etc), custom global options for the Website Builder, etc : 
https://github.com/odoo/tutorials/blob/18.0/website_airproof/views/snippets/options.xml

But in 19, this way has less sense regarding the new files structure that comes with Owl :-)

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: I think both could make sense. In a RD/Product perspective, I 100% agree :-) From a Services perspective, I guess it could be relevant to keep this versioning system to avoid breaking contents of the client when we have to upgrade the module. 

> (Which is obviously why a proper case turned out 2 months ago, and we now have a `GallerySlider001.js`.)

So you rolled back finally? (with the old `data-js` attribute etc) or did you create "an exception" for this case in particular? I'm just wondering if this is something that we still can use or not at all 😄

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: Ok so whenever it's possible always go for the BaseOptionComponent :-) It's funny to discover why some code snippets are built the way they are 😄

content/developer/howtos/website_themes/theming.rst
bram-odoo: It actually does (less sur for the checkVars but meeeh it's ok. As FP said, Designers will figure it out 🫣)

content/developer/howtos/website_themes/gradients.rst
bram-odoo: And the `website.website_builder_assets` is definitely a good choice to keep things secured.

content/developer/howtos/website_themes/theming.rst
duau-odoo: My bad, I forgot the explanation for checkVars ! It's only used in `HeaderTemplateOption`, and basically it's used for the `isApplied` method:
https://github.com/odoo/odoo/blob/19.0/addons/website/static/src/builder/plugins/customize_website_plugin.js#L574-L583

Basically `checkVars` just define if we will compare the `vars` or not to know the "status" of the option

So in the example I gave you, for the header templates
In `actionParam`, there are `views` and `vars`, `checkVars` is set to `false`:
We will only check if the view is active to know if the option is active or not, we won't check the `vars`

content/developer/howtos/website_themes/building_blocks.rst
robinlej: We didn't roll back, in the sense that we didn't change the file names once again (we just have one "exception" and a `data-vjs="001"` on 1 snippet XML). But nothing prevents anyone to use the names they want and patch as they like (even the file structure that we give here is just "the way we do it", but if a user wants a fancy tree that has nothing to do with it, they can :shrug: ).

(The idea behind removing the JS versions (and the reason why we never used it until recently) was that, in JS, we can just keep writing in a single file and add conditions to keep the whole logic without breaking older nor newer snippets. The only reason we ended up making `GallerySlider001` is that literally all the logic was either useless, different or had to be adapted, and nothing was left in common, so it was better for future readability.)

content/developer/howtos/website_themes/theming.rst
bram-odoo: Ok I got it! :-) Thanks, @duau-odoo 🙏🏻

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: Yep, obviously nothing is impossible and a different structure won't change/break anything. We just try to stick as closely as possible to the "standard" (and so the way the R&D built things) to have a clear and consistent way to do things in the doc. There is more than one way to achieve the same result but our goal here is especially to give the most consistent and easy-to-use one as a starting point.

Thank you for these explanation, @robinlej 🙂

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: Indeed! I see the same pattern in the bridge modules. Even the "global" options are way easier to locate in these modules. That's a good point!

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: Last question (I hope!) about this one: I see that Option and Plugin are splitted into 2 separate files: 

- `snippet_name_option.js`: Extends `BaseOptionComponent`
- `snippet_name_option_plugin.js`: Extends `BuilderAction`, `Plugin`, etc 

But I see both `BaseOptionComponent` and `Plugin` can override methods like `setup()`, `getValue()` (and/or declare custom methods). Basically, what's the difference/added value of doing it into one or the other? 🫣

duau-odoo: I will probably forget a lot of details but I'll try to explain as best as I can.

Plugin purpose:
1. Organize the code with a group of determined set of technical and/or functional features
2. Be the entry point to interact with the rest of the system (other plugins)
3. Explicitly specify and require dependencies with other plugins

They interact with other plugins through `shared` methods and `resources`
Plugins live during the whole `Editor` session, from the moment you enter edit mode to the moment you quit it.
Which is a big difference with `BaseOptionComponent`:

`BaseOptionComponent` are set and destroyed based on your selection, they live with the render of the sidebar, the components are registered through `builder_options` resource of a `Plugin`.


Example with a very basic option [`AlertOption`](https://github.com/odoo/odoo/blob/19.0/addons/html_builder/static/src/plugins/alert_option_plugin.js):
When you enter edit mode, `AlertOptionPlugin` will be loaded and setup, this plugin doesn't have a `setup()` and doesn't run any specific code.
It has a resource `builder_actions` with `AlertIconAction` and a `builder_options` `AlertOption`, those two resources act like a registry, so your specific `BuilderAction` and `BaseOptionComponent` are now part of their respective registry.

Now, when you click on `s_alert` snippet, `AlertOption` will be set up, and options related to it will appear on the sidebar. But at the moment where you click on another snippet, obviously the options related to `s_alert` will disappear from the sidebar, and `AlertOption` will be destroyed.
And each time you will select `s_alert` again, the component will be created again.
So deciding where the code runs, in which `setup()` etc, depends of the close link with the option or not, and the lifecycle you need for your code

Also, you're not obliged to split the component in a file and the plugin in another, we only do it 50% of the time :grimacing: again not very consistent.

bram-odoo: Ok got it! Basically, the `BaseOptionComponent` is like the current (electricity) that you can switch on/off and the Plugin is like the wall socket. One can be shut down like the other remains there no matter if the light is on or not.

duau-odoo: I added a sentence about their lifecycle, because it's a key information

https://github.com/odoo/documentation/blob/65b0389b6da99def926f8e4b00f77f110d7728cb/content/developer/howtos/website_themes/building_blocks.rst?plain=1#L686-L687

https://github.com/odoo/documentation/blob/65b0389b6da99def926f8e4b00f77f110d7728cb/content/developer/howtos/website_themes/building_blocks.rst?plain=1#L745-L746

content/developer/howtos/website_themes/layout.rst
```diff
@@ -658,16 +670,21 @@ active footer template first.
...
+
+   <t t-inherit="website.FooterTemplateOption" t-inherit-mode="extension">
+      <xpath expr="//BuilderSelect" position="inside">
+         <BuilderSelectItem
+            title.translate="airproof"
+            actionParam="{
+                  view: 'website_airproof.footer',
+                  vars: { 'footer-template': 'airproof' }
+            }">
+            <Img attrs="{ style: 'width: 100%;' }"
```
bram-odoo:
```suggestion
            <Img class="'w-100'"
```

duau-odoo: Good catch, it's how we did it in the codebase, it's a mistake ! But I think I also overcomplexified this, I think we can add a new footer without doing an xpath on the template, but rather using the resource `footer_templates_providers`, I'm changing that !

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo, line 1: Regarding what we discussed about the `data-vjs`, do you think we should keep this section as it is or remove the mention of the JS from it? 
https://www.odoo.com/documentation/19.0/developer/howtos/website_themes/building_blocks.html#compatibility-system

content/developer/howtos/website_themes/layout.rst
```diff
@@ -657,17 +669,34 @@ active footer template first.
 
 **Option**
 
-.. code-block:: xml
-   :caption: ``/website_airproof/views/website_templates.xml``
+.. code-block:: js
+   :caption: ``/website_airproof/static/src/website_builder/airproof_footer_plugin.js``
+
+   import { Plugin } from "@html_editor/plugin";
```
bram-odoo, line 675: @duau-odoo Fun fact! `website_sale` does not take advantage of the `footer_templates_providers` :
https://github.com/odoo/odoo/blob/11e46faae3350a66a342d2f1f0dd12fbe45198e2/addons/website_sale/static/src/website_builder/footer_option.xml

😅

content/developer/howtos/website_themes/building_blocks.rst
duau-odoo: As we have one example in the codebase, I think we should keep it, it may be useful someday to someone :man_shrugging:

content/developer/howtos/website_themes/layout.rst
duau-odoo: Thanks for reporting it ! I noticed this as well last week randomly, I will open a PR for it

content/developer/howtos/website_themes/building_blocks.rst
```diff
@@ -20,31 +20,38 @@ At the end of this chapter, you will be able to :ref:`create custom snippets
...
 
 .. seealso::
-   `XML templates of the different snippets
-   <https://github.com/odoo/odoo/blob/ccb78f7af2a4413836a969ff8009dc3df6c2df33/addons/website/views/snippets/snippets.xml>`_
+   `Snippets and groups registry in XML
+   <https://github.com/odoo/odoo/blob/8fe9109b8915c940998108ded22960eb699c1f4b/addons/website/views/snippets/snippets.xml>`_
+
+.. tip::
+   JS Files under `/snippets` are interactions:
+   `example <https://github.com/odoo/odoo/blob/dd148d22a03e27c0687f106aeb5179741e8f1e5e/addons/website/static/src/snippets/s_floating_blocks/floating_blocks.js>`_ with the `s_floating_blocks` snippet and the base `Interaction <https://github.com/odoo/odoo/blob/78fdef11cb79a40222e26c3a1f9971c0b45cb000/addons/web/static/src/public/interaction.js>`_ class.
```
AntoineVDV: Since you reference entire files (not just a specific line), you might want to use
```rst
<{GITHUB_PATH}/path/to/file.js>`_
```

AntoineVDV: And if you do, take care to apply it to other occurences too.

robinlej (approved):
Good for me!

duau-odoo:
@robodoo r+

PR #15886, opened 2026-01-06
[REM] pos: http to https connexion

LoredanaLrpz:
task-5417353


davidmonnom:
See https://github.com/odoo/documentation/pull/16041

@LoredanaLrpz  I haven't seen your changes, but I also removed that in another PR that adds information for LNA.

PR #15899, opened 2026-01-06
[FIX] Inventory: Putaway rules for untracked products

erjer-odoo:
An external user noticed a discrepancy in our documentation that stated that untracked products could not use putaway rules. I've fixed the table where that information appears.

This 18.0 PR can be FWP up to master.


Felicious (approved):
@robodoo r+

PR #15908, opened 2026-01-07
[IMP] developer/http: adapt references

Julien00859:
The odoo.http module is a huge file with thousands of lines. In this work we split it in many smaller chunks. The desire to split odoo.http has been in the air for quite some time, notably with the introduction of the many facade objects.

This work only moves the constants, functions and classes in new dedicated odoo.http-submodules. For the documentation it means that a few reference had to be updated.

### DIFF
--- a/content/developer/reference/backend/http.rst
+++ b/content/developer/reference/backend/http.rst
@@ -66,14 +66,14 @@ Request
 The request object is automatically set on :data:`odoo.http.request` at
 the start of the request.
 
-.. autoclass:: odoo.http.Request
+.. autoclass:: odoo.http.requestlib.Request
     :members:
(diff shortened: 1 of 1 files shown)


kmagusiak (approved):

Julien00859:
Hi @odoo/doc-review it only is a minor change, and the community PR is hard to keep afloat, can we get a delegate?

Feyensv (approved):
@robodoo delegate+

LGTM 👌 
Kind remember that a commit/PR should explain its changes regarding the current repository ideally.

AntoineVDV:
@robodoo squash

Julien00859:
@robodoo r+

PR #15910, opened 2026-01-07
[FIX] iot: no bluetooth printers via IoT

Ysoroko:
This PR removes the mention of support of bluetooth printers via iot. Bluetooth printers aren't currently supported at all on raspberry pi iot boxes and can be supported on virtual/windows iot boxes if the windows printing server allows it but this isn't tested / recommended so we remove it altogether from the documentation.

### DIFF
--- a/content/applications/general/iot/devices/printer.rst
+++ b/content/applications/general/iot/devices/printer.rst
@@ -15,7 +15,7 @@ control point or a quality check.
 Connection
 ==========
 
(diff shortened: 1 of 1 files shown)


auva-odoo:
Hello @Ysoroko and thank you for the PR! I was just wondering why you chose to target master specifically since you mentionned that Bluetooth printers aren't currently supported at all?

Ysoroko:
replaced by https://github.com/odoo/documentation/pull/15934

PR #15917, opened 2026-01-07
[IMP] dfn tooltip css improvements

nikibrown:
Improve the styling of the :dfn: role in Odoo Documentation. 

Formats as question mark in superscript, in hyperlink link state what's in backticks after the :dfn: --> goes in a tooltip aligned to the ?

Task https://www.odoo.com/odoo/project/3835/tasks/3997362

### DIFF
--- a/extensions/odoo_theme/static/js/layout.js
+++ b/extensions/odoo_theme/static/js/layout.js
@@ -53,6 +53,43 @@
             // Use the style of L2 toctrees on L1 toctrees.
             toctreeWrapper?.classList.add('o_toc_l1_to_l2');
         }
-    });
 
+        // Position tooltips to prevent overflow at screen edges
+        content.querySelectorAll('.dfn').forEach(dfn => {
+            const tooltip = dfn.querySelector('span');
+            if (!tooltip) return;
+
+            // Calculate and position tooltip when hovering over dfn
+            dfn.addEventListener('mouseenter', function() {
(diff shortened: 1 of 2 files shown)


nikibrown:
@StraubCreative First pass for styling .dfn existing HTML structure into a css tooltip. 

Note that the current formatting in RST files includes parenthesis. We could do a find / replace to remove in the codebase. 

Example of what the tooltip looks like:

image

And tooltip with parenthesis manually removed from RST file

image

StraubCreative (commented):
Hi @nikibrown 

I think the PoC here is good. I think there's definitely improvement for this visual aid feature and the formatting appears to be both clear and consistent.

Before we hand to dev team, can you take a look at the clipping that seems to be happening? Two examples:
1. tooltip expands beyond original x-axis and forces the width of the whole page wider instead of line-breaking against the edge of the page/container and becoming multi-line. Image

2. similar instance, the tooltip's content is clipped off the side edge of the page. Image

Thank you for your work on this, can tag again when ready 👍 

_edit: formatting_

StraubCreative:
Separate: don't forget to also link project tasks in the PR description so reviewers (especially outside our team) can quickly see what the purpose/intent is of the work along with it's progress. Documenting what we do in clear detail saves a lot of "who's on first?" type of questioning 😉

nikibrown:
@StraubCreative Updated with some JS to calculate overflow and position accordingly. 
tooltip-1:26:2026

PR #15943, opened 2026-01-09
[IMP] CRM: Capturing undocumented error message for leads

theRealThagomizer:
Hiya, @jero-odoo! Got a real small PR for you. All I did was add a line, remove some align: center tags, and add a screenshot that I took for another PR. This behavior hasn't been captured, so I just wanted to put it in here, too, since it was relevant. This particular page as a whole could use some attention, but I'm planning to circle back around after doing some CRM tutorials and becoming more familiar with the material. Thanks!


content/applications/sales/crm/acquire_leads/convert.rst
```diff
@@ -39,18 +36,23 @@ To convert a lead into an *opportunity*, navigate to :menuselection:`CRM app -->
 on a lead from the list to open it.
 
 .. warning::
+   Attemping to convert a lead with a 100% probability into an opportunity will result in an error
```
jero-odoo:
```suggestion
   Attempting to convert a lead with a 100% probability into an opportunity will result in an error
```

theRealThagomizer:
Thanks for the quick turnaround, @jero-odoo!

@Felicious, this is ready for final review! This is another page where all I did was update a screenshot and add the line about 100% probability lead/opportunity conversions returning an error. I made this update since I saw the page was focused specifically on the topic of converting leads and made no mention of the message.

theRealThagomizer:
Thanks, @Felicious!

@robodoo r+

PR #15982, opened 2026-01-13
[IMP] Inventory: Update count products

erjer-odoo:
While reviewing our documentation for inconsistencies related to certification questions, I noticed that the information in the "Inventory adjustments" documentation was out of date.

Task in Odoo: https://www.odoo.com/odoo/action-4043/5215187

This PR updates the content to reflect the behavior in 19.0. UI language was updated throughout, and screenshots were updated to match.


content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -11,28 +11,28 @@ damage, human error, theft, or other factors. As such, inventory adjustments mus
...
+To view the :guilabel:`Physical Inventory` page, navigate to :menuselection:`Inventory app -->
 Operations --> Physical Inventory`.
 
-.. image:: count_products/inventory-adjustments-page.png
-   :alt: In stock products listed on the Inventory adjustments page.
+.. image:: count_products/physical-inventory-page.png
+   :alt: In stock products listed on the Physical Inventory page.
 
-The :guilabel:`Inventory Adjustments` page lists all products that are currently in stock.
+The :guilabel:`Physical Inventory` page lists all products that are currently in stock.
```
larm-odoo: I suggest moving this above the image, so it's text, image, admonition block (just a nice flow, totally optional).

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -11,28 +11,28 @@ damage, human error, theft, or other factors. As such, inventory adjustments mus
...
+   page. To view product lines with zero current quantity, go to :menuselection:`Inventory app -->
+   Reporting --> Stock`.
 
 For each product line, the following information is listed:
 
 - :guilabel:`Location`: the specific location in the warehouse where a product is stored. This
   column is **only** visible if :doc:`Storage Locations <use_locations>` are enabled.
-- :guilabel:`Favorite`: identifies products that have been favorited.
 - :guilabel:`Product`: the product whose quantity is listed on the inventory adjustment line.
+- :guilabel:`Product Category`: the category of the product list on the inventory adjustment line.
```
larm-odoo: Is this what you see in the brackets next to the product name on the 'Product' column? If so, I would specify where you see this, since I would assume it is a column, but I do not see one.

Also, these all seem like full sentences, so I would capitalize the first word for each of the entries.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -11,28 +11,28 @@ damage, human error, theft, or other factors. As such, inventory adjustments mus
...
 - :guilabel:`Lot/Serial Number`: the tracking identifier assigned to the specific product listed. It
   can contain letters, numbers, or a combination of both.
```
larm-odoo, line 34: I can't comment on the line, but since the note about Lot/SN's is related to the specific bullet point, I would have the admonition indented to match (so just add 2 spaces to the left of each line of the block).

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -42,36 +42,38 @@ For each product line, the following information is listed:
    product line with its own lot/serial number, displayed under the :guilabel:`Lot/Serial Number`
    column.
 
+- :guilabel:`Last Count Date`: the last time the quantity was updated.
```
larm-odoo: Like with the category, I don't see this column called out so I woudl specify where this information is, since I can't seem to find it

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -42,36 +42,38 @@ For each product line, the following information is listed:
...
+- :guilabel:`Counted`: the real quantity counted during an inventory count. This field is left blank
+  by default but can be changed, depending on if it matches the :guilabel:`On Hand` quantity or not.
 - :guilabel:`Difference`: the difference between the :guilabel:`On Hand Quantity` and
   :guilabel:`Counted Quantity`, once an inventory adjustment is made. The difference is
   automatically calculated after every inventory adjustment.
-- :guilabel:`Scheduled Date`: the date at which a count should be made. If not otherwise specified,
-  this date will default to the 31st of December of the current year.
-- :guilabel:`User`: the person assigned to the count in the database. This can either be the person
-  physically counting the inventory, or applying the count in the database.
+- :guilabel:`Unit`: the *unit of measure* in which the product is measured. Unless otherwise
```
larm-odoo: You could have the italicized 'unit of measure' link to where it is set/explained on a product form if you want -totally optional.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -42,36 +42,38 @@ For each product line, the following information is listed:
...
+  specified (e.g., in :guilabel:`Pounds` or :guilabel:`Ounces`), the default :abbr:`UoM (Unit of
+  Measure)` is :guilabel:`Units`.
```
larm-odoo: General comment for this whole bullet list. I read this, and I assume each bullet point is a labeled column, which does not seem to be the case. I would ensure that everything in the list is either physically shown on the report, and if it's not (such as Category, Expiration Date, etc.), that it is clearly explained where the information can be found, as I was unable to locate several of them. I also suggest creating a blank test database for yourself, so you can go to this rpeort and view what the default items are. I know some runbots will show/hide columns that are not the 'out of the box' way that Odoo may show columns and information. This way, you can say 'these are hidden by default but can be made visible with the (whatever we call it) button...' or something like that. Hopefully that makes sense!

larm-odoo: I see you have the admonition below =)

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -42,36 +42,38 @@ For each product line, the following information is listed:
...
 .. _inventory/create-adjustment:
 
 Create an inventory adjustment
 ------------------------------
 
-To create a new inventory adjustment from the :menuselection:`Inventory Adjustments` page, click
+To update an existing physical inventory entry, click the :guilabel:`Counted` column for the
+product that must be adjusted. Adjust the count.
+
+To create a new inventory adjustment from the :menuselection:`Physical Inventory` page, click
```
larm-odoo:
```suggestion
To update an existing physical inventory entry, click into the corresponding field in the :guilabel:`Counted` column for the
product that must be adjusted. Adjust the count, then click :guilabel:`Apply All`.

To create a new inventory adjustment from the :menuselection:`Physical Inventory` page, click
```
Just because you click in the field, not 'the column (I doubt anyone would click on the word column, but you never know!) and then you need to save it in some way or it won't take.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -81,36 +83,35 @@ To create a new inventory adjustment from the :menuselection:`Inventory Adjustme
...
 
-To the right of the :guilabel:`Counted Quantity` column, the :guilabel:`Scheduled Date` and
-:guilabel:`User` can also be changed via their respective drop-down menus. Changing the
-:guilabel:`Scheduled Date` changes the date that the inventory adjustment should be processed on,
-and selecting a responsible :guilabel:`User` assigns a user to the specific inventory adjustment
-for traceability purposes.
+The :guilabel:`Scheduled` date and :guilabel:`User` can also be changed via their respective
+drop-down menus. Changing the :guilabel:`Scheduled` date changes the date that the inventory
+adjustment should be processed on, and selecting a responsible :guilabel:`User` assigns a user to
+the specific inventory adjustment for traceability purposes.
```
larm-odoo, line 107: Question - when making an adjustment, wouldn't you want to always change the date ot the current date, and the user who did the ocunt, in the user column? So this way, you knwo when an dhwo made the adjustment? I might be completely wrong, so ignore this if it's not at all correct.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -81,36 +83,35 @@ To create a new inventory adjustment from the :menuselection:`Inventory Adjustme
...
 
 Once all changes have been made to the new inventory adjustment line, click away from the line.
 Doing so saves the adjustment, and moves the line to the top of the page.
```
larm-odoo: I did not have this happen - it stayed on the bottom. I also had to click 'Apply' to make it take the new item, and then it moved in alphabetical order, it did not go to the top of the list. I woudl double-check this since I experienced it differently COuld just be my runbot

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -81,36 +83,35 @@ To create a new inventory adjustment from the :menuselection:`Inventory Adjustme
...
-If the :guilabel:`Counted Quantity` is greater than the :guilabel:`On Hand Quantity`, the value in
-the :guilabel:`Difference` column is **green**. If the :guilabel:`Counted Quantity` is less than the
-:guilabel:`On Hand Quantity`, the value in the :guilabel:`Difference` column is **red**. If the
-quantities match, and have not been changed at all, no value appears in the :guilabel:`Difference`
-column.
+If the :guilabel:`Counted` quantity is greater than the :guilabel:`On Hand` quantity, the value in
+the :guilabel:`Difference` column is **green**. If the :guilabel:`Counted` quantity is less than the
+:guilabel:`On Hand` quantity, the value in the :guilabel:`Difference` column is **red** with a
+negative value. If the quantities match, and have not been changed at all, `0.00` appears in the
+:guilabel:`Difference` column.
```
larm-odoo:
```suggestion
the :guilabel:`Difference` column is *green*. If the :guilabel:`Counted` quantity is less than the
:guilabel:`On Hand` quantity, the value in the :guilabel:`Difference` column is *red* with a
negative value. If the quantities match, and have not been changed at all, `0.00` appears in the
:guilabel:`Difference` column.
```

Since it's a descriotion of what you see and not the actual word you see, I recommend either using italics, or just having the colors be plain text.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -81,36 +83,35 @@ To create a new inventory adjustment from the :menuselection:`Inventory Adjustme
...
 
 .. image:: count_products/difference-column.png
-   :alt: Difference column on inventory adjustments page.
+   :alt: Difference column on Physical Inventory page.
 
 At this stage, the count (:dfn:`inventory adjustment`) is recorded, but not yet applied. This means
```
larm-odoo: I've never seen this formatting before- it renders like this, is this what you intended? What does it mean?

image

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -121,35 +122,34 @@ counted quantity.
...
 --------------------
 
-|Ia| can be completed in several ways. The first way is to click the
-:guilabel:`Apply` button on the line at the far right of the page. The second way is to tick the
-checkbox on the far left of the line. Doing so reveals new button options at the top of the page,
-one of which is an :guilabel:`Apply` button. Clicking this button instead causes an
-:guilabel:`Inventory Adjustment` pop-up window to appear.
+|Ia| can be completed in two ways. The first way is to click the :guilabel:`Apply` button on the
+line at the far right of the page. The second way is to select the checkbox on the far left of the
+line. Doing so reveals new button options at the top of the page, one of which is an
```
larm-odoo: You don't have to select the boxes, you can just click 'Apply All' at the top and it will select all the changes. I woudl mention that as well.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -121,35 +122,34 @@ counted quantity.
 Apply adjusted count
```
larm-odoo, line 128: OK, now I understand hwo this is formatted - there is an 'ajust the numbers' then 'apply the numbers'.

Only because I think this migh tbe helpful to uswrs, you may want to ocnsider changing the organization to 'Make a single adjustmet' and 'adjust multiple products at once' - since there are different ways to do this.

It seems like a single item, you just enter the new number, click 'Apply' a tthe end, and you're odne. No real need to click ;apply all' since its only one item. Or you cna do several at a time, and the you would click 'apply al.

I do see that when you click 'Apply All' you get to explain when/why you adjusted the count, whereas when you clikc 'APplly' you only change the quantityt, with no notes.

Just something ot think about - you cna leave it this way, totally fine, or you can do it by how the users may do adjustments - one at a time, or individually. 

And one last thing, if there is a real industry standard to always write a reason for adjusting the nvenotyr (with the 'apply al' for a single adjustment) then I woudl start the section or have a block explaining that. Something like 'for some companies, all inventory adjustments must be logged with a reason and a date. Fo these situatoins, always click the 'apply all' to ensure all neededd information is logged in the dtabaase. For ocmpaniues who can adjust inventory as needed, without the necessity fo logging reasons, click 'apply;'  -- or osmething to that effect. Because I feel like there may be a need for one vs th eother. Exmaple. I worked at an FDA regulated factory, so EVERYTHING had to be traced, logged, explained, etc -0 in casae of a recall. But if we were manufacturing marbles, it doens't matter/who cares if yu change the inventory from 500 to 505, you know? Just somehting ot think about! Totally optional, but wanted ot throw it out there. Because fo rmy apps, sometimes I get stuck in the 'do everything in order' and then someitmes I think of 'what are th eusers using, why are they reading this, what do they want to know' - which is why I htought of the 'single adjustment' vs 'multiple adjustment'.

You can also do an exmaple block to explain why a commpany would need to do the update one way or anothemr instead of in an opening paragrapg.the main body.

Sorry this commend it a novel! =D

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -121,35 +122,34 @@ counted quantity.
...
    <../reporting/moves_history>` in the *Moves History* report for traceability.
 
-.. image:: count_products/apply-inventory-adjustment.png
-   :alt: Apply all option applies the inventory adjustment once a reason is specified.
-
 Relocate products
 =================
 
 |Ia| can also be used to relocate products to different storage locations, or to
 different packages. To relocate a product, tick the checkbox at the far left of the line for the
```
larm-odoo:
```suggestion
different packages. To relocate a product, click on the checkbox at the far left of the line for the
```

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -121,35 +122,34 @@ counted quantity.
...
-desired product. At the top of the page, click the :guilabel:`Relocate` button. Doing so opens a
-pop-up.
+desired product. At the top of the page, click the :icon:`fa-cog` :guilabel:`Actions` menu, then
+select :guilabel:`Relocate`. Doing so opens a pop-up.
```
larm-odoo:
```suggestion
desired product. At the top of the page, click the :icon:`fa-cog` :guilabel:`Actions` button, then
select :guilabel:`Relocate`. Doing so opens a pop-up.
```
It looks like a button to me?

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -121,35 +122,34 @@ counted quantity.
...
 
 .. image:: count_products/relocate-popup.png
```
larm-odoo: I suggest noving th eimage below the epxlained fields, since you mention what to poulate it with, but it's blank when you first see it.

Also, where does the 'package' appear? Is it only for items ina package? if so, I woudl specify that in the line.

Also, add that they need to click 'Confirm' to make the move.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
larm-odoo: I would also mention you can select multiple items at once.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -171,56 +171,61 @@ the checkbox at the far left of the line for the desired product. At the top of
 :icon:`fa-gear` :guilabel:`Actions` button to open a drop-down menu. Click :guilabel:`Set to 0`.
```
larm-odoo, line 205: Update the word 'tick' thorughout the doc, I see it in a few places.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -171,56 +171,61 @@ the checkbox at the far left of the line for the desired product. At the top of
...
 Once this is complete, :ref:`apply <inventory/apply-adjustment>` the adjusted count.
```
larm-odoo, line 206: I didn't have to click 'Apply' - it didn't even show up.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -171,56 +171,61 @@ the checkbox at the far left of the line for the desired product. At the top of
...
 :menuselection:`Inventory app --> Operations --> Physical Inventory` to update the
-:guilabel:`Counted Quantity` column for each product line.
+:guilabel:`Counted` quantity for each product line.
 
-On each product line, identify whether the value in the :guilabel:`On Hand Quantity` column recorded
-in the database matches the newly-counted value. If the recorded value and the counted value do
-match, click the :icon:`fa-bullseye` :guilabel:`Set` icon at the far right of the product line.
+On each product line, identify whether the value in the :guilabel:`On Hand` column recorded in the
+database matches the newly-counted value. If the recorded value and the counted value do match,
+specify the value in the :guilabel:`Counted` column.
```
larm-odoo:
```suggestion
On each product line, identify whether the value in the :guilabel:`On Hand` column recorded in the
database matches the newly-counted value. If the recorded value and the counted value match,
enter the 'On Hand' quantity in the :guilabel:`Counted` column.
```

Only because when I first read this, I was reading 'do not match' so this is just more clear (for me specifically)

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -171,56 +171,61 @@ the checkbox at the far left of the line for the desired product. At the top of
...
-icon, record the real value in the field in the :guilabel:`Counted Quantity` column.
+Hand` column recorded in the database, record the real value in the field in the :guilabel:`Counted`
+column.
 
-To do so, click the field in the :guilabel:`Counted Quantity` column on the specific inventory
-adjustment line for the product whose count is being changed. This automatically assigns a
-:guilabel:`Counted Quantity` of `0.00`.
+To do so, click the field in the :guilabel:`Counted` column on the specific inventory adjustment
+line for the product whose count is being changed. This automatically assigns a :guilabel:`Counted`
+quantity of `0.00`.
```
larm-odoo: Seeing the '0.00' was confusing at first, so you might want to remove the reference of it, since when you click in, it does show zero, but you cna immediately type in the number. Just might be less confusing if you just say 'click onto the 'cpunted field and enter the updated count.'

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -171,56 +171,61 @@ the checkbox at the far left of the line for the desired product. At the top of
...
-If the :guilabel:`Counted Quantity` is greater than the :guilabel:`On Hand Quantity`, the value in
-the :guilabel:`Difference` column is **green**. If the :guilabel:`Counted Quantity` is less than the
-:guilabel:`On Hand Quantity`, the value in the :guilabel:`Difference` column is **red**. If the
-quantities match, and have not been changed at all, no value appears in the :guilabel:`Difference`
-column.
+If the :guilabel:`Counted` quantity is greater than the :guilabel:`On Hand` quantity, the value in
+the :guilabel:`Difference` column is **green**. If the :guilabel:`Counted` quantity is less than the
+:guilabel:`On Hand` quantity, the value in the :guilabel:`Difference` column is **red** and is a
+negative value. If the quantities match, and have not been changed at all, `0.00` appears in the
+:guilabel:`Difference` column.
```
larm-odoo: Same suggestion as before with the red and green. You can also mention that '0.00' appears in black.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: This column isn't enabled by default. You have to enable it in the column adjust menu.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: I just linked to the UoM doc.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: These fields are intended to specify a responsible user and scheduled date when a count *should* take place. You don't need to record yourself or the current date in there because that data is already captured in the Moves History when you perform the adjustment. Does that make sense?

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: That was there when I edited the doc. It looks like how a definition is formatted - I assume that when set up correctly, it shows a definition when you hover over it. I changed the formatting to just be italics, because that's how it shows up in the moves history.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
larm-odoo: Maybe I am looking at the wrong place, but I still do not see 'Category' anywhere on the list:
image
Is this what you see as well? This is what I'm going off of for 'the following information is listed'.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
larm-odoo: Since it doesn't appear in the default list shown to the user, I recommend only listing what they actually see in the 'this information is shown' list, then you can add additional information in a separate section below, stating that 'the following fields must be activated to view them'. Only because if I were reading this, I would expect to see that on the report, but it's not visible.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
larm-odoo: Ah, totally! I didn't realize. Thanks!

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
larm-odoo: It looks like it is still in the doc - so just be sure to italicize and remove the 'dfn', since the rollover does not work. You could also just say:
At this stage, the count (referred to in Odoo as an *inventory adjustment*) is recorded, but not yet applied. This means

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: Product Category is an additional column -- check the adjust menu to find "Product Category" and enable it.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: In my test database that I have, I have to click Apply. Same thing in runbot.

erjer-odoo:
Hi @larm-odoo -- this should be ready for your review again. Hopefully it's making more sense now with some of the changes you suggested!

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -11,67 +11,73 @@ damage, human error, theft, or other factors. As such, inventory adjustments mus
...
+- :guilabel:`Location`: the specific location in the warehouse where a product is stored. This
+  column is **only** visible if :doc:`Storage Locations <use_locations>` are enabled.
+- :guilabel:`Product Category`: the category of the product list on the inventory adjustment line.
+- :guilabel:`Last Count Date`: the last time the quantity was updated.
+- :guilabel:`Expiration Date`: the date on which the goods with this serial number are due to
+  expire.
+- :guilabel:`Package`: the package containing the quantity listed.
+- :guilabel:`Unit`: the :doc:`unit of measure <../../product_management/configure/uom>` in which the
+  product is measured. Unless otherwise specified (e.g., in :guilabel:`Pounds` or
+  :guilabel:`Ounces`), the default :abbr:`UoM (Unit of Measure)` is :guilabel:`Units`.
```
larm-odoo, line 71: In the runbot I was in, I see these options:
image

I'd just oduble-check on a blank database to make sure these ar the default options. I assume I didn't see an expiiration date option becaus enotihng was going ot expire?

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -81,82 +87,105 @@ To create a new inventory adjustment from the :menuselection:`Inventory Adjustme
...
 
 - :guilabel:`To Location`: the new location for the products.
-- :guilabel:`To Package`: the new package for the products.
+- :guilabel:`To Package`: the new package for the products. This field only appears if you have
+  :guilabel:`Packages` enabled in the *Inventory* settings.
 - :guilabel:`Reason for relocation`: the reason for the move.
 
+Click :guilabel:`Confirm` to confirm the move.
+
+.. image:: count_products/relocate-popup.png
```
larm-odoo, line 191: Just a suggestion - fill in a reason, just to give a nice, complete example. Maybe something like 'Expired Products for Compost' or 'Discontinued - for Donations' and the location can be osmehting rleated ot that, like 'WH/Compost' or whatever makes sense. Just thinking out loud - ignore this if you want!

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -166,61 +195,65 @@ On the resulting pop-up, enter the following information:
...
-column.
+If the :guilabel:`Counted` quantity is greater than the :guilabel:`On Hand` quantity, the value in
+the :guilabel:`Difference` column is *green*. If the :guilabel:`Counted` quantity is less than the
+:guilabel:`On Hand` quantity, the value in the :guilabel:`Difference` column is *red* and is a
+negative value. If the quantities match, and have not been changed at all, `0.00` appears in the
+:guilabel:`Difference` column.
 
-Subsequently, once applied, a move with the difference between the :guilabel:`On Hand Quantity` and
-the :guilabel:`Counted Quantity` is recorded in the product's inventory adjustment history.
+Subsequently, once applied, a move with the difference between the :guilabel:`On Hand` quantity and
```
larm-odoo, line 248: I noticed that after I made an entry in the 'Counted' column, I had to click 'Save' before I was able to select products with the chekcboxes on the left side.  I recommend adding that, before the 'Actions' part is explained below.

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -166,61 +195,65 @@ On the resulting pop-up, enter the following information:
...
+the :guilabel:`Counted` quantity is recorded in the product's inventory adjustment history.
 
 .. image:: count_products/history-inventory-adjustments.png
    :alt: Inventory Adjustments History dashboard detailing a list of prior product moves.
 
-The :guilabel:`Actions` menu appears when one or more products' checkboxes are selected. The
-:guilabel:`Actions` menu includes the option to :guilabel:`Set to quantity on hand`, which sets the
-selected products' :guilabel:`Counted Quantity` to the :guilabel:`On Hand Quantity`, and
-:guilabel:`Set to 0`, which sets the selected products' :guilabel:`Counted Quantity` to zero.
+The :icon:`fa-cog` :guilabel:`Actions` menu appears when one or more products' checkboxes are
```
larm-odoo, line 254: I find this sectoin confusing, because it addresses the 'Set To Zero' - which has it's own section above. I recommend rmeoving the part about set to zero, since it's already in that section. Also, it feels almost like an add-on, below the 'how to adjust multiple items at once' instructions. From what I was able to do, it look slike you can djust a bunch of entries, but you have to click Save to record them. 

I almost feel like this seciton can be just for 'set quantities to on hand' and it can be in its own little section. It could also just be in a tip admonition. I am trying ot think of when you would want to do that- if you have a count, you enter the new count, but then you relize you were wrong, and want to revert to the 'on hand' wuantities? I am thinking if this is a common situation or not. Maybe an admonition, or a separate section, since it throws me off for the section it is in. Just an idea!

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
erjer-odoo: I can add a note about Expiration Dates needing to be enabled in the settings here.

erjer-odoo:
Hi @Felicious -- this is ready for final review when you have a moment. Thank you!

content/applications/inventory_and_mrp/inventory/warehouses_storage/inventory_management/count_products.rst
```diff
@@ -6,157 +6,191 @@ Inventory adjustments
...
+
+.. example::
+   An FDA-regulated factory is required by law to trace, log, and explain all inventory adjustments
+   to ensure that it remains in compliance with federal law. Any product stock moves and associated
+   reasonings would need to be documented thoroughly to ensure that all product is accounted for in
+   case of a recall.
+
+.. note::
+   Applying an inventory adjustment simultaneously creates a :doc:`stock move line (SML)
+   <../reporting/moves_history>` in the *Moves History* report for traceability.
```
Felicious: nitpick: before Odoo 19, we used to call these things stock move lines, and they were crucial in how inventory valuation worked. It was a very Odoo naming convention and functionality. They completely refactored that in 19 and we want to move away from the "stock move line" terminology, even though it's still technically a line and it's still a stock move 🫠 

Ideas on what to rename this? how about "stock move record"? because the moves history report still exists, it's not like we take it out

erjer-odoo: I think we can call it a "stock move record." Thanks for the background -- good to know!

erjer-odoo:
@robodoo r+

PR #15986, opened 2026-01-14
[FIX] Timesheets: Add redirect link

larm-odoo:
Need to add a redirect from the old page title in 17 that changed in 18.

[Task card](https://www.odoo.com/odoo/project/3835/tasks/5497522) for this PR.

### DIFF
--- a/redirects/18.0.txt
+++ b/redirects/18.0.txt
@@ -63,6 +63,10 @@ applications/websites/website/pages/seo.rst applications/websites/website/struct
 applications/websites/ecommerce/products/price_management.rst applications/websites/ecommerce/products/prices.rst
 applications/websites/ecommerce/cart.rst applications/websites/ecommerce/checkout.rst
 
+# applications/services
+
+applications/services/timesheets/overview/time_off.rst applications/services/timesheets/time_off.rst
+
 # applications/general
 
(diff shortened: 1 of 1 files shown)


larm-odoo:
Hi @Felicious - this just needs a final/tech review for the fix. Thanks!

redirects/18.0.txt
```diff
@@ -63,6 +63,10 @@ applications/websites/website/pages/seo.rst applications/websites/website/struct
 applications/websites/ecommerce/products/price_management.rst applications/websites/ecommerce/products/prices.rst
 applications/websites/ecommerce/cart.rst applications/websites/ecommerce/checkout.rst
 
+# applications/services
+
+aapplications/services/timesheets/overview/time_off.rst pplications/services/timesheets/time_off.rst
```
nikibrown: @larm-odoo I think the `a` from the second path made its way to the first path. 

Should probably be `applications/services/timesheets/overview/time_off.rst applications/services/timesheets/time_off.rst`

nikibrown (changes requested):

larm-odoo:
Hi @nikibrown - I can't believe that mistake - it is updated, and ready for another look.

Felicious:
@robodoo r+

PR #15989, opened 2026-01-14
[FIX] Referrals: Incorrect definition of points - ealrier versions

larm-odoo:
The text for what the points mean is incorrect - it is not what they earn when an applicant applies (the original text) but it's the TOTAL points once the applicant is HIRED. 

This PR is for versions 15.0, 16.0, and 17.0 ONLY. 

[Task card](https://www.odoo.com/odoo/project/3835/tasks/5497963) for this PR.

Also, it says staging failed because 15 is no longer supported, but we do have the docs available, so I am not sure how to fix that.


larm-odoo:
Hi @Felicious - this is what I messaged you about. The same fix, but the earlier verisons of the doc.

larm-odoo:
We do not support 15 or 16, remaking a new PR for only 17.

PR #15991, opened 2026-01-15
[FIX] Referrals: Update points definition

larm-odoo:
The text for what the points mean is incorrect - it is not what they earn when an applicant applies (the original text) but it's the TOTAL points once the applicant is HIRED. 

This PR is for version 17.0 ONLY.

[Task card](https://www.odoo.com/odoo/project/3835/tasks/5497963) for this PR.

No points for this fix PR.


larm-odoo:
Hi @Felicious - this is just a no-points fix ready for review.

PR #15993, opened 2026-01-15
[ADD] AI: Server Actions

jero-odoo:
Demonstrates how to update fields, summarize text, or make decisions using AI-driven server actions.
Docs task: https://www.odoo.com/odoo/project/3835/tasks/5440130


gelor-odoo:
I have just a few minor suggestions to prevent the NameError issues we see in Support:

Explicitly name the "AI Schema": In the section "Use case walkthrough", the text mentions: "The AI knows to provide folder_id because it is explicitly declared as an argument in the tools configuration." (Lines 91-94)

`   For the *AI: Move to Folder* tool, the Python code expects a variable named `folder_id`. The |AI|
   knows to provide `folder_id` because it is explicitly declared as an argument in the tools
   configuration. The |AI| uses the argument description to understand what the parameter
   represents, what type of value is expected, and when it is appropriate to supply it.`

Suggestion: Please explicitly reference the "**AI Schema**" list (visible in the screenshot). Users need to know they must click "Add a line" in the AI Schema section to define these variables.

image

Draft Edit: `"...declared as an argument in the AI Schema list within the tool's Usage tab."`

Variable Name Matching: In "Creating custom AI tools", regarding the content variable:

Suggestion: Explicitly state that the Name column in the **AI Schema** must exactly match the variable name used in the Python code. This could be a common cause of failure (e.g., user writes record.write(desc) in Python but defines description in the Schema).

-------------------------------------------------------------

Typos:

Under "How AI server actions work" (line 16): AI-driven wokflows -> AI-driven workflows.

Under "Use case walkthrough" (line 85): This code, executes unconditionally -> This code executes unconditionally (remove comma).

Otherwise, this looks fantastic! 🚀

content/applications/productivity/ai/server-actions.rst
```diff
@@ -0,0 +1,140 @@
...
+   Odoo executes the action.
+
+Use case walkthrough
+--------------------
+
+The *Auto-sort documents in Inbox* example demonstrates the full |AI| Server Action pattern.
+
+.. note::
+   This server action is configured on `Runbot <https://runbot.odoo.com/>`_
+
```
Felicious, line 65: i dont think we should talk about runbot in our official documentation. it's an internal tool that customers shouldn't know about. we should only cover server actions that lives in normal customer databases, so can you 

1.verify whether it exists when u create a new DB
2. remove this note and edit the first sentence to say something like "The auto-sort documents server action is a default server action to [describe what it does]. This server action is explored in this example to demonstrate the AI server action pattern"

content/applications/productivity/ai/server-actions.rst
```diff
@@ -0,0 +1,140 @@
...
+The :guilabel:`Model` for the action is listed as *Document*, and the :guilabel:`Type` is *AI*. The
```
Felicious: Before this line, let's tell them how to navigate there to see the model and type (:

content/applications/productivity/ai/server-actions.rst
```diff
@@ -0,0 +1,140 @@
...
+   If an argument is not defined in the Usage tab, the AI cannot provide it, even if the Python code
+   references it.
+
+   .. image:: server-actions/ai-move-to-folder.png
+      :alt: The usage tab of the AI move to folder server action.
+
+Creating custom AI tools
+------------------------
+
+The same pattern in the example above can be implemented using standard Odoo logic. For example, to
```
Felicious:
```suggestion
The same pattern in the *Auto-sort documents in Inbox* example above can be implemented using standard Odoo logic. For example, to
```

Avoiding directional references because when LLMs read this text, it won't know what we're referring to.

jero-odoo:
@robodoo r+

PR #16042, opened 2026-01-20
docs: add _name requirement warning in Chapter 13 inheritance section

### LINKED ISSUE #15038: Missing important information in Chapter 13
Chapter 13 says:

Image

Unfortunatelly - when `_name` is ommited (implicit)
```
from odoo import models
import logging
_logger = logging.getLogger(__name__)

class EstateModel(models.Model):
    _inherit = ["estate_property"]
    _description = "Enhancement of RealEstate Model"

    def action_sold(self):
        _logger.warning("Child Action sold")
        return super().action_sold()
```
The "Child Action sold" is never invoked (only "Parent Action sold" is displayed).

But when name is explicitly set to the same as "parent model" :
```
from odoo import models
import logging
_logger = logging.getLogger(__name__)

class EstateModel(models.Model):
    _name = "estate_property"
    _inherit = ["estate_property"]
    _description = "Enhancement of RealEstate Model"

    def action_sold(self):
        _logger.warning("Child Action sold")
        return super().action_sold()
```

Both "Child Action sold" and "Parent Action sold" are displayed.

I think this should me a `warning` in the tutorial

PrinceThummar011:
## Description
Adds a warning to Chapter 13 about the requirement to explicitly set `_name` when overriding methods in inherited Odoo models.

## Issue
Without explicit `_name`, method overrides are not invoked, causing silent failures.

## Changes
- Added warning box with explanation
- Included correct/incorrect code examples
- Follows existing documentation format

Fixes #15038

### DIFF
--- a/content/developer/tutorials/server_framework_101/13_other_module.rst
+++ b/content/developer/tutorials/server_framework_101/13_other_module.rst
@@ -72,6 +72,34 @@ module for the ``estate.property`` model. For now, the overridden action will si
         def inherited_action(self):
             return super().inherited_action()
 
+.. warning::
+   When inheriting a model to override methods, you must explicitly set ``_name`` 
+   to the same value as the parent model. If you omit ``_name`` (implicit inheritance), 
+   your method overrides may not be invoked correctly.
+   
+   **Incorrect (method won't be called):**
+   
+   .. code-block:: python
+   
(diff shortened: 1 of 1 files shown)


AntoineVDV:
See https://github.com/odoo/documentation/issues/15038#issuecomment-3839695223

PR #16078, opened 2026-01-21
[IMP] Lunch: Update main doc

larm-odoo:
General updates to doc - updated translation image (missing text) and reformatted Alert section, along with minor style changes.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5866262) for this PR.


larm-odoo:
Hi @jero-odoo - this is ready for a peer review. Thanks!

content/applications/hr/lunch.rst
```diff
@@ -39,7 +39,7 @@ Configure the following:
    example, :guilabel:`EN` for English) and a translation pop-up window appears.
 
    The following is an example for the :guilabel:`Reception notification` field in the settings
```
jero-odoo: Would change to something like "the image below", just because "the following" implies that the example is directly beneath this sentence

larm-odoo:
Hi @Felicious - this is ready for a final/tech review. Thanks!

content/applications/hr/lunch.rst
```diff
@@ -38,8 +38,8 @@ Configure the following:
    field on a form. To add translations for that field, click the two letter language code (for
    example, :guilabel:`EN` for English) and a translation pop-up window appears.
 
-   The following is an example for the :guilabel:`Reception notification` field in the settings
-   menu:
+   The image below is an example for the :guilabel:`Reception notification` field in the settings
+   menu.
```
Felicious: in general, an admonition block shouldn't be too long or contain overly specific instructional information. If it's too long, it might signal that it's worthy of its own section or something. 

How about making a subsection and calling it something like "translation support", and taking all of this text out of the tip admonition ? (: 

Image

larm-odoo:
@robodoo r+

PR #16081, opened 2026-01-22
[IMP] developer/view_architextures: remove dropped map view attributes

jeanschoenlaub:
In 19.0 we refactor the base map ui view, removing
some options from the sidepanel such as the panel title.

In this commit we remove the attributes linked to
the panel title: `hide_title` and `panel_title`

see odoo/enterprise#104618
task#5259181


Feyensv:
As far as I can see, the referenced PR targets master, so why do you remove those attributes in the 19.0 doc? :thinking:

jeanschoenlaub:
You are right I meant to target master :sweat_smile: 

also I kind of fumbled trying to keep the same PR open and retarget master, I opened a new one: 
odoo/documentation#16092

PR #16084, opened 2026-01-22
[ADD] Inventory: Pack in pack

erjer-odoo:
As of Odoo 19.0, functionality was included to support pack-in-pack, or nested packages. This is currently only documented in a talk given at Odoo Experience 2025. This PR documents the feature.

[Task card](https://www.odoo.com/odoo/project/3835/tasks/5869720)


content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,160 @@
...
+        Odoo calculates the package's weight by adding the weight of the empty package plus the
+        weight of the items, which can be found in the :guilabel:`Weight` field, in the
+        :guilabel:`Inventory` tab, of each product form.
+
+- :guilabel:`Max Weight`: maximum shipping weight allowed in the package.
+- :guilabel:`Barcode`: Define a barcode to identify the package type from a scan.
+- :guilabel:`Company`: Specify a company to make the package type available only within that
+  company. Leave the field blank if it is available at all companies.
+- :guilabel:`Carrier`: Specify the intended shipping carrier for this package type.
+- :guilabel:`Carrier Code`: Define a code that is linked to the package type.
```
theRealThagomizer: When I was playing around in runbot, I also saw fields for "sequence prefix" and "routes". I'm guessing that these were enabled by some configuration that runbot has turned on by default and that they aren't germane to the "base" experience, but I just wanted to call them out in case it's relevant.

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,160 @@
...
+
+Open the operation type from the :guilabel:`Operations` menu. The list of operations for that
+operation type displays. Open an operation from that list.
+
+To put the contents of a line into a single package, click the :guilabel:`Details` link for the
+product line. The *Detailed Operations* window opens. In this window, specify a :guilabel:`Store to`
+location, the name of the :guilabel:`Destination Package`, and the :guilabel:`Quantity` of items to
+be put in that package. To break the inventory into multiple packages, click the :guilabel:`Add a
+line` link; be sure that the :guilabel:`Quantity` of each line adds up to the total
+:guilabel:`Demand`. Click :guilabel:`Save` to create the package in Odoo.
```
theRealThagomizer, line 94: Re-reading this after you explained it to me, I think this section is actually fine. You might want to call out that opening an operation will open an invoice, just so the reader knows what to expect. You could also put in a screenshot that shows the Details link, but I'd say neither choice is strictly necessary. Just depends on how thorough you want to be.

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,160 @@
...
+    `Sparkling Water Cans`, and `Soda Cans`. When the order arrives, each product arrives in its own
+    box of 20 cans.
+
+    The warehouse manager specifies a separate package for each product line on the receipt.
+
+    .. image:: pack_in_pack/example-packages.png
+       :alt: Four packages are created on the receipt form.
+
+Then, to package all packages into a single larger package (for example, to package multiple cases
+of soda into a pallet), click the :guilabel:`Put in Pack` button. After the *Put in Pack* window is
```
theRealThagomizer: That's a pile of packages! What about something like, "...to organize all packages into a single larger unit..."?

theRealThagomizer (approved):
Content looks good, @erjer-odoo! I'm glad we talked out my question; re-reading it with the context, I think it's actually fine as-is. I left a few philosophical notes for you, but if you don't want to make any changes at this point, this is ready to move on. Thank you!

erjer-odoo:
Hi @Felicious ! This PR has been reviewed and approved. When you get a moment, could you help me merge? Thank you!

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+--> Settings`. Enable the :guilabel:`Packages` setting. Be sure to click :guilabel:`Save` before
+exiting this page.
+
+.. image:: pack_in_pack/enable-setting.png
+   :alt: Enable Packages in the Inventory configuration settings.
+
+Set package type
+================
+
+The :guilabel:`Set Package Type` setting must be enabled for all operations that require
```
Felicious: minor nitpick. since the feature is not yet on the page, how about italcizing it for now? that's what we do when we're referring to UI elements outside the current page we're having the user currently look at (:


```suggestion
The *Set Package Type* setting must be enabled for all operations that require
```

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+Types` page opens. Open an operation type.
+
+On the operation type form, in the :guilabel:`Packages` section, select :guilabel:`Set Package
+Type`.
+
+.. example::
+    :guilabel:`Set Package Type` is enabled for the `Receipts` operation type.
+
+    .. image:: pack_in_pack/set-package-type.png
+       :alt: Update the Set Package Type setting on the operation type form.
```
Felicious: looks like the alignment is just 1 character too far to the right


```suggestion
   :guilabel:`Set Package Type` is enabled for the `Receipts` operation type.

   .. image:: pack_in_pack/set-package-type.png
      :alt: Update the Set Package Type setting on the operation type form.
```

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+
+- :guilabel:`Package Type` (required): Define the package type’s name.
+- :guilabel:`Size`: Define the package dimensions in millimeters (mm). The fields, from left to
+  right, define the Length, Width, and Height.
+- :guilabel:`Weight`: weight of an empty package (e.g., an empty box, pallet).
+
+    .. note::
+        Odoo calculates the package's weight by adding the weight of the empty package plus the
+        weight of the items, which can be found in the :guilabel:`Weight` field, in the
+        :guilabel:`Inventory` tab, of each product form.
```
Felicious:
```suggestion
       Odoo calculates the package's weight by adding the weight of the empty package plus the
       weight of the items, which can be found in the :guilabel:`Weight` field, in the
       :guilabel:`Inventory` tab, of each product form.
```

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+
+- :guilabel:`Max Weight`: maximum shipping weight allowed in the package.
+- :guilabel:`Barcode`: Define a barcode to identify the package type from a scan.
+- :guilabel:`Sequence Prefix`: Specify a prefix to use when a new package name is created by Odoo.
+
+    .. example::
+        Use `PAL` for the `Pallet` package type. When the first package of the `Pallet` type is
+        created, it is named `PAL0000001`.
```
Felicious:
```suggestion
   .. example::
       Use `PAL` for the `Pallet` package type. When the first package of the `Pallet` type is
       created, it is named `PAL0000001`.
```

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+    box of 20 cans.
+
+    The warehouse manager specifies a separate package for each product line on the receipt.
+
+    .. image:: pack_in_pack/example-packages.png
+       :alt: Four packages are created on the receipt form.
+
+Then, to organize all packages into a single larger unit (for example, to package multiple cases of
+soda into a pallet), click the :guilabel:`Put in Pack` button. After the *Put in Pack* window is
+completed, the :guilabel:`Packages` field for the product lines in the greater package updates with
```
Felicious: > After the Pack in pack window is completed

I'm getting snagged a little on the semantics here. It's not the window that's complete, but that process. How about changing the wording to "after closing the pack in pack window"?

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+the new package name.
+
+.. example::
+    Continuing the example from above, the warehouse manager wants to package all of the products
+    onto a single pallet for storage in the warehouse. They click the :guilabel:`Put in Pack`
+    button, then in the *Put in Pack* window, they specify `Pallet` as the :guilabel:`Package Type`.
+    The :guilabel:`Packages` field for all product lines updates to show the pallet package name,
+    `PAL0000001`.
```
Felicious:
```suggestion
   Continuing the example from above, the warehouse manager wants to package all of the products
   onto a single pallet for storage in the warehouse. They click the :guilabel:`Put in Pack`
   button, then in the *Put in Pack* window, they specify `Pallet` as the :guilabel:`Package Type`.
   The :guilabel:`Packages` field for all product lines updates to show the pallet package name,
   `PAL0000001`.
```

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
+============
+Pack-in-pack
+============
+
+Products are often purchased, sold, or moved around the warehouse in containers rather than as
+individual units. For example, a warehouse might receive a full pallet containing 100 cases of soda.
```
Felicious:
```suggestion
individual units. For example, a warehouse might receive a full pallet containing 100 cases of soda, with each case containing 6 sodas.
```
To really drive home the pack in pack concept, as the first time i read it, i skipped over the "cases"

content/applications/inventory_and_mrp/inventory/product_management/configure/pack_in_pack.rst
```diff
@@ -0,0 +1,170 @@
...
+By using the *Packages* feature, you can track and transfer these containers as distinct units. This
+streamlines operations, as you can move the entire pallet at once rather than scanning every
+individual item.
```
Felicious: The value proposition of pack in pack wasn't immediately clear to me. I think you were really concise in this intro, which I generally really appreciate, but from reading this, I was almost inclined to believe we were talking about normal packages. Since we're introducing a new feature here, I wanted to try to suggest a more detailed explanation of the feature and its value. Let me try to reword (: 

> In more complex warehouse operations, it is sometimes necessary to handle groups of containers (for example, grouping pallets together). These containers may be physically combined or logically grouped at the database level, allowing them to be treated as a single unit for certain stock movements while still retaining visibility into their internal structure. This makes it possible to move the entire group with a single scan, rather than scanning each pallet individually.

My explanation might err on the side of being too technical (and long..), so I'll let you be the final decision maker to decide whether this detail adds/subtracts to the user's understanding of the feature.

erjer-odoo:
@robodoo r+

PR #16092, opened 2026-01-23
[IMP] developer/view_architectures: remove dropped map view attributes

jeanschoenlaub:
In master we refactor the base map ui view, removing
some options from the sidepanel such as the panel title.

In this commit we remove the attributes linked to
the panel title: `hide_title` and `panel_title`

see odoo/enterprise#104618
task#5259181


Feyensv:
robodoo r+

PR #16153, opened 2026-01-27
[FIX] Barcode: Supported barcode types

erjer-odoo:
While evaluating documentation for currentness for 19.0, I noticed incorrect information in the Barcodes parent topic. I've corrected the information at the earliest point that we currently support.

This PR isn't technically for 19.0, but it does address an issue noticed while working on the [Inventory batch updates](https://www.odoo.com/odoo/action-4043/5215187).

### DIFF
--- a/content/applications/inventory_and_mrp/barcode.rst
+++ b/content/applications/inventory_and_mrp/barcode.rst
@@ -7,7 +7,7 @@
 Barcode
 =======
 
-**Odoo Barcode** allows users to assign barcodes to individual products and product categories, and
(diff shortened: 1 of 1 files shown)


Felicious:
Hi @erjer-odoo ! FIX, REM, and MOV PRs go straight to Niki (: But thanks for checking this with me, as your PR actually addresses a content issue, so the PR should be tagged IMP! It's such a minor thing though so im just gonna merge even if the PR isn't technically tagged correctly

@robodoo r+

PR #16196, opened 2026-01-30
[FIX] saudi_arabia: add warning about company name length

viso-odoo:
We have had lots of tickets recently about failed ZATCA onboarding. One of the most common reasons is that the company or branch name is too long compared to the strict limits imposed by zatca.

As for the limits, they are technical, imposed by the QR code generated for every invoice.

The limit itself is 127 bytes, which depending on the exact arabic characters used can range from 42 to 63 arabic characters. 63 is in the best-case scenario where we can encode all characters on 2 bytes.
42 characters is the worst-case scenario where special characters requiring 3 bytes are used.

In an effort to keep our documentation concise and clear, we chose to stick to the best-case scenario of 63 characters, assuming that would be the most common scenario.
Pull requests have been made in the main codebase to make the error messages clearer upon failed onboarding, which should cover the "42 characters edge cases".

### DIFF
--- a/content/applications/finance/fiscal_localizations/saudi_arabia.rst
+++ b/content/applications/finance/fiscal_localizations/saudi_arabia.rst
@@ -40,7 +40,8 @@ Company information
 Go to :menuselection:`Settings --> General Settings --> Companies`, click :guilabel:`Update info`,
 and ensure the following company information is complete and up-to-date.
 
-- The full :guilabel:`Company Name`.
+- The :guilabel:`Company Name`, limited to a maximum of 63 characters to comply with ZATCA
+  requirements.
 - All relevant :guilabel:`Address` fields, including the :guilabel:`Building Number` and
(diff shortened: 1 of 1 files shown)


auva-odoo:
Thank you @viso-odoo for this PR! :pray: I wonder if maybe instead of the important note we should add this information on line 43 directly (where we talk about the company name), especially since we say "the full name" there and, well, I guess that might not be correct if the name's limited to 63 characters :roll_eyes:. So here's my suggestion for line 43:

`The :guilabel:`Company Name`, limited to a maximum of 63 characters to comply with ZATCA requirements.`

(and then I would remove the important note)

What do you think?

viso-odoo:
Hey @auva-odoo thanks for the review :)
I was balancing between both options, so i asked a client how they did - and yes your suggestion actually makes more sense.
I modified the commit accordingly.

PR #16287, opened 2026-02-04
[IMP] l10n-ae: modules and other minor updates

afma-odoo:
task-5098009

**This update only goes from 18.0 to 18.4 (with some minor changes in 18.4)**


auva-odoo:
@robodoo up to saas-18.4

PR #16306, opened 2026-02-05
[IMP] Inventory: Storage categories sublocation update

erjer-odoo:
While updating the storage categories tutorial, we received clarification on how sublocations work in putaway rules. This PR introduces that information.

This 18.0 PR can be FWP up to master. As a note to myself, I will have to check the FWPs for 18.3+ to make sure outdated UI language doesn't make its way into the use case section; if it does, I'll fix it in the FWPs and submit an override for those fixes.


erjer-odoo:
@huisit This PR is also ready for review when you get a moment. Thank you!

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
Felicious, line 36: Since the purpose of this PR is to address the finicky and complex configuration of storage locations, 
I suggest adding a dedicated subsection here (for example, “Storage location setup”) that explicitly explains how sublocations are expected to work.

The subsection should guide the reader by going to inventory > Settings > Locations page, then explain the intended setup with a concrete example (i came up with my own, feel free to do your own!):

- Suppose we have Shelf A, Shelf B, and Shelf C, which are locations where we can store our lemonade
- If the same storage category should apply to all three shelves, is a **parent locatin** required?
- If so, the documentation should clearly instruct users to nest Shelves A to C under a parent location and explain that the storage category will be applied to that parent, not the individual shelves

The goal is to remove ambiguity about where the storage category should be set and why.

(also this is just my assumption of how this works based on our discussions. if im wrong, ignore me!! this ask is just to further clarify how the sublocations need to set up in order for the storage category to work :D )

As a validation check: @huisit, please review whether the documentation as a whole (not just these lines changed), as written, is sufficient to correctly configure the following runbot scenario:

- The storage category is set only on the parent location
- Shelf A is full
- Odoo correctly routes the product to Shelf B

If the documentation does not make this setup obvious and reproducible, it's a sign the article might need to be clarified!

Felicious (commented):
adding a comment pre-emptively (:

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
erjer-odoo: Based on the answers I got from the SME, this understanding isn't fully correct.

Here's how it's supposed to work:

* To set a storage category on Shelves A, B, and C, they require a parent location. This location can be as simple as `WH/Stock`, but can be customized as necessary, **as long as they are all child locations/sublocations of whatever the parent location is**. Then, set the storage category for all of those child locations to the same storage category. There doesn't appear to be any fallback for setting it on the parent location.
* When it comes to putaway rules, the parent location needs to be the **Store to** location. To ensure that the storage category with its child locations are used, the **Sublocation** needs to be set to **Closest Location**, with the **Having Category** set to the storage category itself. Every location for that **Having Category** must be children of the **Store to** location.

I attempted to clarify this in the documentation. Hopefully it's as clear as it can be!

erjer-odoo:
@huisit This is ready for review again when you get a moment! Thank you!

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
huisit: we have two admonition blocks in a row here, is there a reason this has to be a note?

huisit:
```suggestion
Then, in the *Warehouse* section, ensure the :guilabel:`Storage Locations` and
```

huisit:
```suggestion
To set capacities by :ref:`package type <inventory/routes/set-capacity-package>`, also
```

huisit: is this the **packages** checkbox under the _operations_ section?

huisit:
```suggestion
Enable storage categories
```

huisit: i think this could be the first paragraph of the article, so that we're leading with an explanation of exactly what a storage category is + how it addresses a user need

huisit: IMO a more descriptive heading would be helpful, since this article is practically all configuration steps for a passive feature

this would also reflect the numbered steps list at the start of the article

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -36,6 +36,33 @@ make sure :guilabel:`Packages` is enabled. Click :guilabel:`Save`.
 .. image:: storage_category/enable-categories.png
    :alt: Enable Storage Locations and Multi-Step Routes to enable storage categories.
 
+Storage location setup
+======================
+
+Set up storage locations to work with the storage category. Go to :menuselection:`Inventory app -->
```
huisit: the workflow in this section implies that storage locations need to be under one parent location, but it might help to explicitly state it here as well

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -36,6 +36,33 @@ make sure :guilabel:`Packages` is enabled. Click :guilabel:`Save`.
...
+Configuration --> Locations`.
+
+First, set up a parent location. This can be as simple as the default `WH/Stock` location.
+Alternatively, create a new parent location by clicking the :guilabel:`New` button on the
+:guilabel:`Locations` page.
+
+On the location form, specify a :guilabel:`Location Name` and :guilabel:`Parent Location`. Select
+:guilabel:`Internal Location` as the :guilabel:`Location Type`.
```
huisit:
```suggestion
On the location form, specify a :guilabel:`Location Name` and :guilabel:`Parent Location`. Select
:guilabel:`Internal` as the :guilabel:`Location Type`.
```

huisit: combine with previous paragraph so it's clear this is configuration for the parent location, or specify "On the _parent_ location form"

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -36,6 +36,33 @@ make sure :guilabel:`Packages` is enabled. Click :guilabel:`Save`.
...
+
+Then, create child locations (called a *sub-location*) of this parent location by clicking the
```
huisit: between child and sub-location, is there an official/more frequently used term? IMO we should stick to one term to avoid confusion

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -36,6 +36,33 @@ make sure :guilabel:`Packages` is enabled. Click :guilabel:`Save`.
...
+:guilabel:`New` button. On the location form, specify a :guilabel:`Location Name`, and set the
+:guilabel:`Parent Location` to the parent location that was just created.
+
+.. example::
+   A beverage company stores all of its cans of lemonade on pallets in one section of its warehouse.
+
+   First, they create a location named `Pallets` in the `WH/Stock` location. Then, they create two
+   sub-locations, named `PAL1` and `PAL2`. These child locations both have the parent location of
```
huisit: update if using "sub-location" instead of "child"

huisit: "both have `WH/Stock/Pallets` set as the parent location" might be more straightforward

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -36,6 +36,33 @@ make sure :guilabel:`Packages` is enabled. Click :guilabel:`Save`.
...
+   `WH/Stock/Pallets`.
+
+   .. image:: storage_category/new-child-location.png
+      :alt: Create a sub-location.
+
 .. _inventory/routes/define-storage:
 
 Define storage category
```
huisit:
```suggestion
Define storage category limitations
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
huisit:
```suggestion
Before a storage category is applied to locations, it must be configured with specific limitations in order to decide the optimal storage location. Capacity can be limited by weight, product, and package type.
```

huisit:
```suggestion
```

huisit:
```suggestion
.. note::
Weight limits can be combined with capacity by package or product (e.g. a maximum of one hundred
products with a total weight of two hundred kilograms).

While it is possible to limit capacity by product and package type at the same location, it may
be more practical to store items in different amounts across various locations, as shown in the
:ref:`limit capacity by package <inventory/routes/set-capacity-package>` example.
```

huisit: consider moving to weight section

huisit: the reader hasn't assigned the category to any locations yet at this point, consider moving to assign section

huisit, line 105: this exact admonition block is also presented in the package section. is it a duplicate?

huisit:
```suggestion
Limit capacity by weight
```

huisit:
```suggestion
Limit capacity by product
```

huisit:
```suggestion
Limit capacity by package type
```

huisit:
```suggestion
A maximum product weight can be set in the :guilabel:`Max Weight` field. This limit applies
to each location assigned this storage category. If a product weight is defined, the value must be set to greater than `0`.
```

huisit: assuming the reader is on the same form as the previous section. will not be as short if the `note` admonition block is moved here

huisit:
```suggestion
In the :guilabel:`Capacity by Product` tab, click :guilabel:`Add a Line` to enter a product, and set
the maximum capacity that should be stored at each location in the :guilabel:`Quantity` field.
```

huisit: just a quick suggestion, could be workshopped. assuming that all 3 of these capacity limits apply to every location in the category?

huisit:
```suggestion
To ensure only a maximum of five `Large Cabinets` and two `Corner Desk Right Sit` are stored at a single storage location, specify those amounts in the :guilabel:`Capacity by Product` tab.
```

huisit, line 143:
```suggestion
Limiting capacity by :doc:`package <../../product_management/configure/package>` makes it
possible to enforce real-time storage capacity checks based on package type (e.g., crates, bins,
boxes, etc.).
```

huisit: > real-time storage checks

do we know exactly how this is performed? is there a separate feature for "real-time storage capacity checks" that the user can manually run?

otherwise if i'm understanding the information in the `important` admonition below, it sound like the opposite -- odoo does not perform capacity checks when using storage categories, so it is possible to overallocate for a given location.

huisit:
```suggestion
Click :guilabel:`Add a line` to add a :ref:`package type <inventory/warehouses_storage/package-type>` to the storage category. Alternatively, create a new package type on the :guilabel:`Inventory` tab of the product form (in the
:guilabel:`Packaging` section), or from the :guilabel:`Product Packagings` page.
```

huisit, line 161: putaway rules haven't been mentioned yet in the article at this point. consider moving this content to the putaway rule section -- the `example` admonition there even says "Continuing the example from above" which asks the reader to recall previous context 😬

huisit: and if you do move this content, remember to update the ref in the define category/limit by weight section's `note` admonition

huisit:
```suggestion
Once the storage category is created, it can be assigned to locations. Go to
```

huisit:
```suggestion
:menuselection:`Inventory app --> Configuration --> Locations`, and select the desired locations.
```

huisit:
```suggestion
.. example::
Assign the **High frequency pallets** storage category (which limits pallets stored at any location
to two pallets) to the **WH/Stock/Pallets/PAL1** sub-location.
```

huisit: bold for the location (latter)? not sure if it would be considered a `user input` or _title_

huisit:
```suggestion
Assign storage locations
```

huisit:
```suggestion
Create a putaway rule
```

huisit:
```suggestion
With the :ref:`storage category <inventory/routes/define-storage>` and :ref:`locations
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -166,14 +193,16 @@ to :menuselection:`Inventory app --> Configuration --> Putaway Rules`.
 Click the :guilabel:`New` button to create the putaway rule. Specify a location to store to in the
```
huisit:
```suggestion
Click the :guilabel:`New` button to create the putaway rule. Specify a location in the
```

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
huisit, line 219:
```suggestion
Continuing the example from above, the **High frequency pallets** storage category is assigned to
the putaway rule directing pallets of lemonade to locations with the **High frequency pallets**
```

huisit, line 230: this example seems like an extension of the package or putaway rule section? consider consolidating

if i'm misreading and it fulfills a unique use case, renaming the heading could help clarify

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
erjer-odoo: Just as a note to self, I didn't accept this suggestion because the "Location" menu item text didn't get removed until 19.0. This is something to address in the FWP.

The other reply from huisit is valid and I will be addressing it after committing the rest of the batch changes.

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
erjer-odoo: I did that just in case someone jumped to this section in the document without reading the package section.

erjer-odoo: My understanding is that it does check capacity in real time as products are transferred in and out of, and around, the warehouse, but it's only checking whether a storage location is **full**. It treats incoming shipments of packages as a single unit -- so if the capacity is 2 packages, and there's already 1 on the shelf, and 2 more packages come in on the same PO, **it will treat the incoming packages as a single unit of 1 package**. That's why we need to manually break it up.

erjer-odoo: I probably need to clarify that we are opening a desired location, not selecting multiple on the Locations page.

erjer-odoo: The way this was written before (and the option I agree with) treats it as user input.

erjer-odoo: Same note as before, would still prefer to treat it like user input.

erjer-odoo: My belief is that this is a unique use case that happens frequently -- when you've set capacity by package and have varying levels of products in stock and incoming, this is specifically how Odoo behaves for each possibility.

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
erjer-odoo: Sublocation is the phrase used in putaway rules, so I think I'll be clearer about it throughout where possible.

erjer-odoo:
@Felicious This should be ready for final review now that I've had a chance to address @huisit 's very thorough and helpful review :)

content/applications/inventory_and_mrp/inventory/shipping_receiving/daily_operations/storage_category.rst
```diff
@@ -16,49 +19,70 @@ Follow these steps to complete the setup:
...
+make sure :guilabel:`Packages` is enabled in the *Operations* section. Click :guilabel:`Save`.
 
 .. image:: storage_category/enable-categories.png
    :alt: Enable Storage Locations and Multi-Step Routes to enable storage categories.
 
+Storage location setup
+======================
+
+Set up storage locations to work with the storage category. A parent location must be set up, with
+child locations (or *sublocations*) assigned to it.
```
Felicious: I thought your reply to me was super clear, @erjer-odoo ! I'm thinking that context would be helpful to share in the documentation as well (: I'm considering adding this "why" statement in. Feel free to reject this feedback if you think this doesn't belong in documentation. 


```suggestion
child locations (or *sublocations*) assigned to it. This is because the putaway rule's :ref:`Store to <inventory/routes/set-putaway-attribute>` location will be set to the parent location, and the :ref:`storage category <inventory/routes/assign-location>` will be set to the child location.
```

Felicious (approved):
Great work, @erjer-odoo !

Appreciated you getting clear answers from the SMEs and clearing up my confusion about how this works!! I just have one optional suggestion for you, and this is ready to go! 

@robodoo delegate+

erjer-odoo:
@robodoo r+

PR #16389, opened 2026-02-13
[IMP] l10n-it: E-invoicing process update for 18.2->18.4

afma-odoo:
task-4710859

**This update only goes from 18.2 to 18.4**


afma-odoo:
@robodoo up to saas-18.4

PR #16434, opened 2026-02-15
[ADD] markdown_builder: custom Sphinx builder for Markdown output

idealson-nexources:
Add a custom Sphinx builder that compiles the RST documentation to Markdown files, preserving directory structure, cross-references as relative .md links, and code block language annotations. This is needed because sphinx-markdown-builder is incompatible with the pinned Sphinx 4.3.2.

New files:
- extensions/markdown_builder/ (builder, translator, init)
- _conf_markdown/conf.py (separate Sphinx config for markdown build)
- Makefile: `make markdown` target
- README.md: build instructions for markdown output
- requirements.txt: comment documenting the custom extension

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ html: $(HTML_BUILD_DIR)/_static/style.css
 	$(SPHINX_BUILD) -c $(CONFIG_DIR) -b html $(SPHINXOPTS) $(SOURCE_DIR) $(HTML_BUILD_DIR)
 	@echo "Build finished."
 
-MD_CONFIG_DIR  = _conf_markdown
+MD_CONFIG_DIR  = extensions/markdown_builder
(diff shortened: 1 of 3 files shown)


Feyensv:
> This is needed because sphinx-markdown-builder is incompatible with the pinned Sphinx 4.3.2.

Well, this documentation is meant to be build in html, and you do not provide any reason we should support markdown rendering.
We certainly don't want to maintain a separate builder only for that, but if there are good reasons, we can consider increasing/adapting the pinned version.
A specific Sphinx version is pinned to reduce maintenance and adaptation costs in our internal processes, but we increase it when needed.

AntoineVDV:
Same as for https://github.com/odoo/documentation/pull/16433

PR #16567, opened 2026-02-26
[FIX] l10n_in: update gsp configuration steps

vivp-odoo:
Improve the documentation to provide clear step-by-step guidance for configuring the GSP provider in Indian Accounting.

Add detailed instructions to help users correctly set up the credentials and re-trigger the e-Invoice and e-Waybill flows. Include a configuration screenshot for better clarity.

task-5960485

### DIFF
--- a/content/applications/finance/fiscal_localizations/india.rst
+++ b/content/applications/finance/fiscal_localizations/india.rst
@@ -45,13 +45,16 @@ GSP configuration
 
 To use the :ref:`e-Invoicing <india/e-invoicing>`, :ref:`E-Way bill <india/e-waybill>`, and
 :ref:`GST return filing <india/gstr>` services in Odoo, configure :guilabel:`BVM IT Consulting
-Services India Private Limited` as the :abbr:`GSP (GST Suvidha Provider)`. To do so, go to
-:menuselection:`Accounting --> Configuration --> Settings`, scroll down to the :guilabel:`Select
-GSP Provider` section, and set the :guilabel:`GSP` field to :guilabel:`BVM IT Consulting`.
+Services India Private Limited` as the :abbr:`GSP (GST Suvidha Provider)`. To do so,
+follow these steps :
 
-.. note::
-   If you do not see this option in the Accounting settings, :ref:`upgrade <general/upgrade>` the
-   :guilabel:`Indian - Accounting` (`l10n_in`) module.
+#. :ref:`Upgrade <general/upgrade>` the :guilabel:`Indian - Accounting` (`l10n_in`) module.
+#. Go to :menuselection:`Accounting --> Configuration --> Settings`.
+#. Scroll down to the :guilabel:`Select GSP Provider` section, and set the :guilabel:`GSP` field to :guilabel:`BVM IT Consulting`.
+#. :ref:`Log into the NIC e-invoice portal <india/e-invoicing-api>` and
+   :ref:`NIC E-Way bill portal <india/e-waybill-api>` and create a new API user for each.
+#. Go to :menuselection:`Accounting --> Configuration --> Settings`, scroll down to
+   the :guilabel:`Indian Integration` section, and enter the new :guilabel:`Username` and :guilabel:`Password` for the :ref:`e-Invoicing <india/e-invoicing-configuration>` and :ref:`E-Way bill <india/e-waybill-configuration>` features.
 
 .. _india/e-invoicing:
 


content/applications/finance/fiscal_localizations/india.rst
```diff
@@ -49,9 +49,16 @@ Services India Private Limited` as the :abbr:`GSP (GST Suvidha Provider)`. To do
...
+.. image:: india/gsp-configuration.png
+   :alt: GSP configuration
+
+Please follow the steps below:
+
+#. Upgrade the :guilabel:`Indian - Accounting` (`l10n_in`) module from App menu.
+#. Visit the E-Invoice portal and follow the steps mentioned in the :ref:`NIC e-Invoice registration section <india/e-invoicing-api>` to get API credentials
+   for e-Invoicing and configure it in odoo :ref:`e-invoice configuration <india/e-invoicing-configuration>`.
+#. Visit the E-Way bill portal and follow the steps mentioned in the :ref:`API registration on NIC E-Way bill section <india/e-waybill-api>` to get API credentials
+   for E-Way bill and configure it in odoo :ref:`e-waybill configuration <india/e-waybill-configuration>`.
```
auva-odoo: I suggest we keep this section focused on GSP configuration, see the e-invoicing and e-way bill instructions are already explained in another section (referenced above).
Also, since customers need to upgrade the module before they can set the GSP provider, I suggest to change the whole GSP configuration section as follows (replace lines 46-61 with the following):

```suggestion
To use the :ref:`e-Invoicing <india/e-invoicing>`, :ref:`E-Way bill <india/e-waybill>`, and
:ref:`GST return filing <india/gstr>` services in Odoo, configure :guilabel:`BVM IT Consulting
Services India Private Limited` as the :abbr:`GSP (GST Suvidha Provider)`. To do so, follow these steps:

#. :ref:`Upgrade <general/upgrade>` the :guilabel:`Indian - Accounting` (`l10n_in`) module.
#. Go to :menuselection:`Accounting --> Configuration --> Settings`.
#. Scroll down to the :guilabel:`Select GSP Provider` section, and set the :guilabel:`GSP` field to :guilabel:`BVM IT Consulting`.

   .. image:: india/gsp-configuration.png
      :alt: GSP configuration
```
And if you think it's absolutely necessary to mention NIC e-invoice registration and NIC E-Way bill registration, I would just add a note or tip under this paragraph saying something like (no need to repeat the instructions already mentionned in those sections):

```suggestion
.. note::
   :guilabel:`BVM IT Consulting Services India Private Limited` must also be set as GSP in the :ref:`NIC e-Invoice <india/e-invoicing-api>` and :ref:`NIC E-Way bill<india/e-waybill-api>` portals.
```

vivp-odoo:
Hello @auva-odoo
Thanks for the review, 
A few points I want to mention:
1. Actually, we are in transition to change our GSP from the previous Tera software to BVM IT Consulting
    and we forcefully want the user to change the GSP.   
2. User can see the option to change GSP in the configuration only if they update the module. Once they update it or
    already updated it, there is no need to update. So this step is not compulsory every time, not for a new user because they
    get the updated module already.
3. Once they switch to the new GSP provider, they need to go to the e-invoice and e-way bill gov portal  (NIC e-invoice )
    to get there API credentials (username and password) once they get it they need to add that same in odoo. I already
    pointed out this in my 2nd and 3rd steps.
    
    let me know your suggestion based on this points

vivp-odoo:
I have just modified the first point with only an update if you do not see the  GSP configuration. I'll change the remaining part if needed after your re-suggestion on my points above.

auva-odoo:
Hi @vivp-odoo, I'm not sure I understand why my suggestion does not work based on the different points you mention:

> 1. Actually, we are in transition to change our GSP from the previous Tera software to BVM IT Consulting
>    and we forcefully want the user to change the GSP.

--> I do understand this but I think my suggested change reflects that?

> 2. User can see the option to change GSP in the configuration only if they update the module. Once they update it or
>    already updated it, there is no need to update. So this step is not compulsory every time, not for a new user because they 
>    get the updated module already.

--> New users will probably not be in 17.0, will they? And if you think the user might have already updated the module, maybe we can just add "If needed, [upgrade the xxx module]"?

> 3. Once they switch to the new GSP provider, they need to go to the e-invoice and e-way bill gov portal  (NIC e-invoice )
>    to get there API credentials (username and password) once they get it they need to add that same in odoo. I already
>    pointed out this in my 2nd and 3rd steps.

--> Yes but those configurations are already explained in the referenced section so I think a note (like I suggested in my comment) would suffice.

vivp-odoo:
Hello @auva-odoo
Acutally how will user decide if he needs to update or not first user need to go to setting to check if gsp configuration is visible or not.
can we do something like make step 2 as step 1 because first user need to go to setting to switch gsp and then if they don't see the gsp configuration then go to app menu and update the module.
and keep the image above so we can tell write something like `if you do not see gsp configuration like above image` go to app menu to update
and we can simply remove step three and merge it with step 2.

In summary.
1. go to setting
2. check for gsp configuration and switch to bvm if do not see then update the module.
3. go to e-invoice and e-way bill portal to generate api credentials as show in below reference (link of both references)

what do you think ?

vivp-odoo:
I know it's not big deal either way :slightly_smiling_face:  but we kept on getting support tickets after our first PR on the gsp changes. so it seems user need way more clear steps.

auva-odoo:
OK, but again if people are in 17.0, they will probably need to upgrade their module in any case, won't they? Otherwise I don't understand why we need to change the doc, it should be fine as it was before this PR, except for the note about the e-invoice and e-way bill portals?

vivp-odoo:
Actually, my purpose was to keep the documentation the same for all the versions from 17.0.
But no issue i can do your suggestions till 19.0, and from 19.0 we can add upgrade module step optional.

vivp-odoo:
Hello @auva-odoo
I added the e-invoce and e-way bill part in steps and not as note because as per spec PO don't like note here for update module before this pr. so kept this as part of steps.
Kindly let me know if this seems fine too you.

auva-odoo:
> Hello @auva-odoo I added the e-invoce and e-way bill part in steps and not as note because as per spec PO don't like note here for update module before this pr. so kept this as part of steps. Kindly let me know if this seems fine too you.

I'm sorry, I really don't understand why. This section focuses on GSP configuration; any other configuration (e-invoice/e-way bill) have dedicated sections already, which mention this specific point, so IMO there is no need to repeat it here and a note is sufficient. The note that was there before this PR was about upgrading the module and you moved that content to a dedicated step in the instructions, so I don't understand how that feedback about the previous note applies here in this PR?

> Actually, my purpose was to keep the documentation the same for all the versions from 17.0. But no issue i can do your suggestions till 19.0, and from 19.0 we can add upgrade module step optional.

Yes let's do that, a simple "if needed" at the beginning of that instruction in the list will be enough I think.

content/applications/finance/fiscal_localizations/india.rst
```diff
@@ -45,13 +45,18 @@ GSP configuration
...
+
+#. :ref:`Upgrade <general/upgrade>` the :guilabel:`Indian - Accounting` (`l10n_in`) module.
+#. Go to :menuselection:`Accounting --> Configuration --> Settings`.
+#. Scroll down to the :guilabel:`Select GSP Provider` section, and set the :guilabel:`GSP` field to :guilabel:`BVM IT Consulting`.
+#. Follow the instructions to :ref:`register on the NIC portal <india/e-invoicing-api>` and
+   :ref:`NIC E-Way bill portal <india/e-waybill-api>` to get your API credentials.
+#. configure username and password for GSP in the :ref:`e-Invoicing <india/e-invoicing-configuration>` and :ref:`E-Way bill <india/e-waybill-configuration>` sections.
+
+.. image:: india/gsp-configuration.png
+   :alt: GSP configuration
```
auva-odoo: and remove this screenshot please

content/applications/finance/fiscal_localizations/india.rst
```diff
@@ -45,13 +45,18 @@ GSP configuration
...
-   :guilabel:`Indian - Accounting` (`l10n_in`) module.
+Services India Private Limited` as the :abbr:`GSP (GST Suvidha Provider)`. To do so,
+follow these steps :
+
+#. :ref:`Upgrade <general/upgrade>` the :guilabel:`Indian - Accounting` (`l10n_in`) module.
+#. Go to :menuselection:`Accounting --> Configuration --> Settings`.
+#. Scroll down to the :guilabel:`Select GSP Provider` section, and set the :guilabel:`GSP` field to :guilabel:`BVM IT Consulting`.
+#. Follow the instructions to :ref:`register on the NIC portal <india/e-invoicing-api>` and
+   :ref:`NIC E-Way bill portal <india/e-waybill-api>` to get your API credentials.
+#. configure username and password for GSP in the :ref:`e-Invoicing <india/e-invoicing-configuration>` and :ref:`E-Way bill <india/e-waybill-configuration>` sections.
```
auva-odoo:
```suggestion
#. :ref:`Log into the NIC e-invoice portal <india/e-invoicing-api>` and
   :ref:`NIC E-Way bill portal <india/e-waybill-api>` and create a new API user for each.
#. Go to :menuselection:`Accounting --> Configuration --> Settings`, scroll down to the :guilabel:`Indian Integration` section, and enter the new :guilabel:`Username` and :guilabel:`Password` for the :ref:`e-Invoicing <india/e-invoicing-configuration>` and :ref:`E-Way bill <india/e-waybill-configuration>` features.
```

auva-odoo (approved):
@robodoo r+

PR #16572, opened 2026-02-26
[IMP] add highlight fade to ref targets

nikibrown:
Task: https://www.odoo.com/odoo/project/3835/tasks/5970579

When a user clicks on a ref link and the page scrolls to a heading its easy to not see the specific heading. This PR adds a temporary highlight to the heading. Styling is similar to the search term highlighting that currently exists. See screenshot and example below:

ref-highlight
search-highlight

### DIFF
--- a/extensions/odoo_theme/static/style.scss
+++ b/extensions/odoo_theme/static/style.scss
@@ -973,6 +973,29 @@ span.viewcode-link {
 }
 
 //------------------------------------------------------------------------------
+// Ref target highlighting
+//------------------------------------------------------------------------------
+
+span:target + h1,
+span:target + h2,
+span:target + h3,
+span:target + h4 {
+    animation: highlight-fade 5s ease-out 0s 1;
+}
+
+@keyframes highlight-fade {
+    0% {
+        background-color: lighten($warning, 15%);
+    }
+    50% {
+        background-color: lighten($warning, 15%);
+    }
+    100% {
+        background-color: transparent;
+    }
+}
+
+//------------------------------------------------------------------------------
 // Alerts
 //------------------------------------------------------------------------------
 


nikibrown:
@AntoineVDV 

Thanks for the review. A few questions to move this forward:

1. Is there a line count or complexity guideline for UI improvements? Happy to work within it.
2. Would a pure CSS :target approach (no translator patching) address the maintenance concern? 
2a. Gif of CSS only approach (highlights the whole line)
full-line-hilight
3. Is this a "not now" or "never"? If there's a milestone where tooling work is in scope, I can retarget.

AntoineVDV:
> 1. Is there a line count or complexity guideline for UI improvements? Happy to work within it.

Not really. It's more a rule of common sense where we must weigh the impact and the cost. Here, the impact is low, IMHO, and the cost is not trivial because the translator is the # 1 touchy thing in the documentation.

> 2. Would a pure CSS :target approach (no translator patching) address the maintenance concern?

Except if it requires hundreds of new lines of CSS to maintain, yes it would be better than touching the translator.

> 3. Is this a "not now" or "never"?

At the moment and for the foreseeable future, it's a no, because:
1. The documentation content itself is far from being mature, and we should make it a priority over low-impact work;
2. The development resources available to work on the documentation are currently about ~0.01 full-time devs. This means we can't afford to spend countless hours reviewing, maintaining, fixing, and migrating to future versions of Sphinx a gread deal of technical tweaks.

nikibrown:
@AntoineVDV Thanks for the additional info re: translator being touchy. I've removed my edits to this file and made some additional updates to have the same effect with just CSS changes. 

Re: maintenance and upkeep: I'm happy to help contribute to the upkeep of this repo as I continue to learn how things work.  

I disagree that this is low-impact. This solves a legitimate UX problem. Our pages are long (especially on mobile) and have lots of headings. Its easy to lose your place when clicking on an in page link. Google chrome does something similar in search results with scroll to text fragments. I think our use cases are similar: https://github.com/WICG/scroll-to-text-fragment?tab=readme-ov-file#motivating-use-cases

The goal is to help the user find the relevant information they are looking for faster. 

See updated gif of changes

hightlight-v3

edi-odoo:
Hi @nikibrown 
I'm not sure what the issue with the current ui is.
ezgif com-optimize

>When a user clicks on a ref link and the page scrolls to a heading its easy to not see the specific heading

The heading is right at the top of the page, you can't really miss it

>Its easy to lose your place when clicking on an in page link

The right menu scrolls to show you exactly where you are in the page. 

What does this PR fix ? Maybe I'm misunderstanding something.

nikibrown:
@edi-odoo This is just a UI suggestion for improvement from a conversation with some of the technical writers. 

There are lots of mismatches in link text / heading name. Easy to visually get lost in longer pages. Especially if a heading is at the bottom of the page and can't be scrolled to the top.  Example - Navigate here: Click on `group by` link in the first paragraph. It links to a heading called `Group records` which is a mismatch  https://runbot281.odoo.com/runbot/static/build/103181960-17-0/logs/build/html/applications/essentials/search.html 

As I previously mentioned google chrome does this with search results.  Here's an example: https://www.odoo.com/documentation/19.0/applications/essentials/stages.html#:~:text=the%20progress%20bar.-,Add%20email%20and%20SMS%20templates%20to%20stages,-When%20an%20Email

Thoughts?

extensions/odoo_theme/static/style.scss
```diff
@@ -972,6 +972,45 @@ span.viewcode-link {
...
+//------------------------------------------------------------------------------
+// Ref target highlighting
+//------------------------------------------------------------------------------
+
+span:target {
+    scroll-margin-top: $o-headers-mobile-height;
+    @include media-breakpoint-up(lg) {
+        scroll-margin-top: $o-headers-height;
+    }
+}
```
edi-odoo:
```suggestion
```

extensions/odoo_theme/static/style.scss
```diff
@@ -972,6 +972,45 @@ span.viewcode-link {
...
+span:target + h3,
+span:target + h4 {
+    animation: highlight-fade 5s ease-out 0s 1;
+    display: inline-block;
+
+    // show the permalink icon
+    a.headerlink:after {
+        visibility: visible;
+        opacity: 1;
+    }
```
edi-odoo:
```suggestion
```

extensions/odoo_theme/static/style.scss
```diff
@@ -972,6 +972,45 @@ span.viewcode-link {
...
+}
+
+// Same background color as span.highlight from Search items highlighted
+@keyframes highlight-fade {
+    0% {
+        background-color: lighten($warning, 15%);
+    }
+
```
edi-odoo:
```suggestion
```

nikibrown:
@edi-odoo 

Appreciate the back and forth on these tiny UI details!

Thanks for spotting the border! 👀 I didn't even see the border bottom on the heading since #D5D5D5 is not that much contrast on a white background. 

Thoughts on highlighting the whole line vs just the border? The border is hard to see even when its fading out orange. 

highlight-take4

border-animation

edi-odoo:
@nikibrown thanks for testing it.
Indeed the border animation isn't visible enough this way. I was thinking more along the lines of this for example:
https://codepen.io/SMLMRKHLMS/pen/xxyPMp

But I don't mind the background being on the whole line either. I'll let you decide what you want to do. Either have fun with the border (loads of examples on codepen and the likes) or just keep the background . 

Note that we do want to keep it short though ;) 

> I didn't even see the border bottom on the heading since #D5D5D5 is not that much contrast on a white background.

Is your screen maybe too bright or not calibrated properly?

nikibrown:
@edi-odoo Can you take a look at my last commit? Simplified per your suggestions.

edi-odoo:
LGTM @robodoo delegate+

nikibrown:
@robodoo r+

StraubCreative:
@robodoo r+

PR #16588, opened 2026-02-27
[IMP] howtos/website_theme: Adapts Theming chapter

bram-odoo:
This PR adapts the "Theming" chapter to Odoo 19.

Task-5237456


content/developer/howtos/website_themes/theming.rst
```diff
@@ -900,8 +933,8 @@ brings the benefit of a better developer experience with better integration with
    folder of your module.
 
 .. tip::
-   - Use a linter (JSHint, ...).
-   - Never add minified JavaScript libraries.
+   - Use a linter (ESLint, ...).
+   - Always add minified JavaScript libraries
    - Add `'use strict';` at the top of every old-style module (this is automatic for new-style
```
duau-odoo: I should have probably removed it in my PR as well, but we don't do `use strict;` anymore.

content/developer/howtos/website_themes/theming.rst
```diff
@@ -900,8 +933,8 @@ brings the benefit of a better developer experience with better integration with
    folder of your module.
 
 .. tip::
-   - Use a linter (JSHint, ...).
-   - Never add minified JavaScript libraries.
+   - Use a linter (ESLint, ...).
```
duau-odoo: We can maybe include a link to the ESLint config somewhere?
https://github.com/odoo/odoo/blob/19.0/addons/web/tooling/_eslintrc.json

content/developer/howtos/website_themes/theming.rst
```diff
@@ -915,11 +948,11 @@ brings the benefit of a better developer experience with better integration with
    - Use double quotes for all textual strings (such as `"Hello"`) and single quotes for all other
```
duau-odoo: Now we use double quotes for every string, easier to do (also it's handled by eslint)

content/developer/howtos/website_themes/theming.rst
```diff
@@ -900,8 +933,8 @@ brings the benefit of a better developer experience with better integration with
    folder of your module.
 
 .. tip::
-   - Use a linter (JSHint, ...).
-   - Never add minified JavaScript libraries.
+   - Use a linter (ESLint, ...).
+   - Always add minified JavaScript libraries
    - Add `'use strict';` at the top of every old-style module (this is automatic for new-style
      modules).
    - Use `js_` prefixed CSS classes on elements you target with JavaScript.
```
duau-odoo, line 949: There are still some `js_` class in the codebase but technically it's not a "rule" anymore as you can see for most of the `Interaction`s for example, `selector` is not `js_*`

Also for variable name (I can't comment on the line directly), we still use `camelCase` and for classes we use PascalCase

content/developer/howtos/website_themes/building_blocks.rst
karlosvalverde, line 1: This should go [here](https://github.com/odoo/documentation/pull/16723)

content/developer/howtos/website_themes/theming.rst
```diff
@@ -692,6 +695,23 @@ structure below.
...
+         <field name="shop_opt_products_design_classes">
+               o_wsale_products_opt_layout_catalog
+               o_wsale_products_opt_design_cards
+               o_wsale_products_opt_name_color_regular
+               o_wsale_products_opt_thumb_cover
+               o_wsale_products_opt_img_hover_zoom_out_light
+               o_wsale_products_opt_has_cta
+               o_wsale_products_opt_actions_onhover
+               o_wsale_products_opt_wishlist_fixed
+               .o_wsale_products_opt_cc
```
karlosvalverde: Remove the `.` here

content/developer/howtos/website_themes/theming.rst
```diff
@@ -692,6 +695,23 @@ structure below.
...
+               o_wsale_products_opt_cc1
+               o_wsale_products_opt_rounded_2
+               o_wsale_products_opt_actions_promote
+               o_wsale_products_opt_actions_subtle
+               o_wsale_products_opt_name_weight_bold
+         </field>
```
karlosvalverde, line 714: I would add more eCommerce fields, such as `shop_page_container`, `product_page_container`  and `product_page_image_width`.

I would mention on the Website settings paragraph that now, there are several eCommerce option that have to be specified within `website.xml`.

content/developer/howtos/website_themes/theming.rst
```diff
@@ -911,20 +935,18 @@ brings the benefit of a better developer experience with better integration with
...
-   - Use a linter (ESLint, ...).
-   - Always add minified JavaScript libraries.
-   - Add `'use strict';` at the top of every old-style module (this is automatic for new-style
-     modules).
+   - Use a linter (`ESLint <{GITHUB_PATH}/addons/web/tooling/_eslintrc.json>`_, ...).
+   - Always add minified JavaScript libraries
    - Use `js_` prefixed CSS classes on elements you target with JavaScript.
    - Variables and functions should be *camelCased* (`myVariable`) instead of *snake_cased*
      (`my_variable`).
+   - JavaScript classes use *PascalCase* (`MyClass`).
```
karlosvalverde, line 952: Is it worth also mentioning to use `*kebab-case* (`my-file.xml`)` for XML and regular file names (.svg, .webp...) and `*snake-case* (`my_js_file.js`)` for programming language files (.js, .py, etc)?

duau-odoo: In the website codebase we are not using kebab-case anymore, there are still some legacy files (some .gif, .png, etc), but everything should use snake_case : .xml, .js, .py, .webp, .png, .svg, etc

karlosvalverde: > In the website codebase we are not using kebab-case anymore, there are still some legacy files (some .gif, .png, etc), but everything should use snake_case : .xml, .js, .py, .webp, .png, .svg, etc

So this affects so far this [one](https://github.com/odoo/documentation/pull/16589/changes/45abae01dbcd234a3f16aa883e2a63dc41ba0e92#diff-d6e25840cb074cadbdc8b3de480f5293ea13782c382819483c5a9e4c20c40575R722) and this [guy](https://github.com/odoo/documentation/pull/16590/changes/f3728d900f2547f6f3fa1fef2b35bfd37e202f36#diff-652a6b5f5f0bab1f64bae9c6f3bddfe34bd540cc74ac6d9c107f3d6c36b5d0a1R270) too...

content/developer/howtos/website_themes/theming.rst
karlosvalverde: Actually, I would add a link or a `See More` linked to [this](https://github.com/odoo/documentation/pull/16723/changes/a3c9290574c3c983705ec37ddfae98163a00fbec#diff-83b6712da3c9a2b3ef07c080079ba76c683ee2131a058ce24080583694c51a11R1429)

content/developer/howtos/website_themes/theming.rst
bram-odoo: It's actually the correct class name even if it seems incorrect ;-) 

Maybe @AntoineVDV could have more information about this one? 

(FYI @duau-odoo as we discussed about it).

content/developer/howtos/website_themes/theming.rst
bram-odoo: Not exactly: 
- This is not a exhaustive list
- We use the website_airproof case here as an example : `shop_page_container` used the default value `regular` so it is not required to specify again the default as it is already set to it. `product_page_container` used the default `regular` also and same remark for `product_page_image_width`. There are a lot more fields related to the shop/product page. Here we are just talking about the website model and some fields as example but we can't obviously not cover all existing fields :-) 

For the link between the snippet and the shop page, it's a bit more subtle. Page and snippet share some options but not all (Snippet doesn't have the secondary image on hover for example). They also still have two separate templates (with the same look/structure) but as we do not cover the `/shop` page configuration (or any other one from other bridge module like `website_event`, etc) it might be confusing to tell "Oh btw, this page works in a similar way" without deep dive into more explanation (similarities, differences, etc).

content/developer/howtos/website_themes/theming.rst
bram-odoo: @duau-odoo Arf this one is annoying. We use a pre-commit template and we always use snake-case for image file names and SCSS/CSS variables/mixins/functions name (as it's considered as a best practice)
This is a recommandation we had even before I start working here (2022) but indeed the Odoo Coding Guidelines tell: 

- Use a BEM naming for CSS variable (but tbh it's a bit overkill for our custom module even worst when this convention is only applied for CSS variable names and not for the selectors)
- Use kebab-case for SCSS/SASS variable names
- Last but not least (thanks @tocodoo for the reminder!), for SEO purpose "my-image.jpg" is indexed as "my image" but "my_image.jpg" ends to "myimage". 

But in the context of Odoo, the wide majority of apps are only accessible under a login/password so the SEO/Indexation is not taken into account. In fact, website related modules are the only ones to present data publicly so I guess that the website modules were aligned back in the days with the existing codebase.

@karlosvalverde But to "refocus", this tip focuses on "JS guidelines" that are a bit different from the usual (python) convention we meet everywhere else.

content/developer/howtos/website_themes/theming.rst
karlosvalverde: Indeed, I agree we don't need to list all of them, but I think it would be useful to include a side note with some links somewhere indicating that we can also add those from `website_sale` or `website_event`.

@CVDE-odoo What do you think?

content/developer/howtos/website_themes/theming.rst
AntoineVDV: Hi, sorry but I don't :sweat_smile:

content/developer/howtos/website_themes/theming.rst
karlosvalverde: @bram-odoo So that's why I was asking 😜 Indeed, the focus here is JS, but I wonder whether it makes sense to mention this here or somewhere else.

bram-odoo: Maybe more a "See also" pointing to the [Coding Guidelines](https://www.odoo.com/documentation/19.0/th/contributing/development/coding_guidelines.html) but some guidelines are more relevant for the back-office so I don't know if it's totally relevant nor really the purpose of the theming doc. @CVDE-odoo what's your opinion?

content/developer/howtos/website_themes/theming.rst
bram-odoo: I created a ticket for that. I think this is just a typo but prefer to double-check 🫣
https://www.odoo.com/odoo/project/6853/tasks/6047633

bram-odoo: Ok this is a typo that will be fixed :-) I adapted the code accordingly.

content/developer/howtos/website_themes/theming.rst
bram-odoo: I agree but the fact is that there aren't a "clear" file listing the field related to what we explain. We can link to the website.py from website_sale but not all the content is relevant for this doc for example. We can also link to the option template but it will be a bit complicated for some people to understand where to look at (as it only shows the possible values, not the field itself).

karlosvalverde: A link to the `website.py` and even to [this](https://github.com/odoo/documentation/pull/16723/changes/201e46198de340a6fdbaa94eaf74ec88231434bd#diff-83b6712da3c9a2b3ef07c080079ba76c683ee2131a058ce24080583694c51a11R1436) would be enough to me. 👍 

(@CVDE-odoo are you alive? 😄)

bram-odoo:
MIssing label coming from https://github.com/odoo/documentation/pull/16723 (Should be ok after the merge of the related PR).

content/developer/howtos/website_themes/building_blocks.rst
bram-odoo: Done

bram-odoo:
@robodoo r+

PR #16602, opened 2026-02-27
[FIX] Attendances - Update location coordinate details

larm-odoo:

The documentation from 17, 18, and 19 have changed significantly, so making separate PR's for each version.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5980920) for this PR.


larm-odoo:
Hi @Felicious - this is just a FIX request from GMF, so I am sending this to you directly. Thanks!

Felicious:
@robodoo r+

PR #16603, opened 2026-02-27
[FIX] Attendances - Update location coordinate details 18.0

larm-odoo:

The documentation from 18 has this information on the main Attendances document, but in 19, it is in a different document.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5981002) for this PR.

NO POINTS for this, since it is just to ensure the various versions stay separate and keep version changes OK.


larm-odoo:
Hi @Felicious - this is another fix, but this is a separate zero points PR since the document changed a little from 17 to 18 and I did not want to overwrite/change in error the 18 version. There will be a PR for version 19 and up because this information has moved to a different document.

larm-odoo:
@robodoo r+

PR #16606, opened 2026-02-27
[FIX] Attendances - Update attendnace log location coordinate details

larm-odoo:

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5981039) for this PR.


larm-odoo:
Hi @Felicious - this is the last FIX PR.  This information is on the Attendance Logs document, not the main Attendances doc like the other PR's.

Felicious (approved):
@robodoo r+

PR #16689, opened 2026-03-04
[IMP] general integrations: 19.2 mail plugins update follow-up

huisit:
documentation task card:  https://www.odoo.com/odoo/my-tasks/6001333
17.0/18.0 PR: https://github.com/odoo/documentation/pull/16713
19.0/saas-19.1 PR: https://github.com/odoo/documentation/pull/16714

further updates to mail plugin docs for 19.2:

- mail plugins article:
    - fix grammar suggestions from feku's review
    - modify introduction to match lna's suggestions
- outlook plugin article:
    - add note that plugin is available for both web and desktop apps
    - add note that versions 19.2+ should refer to latest documentation for install
    - update introduction to match gmail article (multiple record types)
    - add subsectioning for install from marketplace vs. inbox, connect an odoo database
    - remove unused images
- gmail plugin article:
    - update installation writing to match outlook article
    - add subsectioning for install from marketplace vs. inbox, connect an odoo database
    - update connect an odoo database section to reflect current gmail ui
    - remove unused images

This saas-19.2 PR can be FWP up to master.


huisit:
@erjer-odoo this is ready for review! here's the [mail plugins article](https://runbot182.odoo.com/runbot/static/build/103060989-saas-19-2/logs/build/html/applications/general/integrations/mail_plugins.html) in the ci runbot. felicia is reviewing the related pr's for the older versions, so for this review on  your end it's just the 3 mail plugin articles in 19.2.

content/applications/general/integrations/mail_plugins.rst
erjer-odoo, line 17: Not that this is mandatory to change, but is there a reason these instructions aren't in the documentation for the individual plugins? The reason I ask is that at least in Gmail, the icons are different, or they use buttons instead of icons (in the case of Tasks, there's a **Create** button). In Outlook, when you're creating a contact, there's an **Add to Odoo** button instead of a plus icon. I also feel like if I'm looking for instructions for how to use this plugin, I'm looking on the individual plugin page, not the overview doc, if that makes sense.

If you want to keep them more general and leave it here, I don't think it's _mandatory_ to move it; I just feel like it makes more sense on the individual plugin documentation pages.

content/applications/general/integrations/mail_plugins/gmail.rst
```diff
@@ -9,146 +9,54 @@ records (such as opportunities, tasks, and tickets) directly in Gmail.
...
 Install the Gmail Plugin
-------------------------
+========================
+
+From the Google Workspace Marketplace
+-------------------------------------
+
+To install the Odoo Gmail plugin, sign in to the Gmail account to be connected to Odoo, then go to
+the :guilabel:`Odoo Inbox Addin` page in the `Google Workspace Marketplace
+<https://workspace.google.com/marketplace/app/odoo_inbox_addin/873497133275>`_. Click
```
erjer-odoo:
```suggestion
To install the Odoo Gmail plugin, sign in to the Gmail account to be connected to Odoo, then go to
the `Odoo Inbox Addin <https://workspace.google.com/marketplace/app/odoo_inbox_addin/873497133275>`_ page in the Google Workspace Marketplace. Click
```

Because the link is actually for the plugin, I would ink the Odoo Inbox Addin text, not the Google Workspace Marketplace. It's misleading text for screen readers, SEO, and LLMs. You want your link to go where you say it goes.

I understand why you used `:guilabel:`, but I think that if you include that in the link, it might not transform correctly. The link will transform as bold, anyway, so you get that visual emphasis that you're looking for, and it has the added bonus of being accurate for whatever it's linking to.

content/applications/general/integrations/mail_plugins/outlook.rst
```diff
@@ -2,29 +2,52 @@
...
 Install the Outlook plugin
 ==========================
 
+.. important::
+   Make sure to check the database version in the :guilabel:`Settings app --> General Settings`, at
+   the bottom of the page.
+
+   For database versions earlier than 19.2, see the `19.0 documentation
+   <https://www.odoo.com/documentation/19.0/applications/general/integrations/mail_plugins/outlook.html>`_
+   for installation instructions.
```
erjer-odoo, line 24:
```suggestion
.. important::
   Make sure to check the database version in the :guilabel:`Settings app --> General Settings`, at
   the bottom of the page. These instructions are valid for databases that are version 19.2 or higher.
```

I get why this is here, but I worry about how this needs to be handled after 19.2. To explain: In the past, when I have had to reference a specific version from which a feature is valid, I had to remove references to it in future branches. (For example, "As of version 18.3, it is possible to do x." And then in 18.4, the feature is just _there and how the product works_, so we just say "It is possible to do X.")

I'm not quite sure how to handle this, because the database needs to be version 19.2 or greater, but this documentation is specific to 19.2. I don't feel like we should be linking off to 19.0 documentation...19.0 users should _already be_ in the 19.0 documentation. This documentation should be _only for people already running a 19.2 database_.

So I've made a suggestion, and then in the `master` forward port, maybe this note could be something like "Be sure that you are running the latest database version in the Settings app --> General Settings"? It's not insanely graceful, but...just a thought.

Could be something to bring up to Felicia.

erjer-odoo (approved):
Hi @huisit -- I only really have a few comments, but they aren't necessarily mandatory. One or two might require a bit more work, but I think in general it's ready to be published, as long as Felicia also approves.

content/applications/general/integrations/mail_plugins/outlook.rst
huisit: Discussed offline to provide full context, neutral but will defer to PO's request

content/applications/general/integrations/mail_plugins.rst
huisit: Discussed offline to share context; information is roughly similar for both plugins, holding off on detailed feature documentation until old Gmail plugin is also replaced with 19.2+ version

huisit:
@erjer-odoo thanks for the review!
@Felicious this is ready for final!

content/applications/general/integrations/mail_plugins/outlook.rst
```diff
@@ -2,29 +2,52 @@
...
+
```
Felicious, line 25: extra line?

```suggestion
```

content/applications/general/integrations/mail_plugins/outlook.rst
```diff
@@ -2,29 +2,52 @@
...
+   `Microsoft's add-in article
+   <https://support.microsoft.com/en-us/office/use-add-ins-in-outlook-1ee261f9-49bf-4ba6-b3e2-2ba7bcab64c8>`_.
 
 .. _mail-plugin/outlook/install-plugin:
 
 Install the Outlook plugin
 ==========================
 
+.. important::
+   Make sure to check the database version in the :guilabel:`Settings app --> General Settings`, at
```
Felicious:
```suggestion
   Make sure to check the database version in the :menuselection:`Settings app --> General Settings`, at
```

Felicious (approved):
Looks good to me, @huisit !

Since this PR updates 3 docs, let's increase the scope of the PR to 5pts: 2+2+1 for each doc updated

@robodoo delegate+

huisit:
@robodoo r+

PR #16695, opened 2026-03-04
[IMP] pos: update Upsell note in Products 19.1

emmi-odoo:
task-5980038

### DIFF
--- a/content/applications/sales/point_of_sale/products.rst
+++ b/content/applications/sales/point_of_sale/products.rst
@@ -76,26 +76,25 @@ To create the combo choices that will be added to the :ref:`product combo
 #. Go to :menuselection:`Point of Sale --> Products --> Combo Choices` and click :guilabel:`New`.
 #. Enter a name for the :guilabel:`Combo Choice`.
 #. Set the maximum selectable items for the combo choice using the :guilabel:`Maximum items` field.
-#. Set the number of items included in the combo choice using the :guilabel:`Includes items` field.
-#. Optionally, enable the :guilabel:`Is Upsell` option if the combo choice is an upsell in the
-   :ref:`POS register <pos/use/open-register>`.
+#. Choose between the following options:
+
+   - Set the number of items included in the combo choice using the :guilabel:`Includes items`
+     field.
+   - Enable the :guilabel:`Is Upsell` option if the combo choice is optional and offered as an
+     additional paid upgrade.
 #. Click :guilabel:`Add a line` under the :guilabel:`Options` section to add the products that
    constitute the :guilabel:`Combo Choices`.
 #. If needed, click a product to add an :guilabel:`Extra Price`.
 
 .. admonition:: Combo Price vs. Extra Price
 
-  - The :guilabel:`Combo Price` field shows the price applied to any additional product a customer
-    might select (i.e., when the :guilabel:`Maximum items` field is set to `2` or higher). This
-    price is automatically calculated based on the price of the least expensive product defined in
-    the :guilabel:`Combo Choice`.
+  - The :guilabel:`Combo Price` field defines the price charged for any additional product selected
+    by the customer. It is applied when the :guilabel:`Maximum items` field is set to `2` or more,
+    or when a combo choice is configured as an optional upsell. The price is automatically
+    calculated based on the least expensive product in in the combo choice.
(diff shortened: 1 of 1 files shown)


content/applications/sales/point_of_sale/products.rst
```diff
@@ -94,8 +94,9 @@ To create the combo choices that will be added to the :ref:`product combo
     applied each time a customer selects that product within the combo choice.
 
 .. note::
-   Enabling the :guilabel:`Is Upsell` option removes the :guilabel:`Includes items` field, which
-   prevents from including items for free.
+   Enabling the :guilabel:`Is Upsell` option removes the :guilabel:`Includes items` field, and
+   consists of making a product optional inside the combo. However, selecting the optional product
+   automatically adds the amount of the :guilabel:`Combo Price` to the total price.
```
auva-odoo: 1) I would rephrase the instructions on line 80-81 to clarify the purpose of the field a bit and then make it clear, directly in the instructions that users can either set Includes Items or Is Upsell but not both. Here's a suggestion for that:

(suggestion for **lines 80-81**):
```suggestion
  #. Either:
  
      - Set the number of items included in the combo choice using the :guilabel:`Includes items` field, or; 
       - Enable the :guilabel:`Is Upsell` option if the combo choice is optional and offered as an additional paid upgrade.
```
(or something similar)

Then we don't need to talk about the Is Upsell field removing the Includes items field, do we?

2) Then, I would move the pricing info to the admonition on line 88 where we talk about the combo price. Maybe you could just add on line 91 something like "and is automatically applied when an item in a combo choice configured as an upsell is selected"? 
3) This would mean we wouldn't need any note here anymore.
4) I would also update the note on line 132, maybe, since we talk about price calculation there and this impacts it as well?

content/applications/sales/point_of_sale/products.rst
```diff
@@ -76,27 +76,26 @@ To create the combo choices that will be added to the :ref:`product combo
...
 .. admonition:: Combo Price vs. Extra Price
 
-  - The :guilabel:`Combo Price` field shows the price applied to any additional product a customer
-    might select (i.e., when the :guilabel:`Maximum items` field is set to `2` or higher). This
-    price is automatically calculated based on the price of the least expensive product defined in
-    the :guilabel:`Combo Choice`.
+  - The :guilabel:`Combo Price` field defines the price applied to any additional product selected
+    by the customer. This applies when the :guilabel:`Maximum items` field is set to `2` or more, or
+    when a combo choice is configured as an optional upsell. The price is automatically calculated
+    based on the least expensive product in in the combo choice.
```
auva-odoo:
```suggestion
  - The :guilabel:`Combo Price` field defines the price charged for any additional product selected
    by the customer. It is applied when the :guilabel:`Maximum items` field is set to `2` or more, or
    when a combo choice is configured as an optional upsell. The price is automatically calculated
    based on the least expensive product in in the combo choice.
```
to avoid using *apply* twice; I used the passive voice because I think it clarifies the sentence a bit (the price is added/applied)

content/applications/sales/point_of_sale/products.rst
```diff
@@ -130,8 +129,9 @@ To apply combos, follow these steps:
 .. note::
    The total price of the product combo, as displayed in the :ref:`POS register
    <pos/use/open-register>`, is based on the :guilabel:`Sales Price` defined on the :ref:`product
-   combo's form <pos/products/combo-creation>`. Selecting several products in the combo choices and
-   selecting a product with an :guilabel:`Extra Price` influences the total price.
+   combo's form <pos/products/combo-creation>`. Selecting several products in the combo choices,
+   selecting a product with an :guilabel:`Extra Price`, and selecting optional products configured
+   as upsells all influence the total price.
```
auva-odoo:
```suggestion
   combo's form <pos/products/combo-creation>`. Selecting several products in the combo choices,
   a product with an :guilabel:`Extra Price`, and/or products in a combo choice configured
   as an upsell all affect the total price.
```
to avoid repeating *selecting* + be more accurate (the combo choice is configured as an upsell, not the individual products inside it)

auva-odoo (changes requested):
Thank you for the changes @emmi-odoo, just a few more clarifications 😉

PR #16698, opened 2026-03-04
[IMP] Accounting: update send and print

jacbodoo:
task - 5223205


auva-odoo:
@jacbodoo did you also check /applications/finance/accounting/payments/follow_up ([this section more specifically](https://www.odoo.com/documentation/18.0/applications/finance/accounting/payments/follow_up.html#reports)) like I mentionned in my first review comment?

PR #16791, opened 2026-03-10
[IMP] db mobile: remove outdated information about iOS app discontinuation

xpl-odoo:
task-6022268


xpl-odoo:
Just a quick fix @auva-odoo ! FP confirmed on LinkedIn that Odoo doesn't intend to discontinue the mobile app.

PR #16812, opened 2026-03-12
[FIX] db: restore xmlrpc in web service example

xpl-odoo:
This commit partially reverts https://github.com/odoo/documentation/commit/36fb624162304ebd27826a03e48ca6b0ec84ecaa,
indeed the JSON-2 API only exists starting 19.0

### DIFF
--- a/content/administration/odoo_online.rst
+++ b/content/administration/odoo_online.rst
@@ -161,20 +161,51 @@ Web services
 
 To retrieve a list of all databases displayed under the `database manager
 <https://www.odoo.com/my/databases>`_ programmatically, call the `list` method of the
-`odoo.database` model via an :doc:`external JSON-2 API <../developer/reference/external_api>` call.
+`odoo.database` model via an :doc:`web service <../developer/howtos/web_services>` call.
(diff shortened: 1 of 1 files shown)


xpl-odoo:
@robodoo up to 18.0

Julien00859:
Commit title and message should be:

```
[FIX] <title>: restore xmlrpc in web service example

This commit partially reverts <hash of broken commit>, indeed the JSON-2 API only exists starting 19.0
```
I don't know for the title

Julien00859:
I think JSON-2 landed in 18.4... but in all cases xmlrpc was deprecated only in 19, and I don't know if internal enabled JSON-2 in 18.4. So up to 18.4 it is :+1:

@robodoo up to 18.4

Julien00859:
@robodoo up to saas-18.4

AntoineVDV:
@Julien00859 how do you dare up to 18.4??

@robodoo up to saas-18.4

AntoineVDV:
@robodoo talk to me like that @Julien00859

PR #16841, opened 2026-03-13
[IMP] on_premise: recommend users to fetch a single branch

vchu-odoo:
Cloning the whole repo is time-consuming and sometimes not needed, encourage people to start only with the single version.


AntoineVDV:
@robodoo fw=skipmerge

PR #16908, opened 2026-03-16
[IMP] Inventory: Cycle counts feedback

erjer-odoo:
While reviewing published content for our inventory adjustments and cycle counts tutorial, I ran across a [task card to update our cycle counts documentation](https://www.odoo.com/odoo/my-tasks/4435942).

After discussion with a higher-up and an SME, I've made changes and updated screenshots. Hopefully, this makes the content a bit clearer and more explicitly explains what must be done to ensure cycle counts are set up properly.

This 17.0 PR **should not be FWP**. There are further changes to both 18.0 and 19.0 that will be made in future PRs.


erjer-odoo:
@jero-odoo This PR is ready for your review when you have a moment. Thank you!

PR #17002, opened 2026-03-19
[IMP] Email: inbound redirection

jero-odoo:
Docs task: https://www.odoo.com/odoo/project.task/6046741

Added admonition block regarding local part detection
Fixed some line length issues


jero-odoo:
@stepperpig this is ready for review. Thanks!

jero-odoo:
Thank you @stepperpig ! A few of those are from a chart, so I wasn't able to change them too much, but I was able to update a few of them. Nice catch!

@Felicious this is ready for review!

jero-odoo:
@robodoo r+

PR #17012, opened 2026-03-20
[IMP] Users: update password management

jacbodoo:
task-6041661


content/applications/general/users.rst
```diff
@@ -164,9 +164,10 @@ Password management is an important part of granting users autonomous access to
 times. Odoo offers a few different methods to reset a user's password.
 
 .. tip::
-   Odoo has a setting to specify the length needed for a password. This setting can be accessed by
-   navigating to :menuselection:`Settings app --> Permissions` section, and entering the desired
-   password length in the :guilabel:`Minimum Password Length` field. By default the value is `8`.
+   Odoo has a setting to specify the length needed for a password. To access it, make sure the
+   `auth_password_policy` module is installed. Then, navigate to
+   :menuselection:`Settings app --> Permissions` section, and enter the desired password length
+   in the :guilabel:`Minimum Password Length` field. By default the value is `8`.
```
xpl-odoo:
```suggestion
   To enforce a minimum password length requirement, :ref:`install <general/install>` the :guilabel:`Password Policy`
   (`auth_password_policy`) module. Then, open the
   :guilabel:`Settings` app, navigate to the :guilabel:`Permissions` section, and enter the desired password length
   in the :guilabel:`Minimum Password Length` field. By default, the value is `8`.
```
I know you just added the info as requested, but I would improve the whole 
- "_Odoo has a setting to specify the length needed for a password._" is a bit vague, I would be more direct/precise and say "To enforce a minimum password length requirement, ..."
- I would add a ref to the install documentation and include the regular module name first, then its technical name in brackets.
- "Settings app" -> app shouldn't be part of the menuselection, since it's not displayed on the UI. Also, each step of a menuselection should represent a click, and you don't click to go to the Permission section. See my suggestion.

content/applications/general/users.rst
```diff
@@ -164,9 +164,10 @@ Password management is an important part of granting users autonomous access to
 times. Odoo offers a few different methods to reset a user's password.
 
 .. tip::
-   Odoo has a setting to specify the length needed for a password. This setting can be accessed by
-   navigating to :menuselection:`Settings app --> Permissions` section, and entering the desired
-   password length in the :guilabel:`Minimum Password Length` field. By default the value is `8`.
+   To enforce a minimum password length requirement, :ref:`install <general/install>` the :guilabel:`Password Policy`
+   (`auth_password_policy`) module. Then, open the
+   :guilabel:`Settings` app, navigate to the :guilabel:`Permissions` section, and enter the desired password length
+   in the :guilabel:`Minimum Password Length` field. By default, the value is `8`.
```
xpl-odoo:
```suggestion
   To enforce a minimum password length requirement, :ref:`install <general/install>` the
   :guilabel:`Password Policy` (`auth_password_policy`) module. Then, open the :guilabel:`Settings`
   app, navigate to the :guilabel:`Permissions` section, and enter the desired password length in
   the :guilabel:`Minimum Password Length` field. By default, the value is `8`.
```
It seems you copy/pasted my suggestion straight into the .rst file. However, respecting the [100-character limit](https://www.odoo.com/documentation/19.0/contributing/documentation/rst_guidelines.html#th-character-limit) when suggesting isn't really convenient (plus, if you don't have the same number of lines, you don't see the differences, which I find quite useful, so sometimes I will just add arbitrary line breaks so the differences appear)
Long story short, you should still pay attention to that when actually making the changes (this time you can actually use my suggestion as it is ;))

content/applications/general/users.rst
```diff
@@ -164,9 +164,10 @@ Password management is an important part of granting users autonomous access to
...
-   Odoo has a setting to specify the length needed for a password. This setting can be accessed by
-   navigating to :menuselection:`Settings app --> Permissions` section, and entering the desired
-   password length in the :guilabel:`Minimum Password Length` field. By default the value is `8`.
+   To enforce a minimum password length requirement, :ref:`install <general/install>` the
+   :guilabel:`Password Policy` (`auth_password_policy`) module. Then, open the
+   :guilabel:`Settings` app, navigate to the :guilabel:`Permissions` section, and enter the desired
+   password length in the :guilabel:`Minimum Password Length` field. By default, the value is `8`.
 
 .. image:: users/minimum-password-length.png
    :alt: Minimum Password Length highlighted in the Permissions section of General Settings.
```
xpl-odoo:
```suggestion
```
Again, I know you didn't update that part, but that kind of screenshot isn't very useful. Also, it should have been inside the tip, and the field used sentence capitalization back when it was taken... so I would remove it (the rst and the file) :)

PR #17125, opened 2026-03-26
[FIX] payment_methods/terminals/mercado_pago: Correct supported countries

masi-odoo:
The documentation for the Mercado Pago terminal is correct from versions `19.0` to `master`.

Still documentation for version `17.0` and `18.0` incorrectly state that the integration works also for Colombia, Peru, Chile and Uruguay, which is misleading clients, generating tickets and causing general confusion.

With this PR, the objective is to change documentation in these versions to match the correct supported countries (only Mexico, Argentina and Brazil).


auva-odoo:
@robodoo up to saas-18.4

PR #17190, opened 2026-03-31
[IMP]  Add tabs nesting check for RST

nikibrown:
Task: https://www.odoo.com/odoo/project/3835/tasks/6083413

**Issue:**

Forgetting a parent `.. tab::` in a group of tabs causes make fast issues
image

**Solution:**
Add checker to detect incorrect nesting in tabs

**Testing:**
Edit an RST file with tabs example: content/applications/marketing/events/event_reporting/attendees_report.rst and delete `.. tabs::` on line 22

Run `make fast` and reproduce errors

Run `make review` on content/applications/marketing/events/event_reporting/attendees_report.rst 

Expected output should be:
```
Enter relative content path: content/applications/marketing/events/event_reporting/attendees_report.rst 
Enter max line length (default: 100): 
content/applications/marketing/events/event_reporting/attendees_report.rst:24: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
content/applications/marketing/events/event_reporting/attendees_report.rst:42: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
content/applications/marketing/events/event_reporting/attendees_report.rst:58: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
content/applications/marketing/events/event_reporting/attendees_report.rst:78: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
content/applications/marketing/events/event_reporting/attendees_report.rst:94: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
make: *** [review] Error 1
```

Re-add parent `.. tab::` and run make review. Expected output should be: `No problems found.`

### DIFF
--- a/tests/checkers/rst_style.py
+++ b/tests/checkers/rst_style.py
@@ -18,6 +18,37 @@ FORBIDDEN_HEADING_DELIMITER_RE = re.compile(
 GIT_CONFLICT_MARKERS = ['<' * 7, '>' * 7]
 ALLOWED_EARLY_BREAK_RE = re.compile(r'^\s*(\.\. |:\S+:\s+)', re.IGNORECASE)  # Contains markup.
 TOCTREE_DIRECTIVE_RE = re.compile(r'^\s*\.\.\s+toctree::\s*$', re.IGNORECASE)
+# sphinx-tabs: tab-like directives must be nested under .. tabs:: (see tests/main.py CUSTOM_RST_DIRECTIVES).
+TABS_DIRECTIVE_RE = re.compile(r'^(\s*)\.\.\s+tabs::\s*')
+TAB_CHILD_DIRECTIVE_RE = re.compile(
+    r'^(\s*)\.\.\s+(tab|group-tab|code-tab)::\s*'
+)
+
+
+@sphinxlint.checker('.rst')
+def check_tabs_directive_nesting(file, lines, options=None):
+    """ Check that .. tab:: / .. group-tab:: / .. code-tab:: are nested under .. tabs::. """
+    tabs_indent_stack = []
+
+    for lno, line in enumerate(lines):
+        stripped = line.strip()
+        if not stripped or stripped.startswith('#'):
+            continue
+
+        indent_match = re.match(r'^(\s*)', line)
+        indent = len(indent_match.group(1)) if indent_match else 0
+
+        while tabs_indent_stack and indent <= tabs_indent_stack[-1]:
+            tabs_indent_stack.pop()
+
+        if TABS_DIRECTIVE_RE.match(line):
+            tabs_indent_stack.append(indent)
+        elif TAB_CHILD_DIRECTIVE_RE.match(line):
+            if not tabs_indent_stack or indent <= tabs_indent_stack[-1]:
+                yield lno + 1, (
+                    "tab directive must be nested under .. tabs:: (indent it under a parent "
+                    ".. tabs:: block)"
+                )
 
 
 @sphinxlint.checker('.rst')


theRealThagomizer (approved):
Hiya, @nikibrown! I got the same failstates as yesterday with a handy error message in make review. Here's what I did.

1. Opened content/applications/sales/crm/performance/expected_revenue_report.rst and deleted .. tabs:: at line 138 (I couldn't find the file at the path you suggested so I went digging for another one with tabs in it.)
2. Ran make clean/make fast. Make fast failed. Error follows:
`waiting for workers...

Traceback (most recent call last):
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/cmd/build.py", line 280, in build_main
app.build(args.force_all, filenames)
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/application.py", line 344, in build
self.builder.build_update()
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/builders/init.py", line 294, in build_update
self.build(to_build,
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/builders/init.py", line 308, in build
updated_docnames = set(self.read())
^^^^^^^^^^^
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/builders/init.py", line 413, in read
self._read_parallel(docnames, nproc=self.app.parallel)
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/builders/init.py", line 464, in _read_parallel
tasks.join()
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/util/parallel.py", line 108, in join
if not self._join_one():
^^^^^^^^^^^^^^^^
File "/home/odoo/github/odoo/.venv/lib/python3.12/site-packages/sphinx/util/parallel.py", line 129, in _join_one
raise SphinxParallelError(*result)
sphinx.errors.SphinxParallelError: KeyError: 'tabs_stack'

Sphinx parallel build error:
KeyError: 'tabs_stack'
make: *** [Makefile:61: html] Error 2`

3. Ran make review on same file. Errors follow:
`content/applications/sales/crm/performance/expected_revenue_report.rst:57: consider moving "Type" to line 57 (early-line-breaks) content/applications/sales/crm/performance/expected_revenue_report.rst:65: consider moving "filter." to line 65 (early-line-breaks) content/applications/sales/crm/performance/expected_revenue_report.rst:207: consider moving ":guilabel:➖" to line 207 (early-line-breaks)
content/applications/sales/crm/performance/expected_revenue_report.rst:139: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
content/applications/sales/crm/performance/expected_revenue_report.rst:163: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)
content/applications/sales/crm/performance/expected_revenue_report.rst:188: tab directive must be nested under .. tabs:: (indent it under a parent .. tabs:: block) (tabs-directive-nesting)`

Seems like it's functioning as you intended? If so, you've got my approval!

Felicious:
Hi @nikibrown!  I reviewed your PR and wanted to understand the reasoning behind the approach.

From what I see, the issue is that when a parent `.. tabs::` directive is missing, `make fast` throws an error that isn’t very descriptive and points toward Sphinx lint, which can make it seem like there’s an issue with the make process itself rather than with the tab nesting.

Given that, I’m trying to reconcile the chosen solution. Instead of improving the `make fast` feedback (e.g., making the error message explicitly indicate incorrect tab nesting or adding a failsafe check there), the fix introduces a checker in `make review`, which is a separate command.

Is there a specific reason this validation was added to `make review` rather than addressing the ambiguity in `make fast` directly? Am I missing a constraint or design consideration that makes this approach preferable?

nikibrown:
@Felicious Good question. 

The error message from running `make fast`:

```
File "/venv/main/lib/python3.12/site-packages/sphinx_tabs/tabs.py", line 143, in run
    tabs_id = self.env.temp_data["tabs_stack"][-1]
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
IndexError: list index out of range
 ```

comes from sphinx_tabs which is a package that sphinx uses to build out the tabs. Not sure if we can override that.  We can edit the checks run by make review which is why I went that route. 

@AntoineVDV Is there a better way to check for `.rst` syntax errors when running `make fast`?

AntoineVDV (commented):
Hi, I kinda agree with @Felicious that, although a new `make review` check might help some users, it doesn't address the root cause of the problem.

The root cause is clearly that the `sphinx-tab` extension should more carefully check whether the `.. tabs` directive has been provided, and log a warning if not.

I see three root cause-oriented approaches we could take:
1. Live with it. If only one or two writers reported the issue, we should probably handle it the Odoo way and not overreact: the issue is rare enough that it doesn't justify investing resources in solving it, and these writers would already have learned from their mistake.
2. Check if the latest updates of `sphinx-tabs` fixed the issue. It might be tricky to test in our documentation project because the extension depends on `sphinx`, which would need to be upgraded as well. But it's a good exercise to understand why we want to keep customizations in the documentation project to a minimum 😁 
3. Import the sources of the `sphinx-tabs` extension in our documentation sources and fix the bug ourselves.

nikibrown:
@AntoineVDV Thanks for the feedback. Agree that this is probably a rare issue. I think the effort to fix outweighs the benefits.

PR #17277, opened 2026-04-07
[ADD] general: offline mode

mcsi-odoo:
task-6080512

### DIFF
--- a/content/applications/general/offline_mode.rst
+++ b/content/applications/general/offline_mode.rst
@@ -1,0 +1,26 @@
+============
+Offline mode
+============
+
+To ensure business continuity when connectivity is unreliable or unavailable, such as when commuting
+or while working in locations with poor or no coverage, Odoo makes certain data available offline
+for logged-in users.
+
+When the internet connection is no longer active, views or records that were previously opened while
+online can be re-opened, regardless of whether they were active in the browser at the moment
+connectivity was lost.
+
+Previously opened records can be viewed but not modified while offline.
+
+.. tip::
+   If you foresee connectivity issues, briefly open any views or records you expect to need so they
+   can still be accessed if/when connectivity is lost.
+
+.. note::
+   - Odoo supports offline access for apps and modules built using :doc:`standard Odoo
+     framework components </developer/reference/frontend/framework_overview>`.
+   - For performance reasons, Odoo allocates a maximum of 2 GB for offline data. When this limit is
+     reached, the oldest data is removed to make room for the most recent activity.
+   - Performing a hard refresh using `Ctrl + F5` or `Cmd + Shift + R`, e.g., for troubleshooting
+     reasons, removes all locally stored data. This impacts the availability of views and records in
+     offline mode.
--- a/content/applications/general.rst
+++ b/content/applications/general.rst
@@ -11,3 +11,4 @@ General settings
     general/email_communication
     general/integrations
     general/developer_mode
+    general/offline_mode


content/applications/general/offline_mode.rst
```diff
@@ -0,0 +1,23 @@
+============
+Offline mode
+============
+
+To ensure business continuity when connectivity is unreliable or unavailable, such as when commuting
+or while working in locations with poor or no coverage, Odoo makes certain data available offline.
+
+When the internet connection is no longer active, any views or records that were previously
+opened while online remain accessible; previously opened records can be viewed but not modified
```
auva-odoo: Is there a specific timeframe for this? How does this work exactly? 
It might be me being a non-native but when you say "previously" opened, I understand that if I opened 100 records, all 100 will be accessible offline, even if I closed some of them prior to being disconnected (because maybe they're cached or whatever other technical thing? 🙈) Is that correct?
Or do you mean that the records that were still opened (in a tab for example) prior to being disconnected remain accessible?

EDIT: Reading the rest of the doc, I'm leaning towards option 1 but I'm still not 100% sure. And also, will this still be the case if I close Odoo entirely on the the mobile device before being "offline"? or if I shut it off? or the next day/anytime actually, until the 2GB are reached?

content/applications/general/offline_mode.rst
```diff
@@ -0,0 +1,23 @@
...
+while offline.
+
+.. tip::
+   If you foresee connectivity issues, briefly open any views or records you expect to need so they
+   can still be accessed if/when connectivity is lost.
+
+.. note::
+   - Odoo supports offline access for apps and modules using standard Odoo framework components.
```
auva-odoo: Is there maybe a link we could add for the *standard Odoo framework components*?

content/applications/general/offline_mode.rst
```diff
@@ -0,0 +1,23 @@
...
+     Offline access cannot be guaranteed for apps or modules using custom components.
```
auva-odoo: What is included in those "custom components"? Do we mean actual customizations (i.e. studio/custom web design stuff)? or are there other things involved?

auva-odoo (commented):
Hi @mcsi-odoo, sorry for my probably stupid question on this PR. It might be simpler to discuss this IRL, let me know what you think!

mcsi-odoo:
Hey @auva-odoo , some small updates made, as discussed. See what you think.

content/applications/general/offline_mode.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Previously opened records can be viewed but not modified while offline.
+
+.. tip::
+   If you foresee connectivity issues, briefly open any views or records you expect to need so they
+   can still be accessed if/when connectivity is lost.
+
+.. note::
+   - Odoo supports offline access for apps and modules built using :doc:`standard Odoo
+     framework components <../../developer/reference/frontend/framework_overview>`.
```
auva-odoo: just an FYI here but you can definitely use an absolute path here (same as when you have to go all the way back up to /applications)

```suggestion
     framework components </developer/reference/frontend/framework_overview>`.
```

auva-odoo (commented):
Thank you for taking my feedback into consideration @mcsi-odoo. This LGTM! 🙏 

@robodoo delegate+

mcsi-odoo:
@robodoo r+

PR #17292, opened 2026-04-07
[IMP] Manufacturing: Shop Floor time tracking improvements

erjer-odoo:
[Task card](https://www.odoo.com/odoo/my-tasks/4873467).

In addition to the requested update, I noticed that the 19.0 version of this article had already been updated in a way that would also greatly improve the 18.0 version. I brought many of those changes in here, while keeping it in line with how 18.0 operates.

This 18.0 PR **should not be FWP**. There may be a couple of changes in 19.0+, but those are minor and can be handled in a separate PR.


erjer-odoo:
Hi @theRealThagomizer -- I forgot to tag you in this yesterday. This is ready for review when you get a moment. Thank you!

content/applications/inventory_and_mrp/manufacturing/shop_floor/shop_floor_tracking.rst
```diff
@@ -5,115 +5,150 @@ Shop Floor time tracking
...
-To sign out a specific employee from the module, click the :guilabel:`X (remove)` button next to
-their name, in the operator panel.
+To sign out a specific employee from the module, click the :icon:`oi-close` :guilabel:`(logout)`
+icon next to their name in the operator panel.
 
 Track work order duration
 =========================
 
 To track time spent working on a work order, begin by selecting the employee working on it from the
 operator panel.
```
theRealThagomizer:
```suggestion
To track time spent on a work order, begin by selecting the employee assigned to it from the
operator panel.
```
The amount of work this sentence was doing jumped out to me. Just a suggestion.

theRealThagomizer (approved):
Great update, @erjer-odoo! I just had one little suggestion where a sentence sounded funny to my ear. I also encountered an issue on the runbot instance where MO was being recognized as an abbreviation, but wasn't actually popping up the full term. Maybe a runbot quirk, maybe something in the way the abbreviation replacement wasn't set up? In any case, once that's been double-checked, this is approved. Thank you!

erjer-odoo:
Thanks @theRealThagomizer ! I didn't see/couldn't reproduce the issue with the abbreviation. Everything looked fine to me, but it's also possible that I'm too close to the code and blind to it.

@Felicious This should be ready for final review. If you see the same issue Thomas pointed out, please let me know, and I'll investigate.

Felicious:
@robodoo fw=no

Felicious:
I didn't see an issue with the abbreviation! Merging now (: 

@robodoo r+

PR #17305, opened 2026-04-08
[FIX] Sales: Removed version specific reference

theRealThagomizer:
This is a quick fix to the Sales documentation. It removes a version-specific callout in a note admonition, which was against our style guidelines. It also removes the file for an image that no longer appears in the documentation.


theRealThagomizer:
Hiya, @Felicious! Got a super minor fix for you to review. This removes the version-specific admonition block you mentioned yesterday and also an outdated image file. There are no other changes. Thank you!

Felicious:
Hi Thomas! It looks like this PR only targets 19.0 and up. I'd like that call out removed for 17 and 18 as well (:

theRealThagomizer:
@Felicious Gotcha. I'm going to just close this PR out then and open a new 17.0 one, because I don't trust myself to rename everything correctly.

PR #17306, opened 2026-04-09
[FIX] Sales: Removed style-inappropriate language

theRealThagomizer:
This is a quick fix to the Sales documentation. It removes a version-specific callout in a note admonition, which was against our style guidelines. It also removes the file for an image that no longer appears in the documentation.

This PR addresses the same fix as a previous closed PR (https://github.com/odoo/documentation/pull/17305).


theRealThagomizer:
@Felicious Take two!

Felicious:
@robodoo r+

PR #17382, opened 2026-04-15
[IMP] cli: document the log coloring using env variables

KangOl:


content/developer/reference/cli.rst
```diff
@@ -587,6 +587,14 @@ logging to other destinations and to customize the verbosity.
         In case of conflict between :option:`--log-level` and
         :option:`--log-handler`, the latter is used
 
+.. versionadded:: 20.0
```
AntoineVDV: The documentation itself is already versioned; no need to specify this. In particular, it won't be strictly true in 19.3 and 19.4.

AntoineVDV: It would however fit just nicely as a second bullet point of the note above.

content/developer/reference/cli.rst
```diff
@@ -587,6 +587,14 @@ logging to other destinations and to customize the verbosity.
...
+
+    When logs are output on the terminal (``tty``), they are colored to ease reading. You can control
+    the coloring using the `NO_COLOR <https://no-color.org>`_ and `FORCE_COLOR <https://force-color.org>`_
+    environment variables (the former take precedence over the latter).
```
AntoineVDV:
```suggestion
    environment variables (the former takes precedence over the latter).
```

AntoineVDV (changes requested):
@robodoo delegate+

PR #17436, opened 2026-04-17
[IMP] sign: 19.2 updates

auva-odoo:
task-5882197


content/applications/productivity/sign/configuration.rst
```diff
@@ -82,29 +82,29 @@ To create or edit a field:
...
    - :guilabel:`Placeholder`: Optionally, specify a placeholder text that provides the signer with a
      hint or example :ref:`during the signing process <sign/sign-document/signature-request>`. If
      this field is left blank, the :guilabel:`Field Name` is used as placeholder text.
    - :guilabel:`Tip`: Add a tip that will be displayed inside an arrow to the left
      of the document during the signing process to help the signer understand what action to take
      (e.g., "Sign here").
-   - For :guilabel:`Text` fields only, optionally change the default :guilabel:`Field Size` from
-     :guilabel:`Regular Text` to :guilabel:`Short Text` or :guilabel:`Long Text`.
+   - :guilabel:`Mandatory`: Enable the check box to make the field mandatory by default.
+   - :guilabel:`Read Only`: Enable the check box to make the field read-only by default.
```
auva-odoo: I know this check box is currently not available for all field types but after discussing this with the PO, it should be (the current logic is a bit weird, hence why I chose not to document it as is and stay a bit more vague)

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -143,50 +143,32 @@ Edit the template:
 - :ref:`Prepare the document(s) for signing <sign/request-signatures/prepare-document>` by adding
   or editing fields, and, if relevant, adding or removing signers and/or documents.
 - Optionally, :ref:`define or edit settings for specific signers
-  <sign/request-signatures/signer-settings>`, such as requiring additional authentication, allowing a
-  signer to be replaced, or assigning a fixed signer.
+  <sign/request-signatures/signer-settings>`, such as requiring additional authentication, allowing
+  a signer to :ref:`delegate their signature <sign/sign-document/delegate>` and to be :ref:`replaced
+  <sign/request-signatures/change-signers>`, or assigning a fixed signer.
```
mcsi-odoo, line 148:
```suggestion
  <sign/request-signatures/signer-settings>`, such as requiring additional authentication, allowing
  a signer to :ref:`delegate their signature <sign/sign-document/delegate>`, allowing a signer to be :ref:`replaced
  <sign/request-signatures/change-signers>`, or assigning a fixed signer.
```

If I understand correctly, allowing a signer to delegate their own signature and allowing anyone with the appropriate rights to replace a signer are two different things, right? And both are only possible if 'Can delegate' has been enabled.

content/applications/productivity/sign.rst
```diff
@@ -24,20 +24,20 @@ meet the requirements for electronic signatures in most countries.
 
 With Sign, you can:
 
-- :ref:`prepare documents for signing <sign/request-signatures/prepare-document>` by one or more signers using
-  drag-and-drop fields
+- :ref:`prepare documents for signing <sign/request-signatures/prepare-document>` by one or more
+   signers using drag-and-drop fields
```
mcsi-odoo:
```suggestion
- :ref:`prepare documents for signing <sign/request-signatures/prepare-document>` by one or more
  signers using drag-and-drop fields
```

There's an extra space before 'signers' that's messing with the formatting.

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -143,50 +143,32 @@ Edit the template:
...
-    template available from any model.
-  - :guilabel:`Redirect Link`: Add a url to which the user is redirected after signing the document,
-    e.g., to a *Thank you* page, or to your company's homepage.
+  - By default, the :guilabel:`Name` of the template is the name of the file that was uploaded,
+    or, in the case of a :ref:`document envelope <sign/request-signatures/create-envelope>`, the
+    name of the first file uploaded. If desired, edit the name of the template.
+  - :guilabel:`Available in`: Select the :doc:`Odoo model <../../studio/models_modules_apps>` to
+     which the template is linked to restrict the template to :ref:`signatures requests made from
+     records of that model <sign/request-signatures/template-odoo-record>`. Leave the field empty to
+     make the template available from any model.
```
mcsi-odoo:
```suggestion
  - :guilabel:`Available in`: Select the :doc:`Odoo model <../../studio/models_modules_apps>` to
    which the template is linked to restrict the template to :ref:`signatures requests made from
    records of that model <sign/request-signatures/template-odoo-record>`. Leave the field empty to
    make the template available from any model.
```

Just some extra spaces messing with the formatting.

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -143,50 +143,32 @@ Edit the template:
...
-      this template.
-    - :guilabel:`Responsible` user: By default, this is set to the user who created the template,
-      but a different user can be selected.
+.. tip::
+   To replace a document in the template, such as with an updated version, while preserving the
+   fields already added, click the :icon:`fa-ellipsis-v` (:guilabel:`vertical ellipsis`) icon next
+   to the relevant document, select :guilabel:`Replace document`, choose the new document, then
+   click :guilabel:`Replace document`. Each replacement creates a duplicate of the template with
+   the fields placed at the same position as in the original. If the new document has a different
+   layout or added text, the fields may no longer align correctly. Reposition them as needed.
```
mcsi-odoo:
```suggestion
   To replace a document in the template, such as with an updated version, while preserving the
   fields already added, click the :icon:`fa-ellipsis-v` (:guilabel:`vertical ellipsis`) icon next
   to the relevant document, select :guilabel:`Replace document`, choose the new document, then
   click :guilabel:`Replace document`. 
   
   Each replacement creates a duplicate of the template with
   the fields placed at the same position as in the original. If the new document has a different
   layout or added text, the fields may need to be repositioned.
```

1) I think this block is a bit too big to keep in one piece, so I'd suggested starting a new paragraph at 'Each replacement ...'
2) About repositioning: maybe worth (adding an anchor to and) linking to this content (in the 'Add fields to a document' part), and to think about moving the tip so it's closer to the point about repositioning? I'm thinking that if someone needs to be moving around fields on such a replacement document, they might appreciate the help

Image

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
 
+Export template answers
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Answers from all signature requests sent using a specific template can be exported, provided the
+:doc:`../documents` app is installed. To export them, go to :menuselection:`Sign --> Templates`,
+click the relevant template's card, then click the :icon:`fa-cog` (:guilabel:`Actions`) icon and
+select :guilabel:`Export Answers`.
+
+Answers are exported to a :doc:`../spreadsheet`, with one column per :ref:`field
```
mcsi-odoo:
```suggestion
Answers are exported to an :doc:`Odoo spreadsheet <../spreadsheet>`, with one column per :ref:`field
```

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
+- :ref:`Edit a template <sign/request-signatures/edit-template>` that has *not yet been used* to
+  send a signature request. Alternatively, click on a template to open it for editing.
+- :guilabel:`Duplicate` a template. The duplicated template can then be :ref:`edited
+  <sign/request-signatures/edit-template>` as needed.
 - :guilabel:`Archive` a template that is not currently needed.
+- Modify a template's :ref:`configuration settings <sign/request-signatures/template-form>`.
 - Select a :guilabel:`Color` to add a colored bar to the left side of a card to visually
   differentiate it.
 
+Export template answers
```
mcsi-odoo: General comment: I know they've used the term 'answers' in the app, but I really don't think that works well. Could we use 'response(s)' except when referring to actual GUI labels that use 'answers'?

For this title, I'm not sure we need to mention 'template', since we're in the 'Manage templates' section, and the first sentence explains/will explain it clearly enough.

So, my suggestions for a title are:

- Export responses
- Export response data
- Export signer responses

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Answers from all signature requests sent using a specific template can be exported, provided the
```
mcsi-odoo: I'm not sure that '... from all signature requests sent' really captures this accurately enough, since it's more like 'responses entered by signers in template-based documents or document envelopes sent via a signature request...'. 

Maybe it could be something as simple as:

- All signer responses related to a specific template can be exported, provided the Documents app is installed.

Lastly, for this one ... I'm wondering if it might be useful here to give an example of what this could be used for?

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
+click the relevant template's card, then click the :icon:`fa-cog` (:guilabel:`Actions`) icon and
+select :guilabel:`Export Answers`.
+
+Answers are exported to a :doc:`../spreadsheet`, with one column per :ref:`field
+<sign/request-signatures/add-fields>`, one line per :ref:`signer <sign/request-signatures/signers>`,
+and one tab per document for :ref:`document envelopes <sign/request-signatures/create-envelope>`.
+
+.. note::
+   The spreadsheet is named after the template and saved in the :guilabel:`Sign Request Answers`
+   folder in Documents.
```
mcsi-odoo: Maybe it's just me, but I feel 'is named after' is more for people, e.g., naming a child after a grandparent. But ... I'm struggling to find a concise way to word this. Something like this maybe?

```suggestion
.. note::
   The spreadsheet is saved in the :guilabel:`Sign Request Answers`
   folder in Documents, with the template name referenced in the name of the spreadsheet.
```

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
+
+.. _sign/request-signatures/template-form:
+
+Template configuration settings
```
mcsi-odoo: Just 'Template settings'?

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -288,17 +274,16 @@ a Kanban view, with cards shown horizontally.
...
 For each template, the following information is shown:
 
 - the name of the template
 - the month and year in which the template was created
-- the avatar of the :guilabel:`Responsible` user for the template; this is by default the user who
-  created the template, but this can be changed in the configuration settings of the template if
-  :guilabel:`Manage template access` is enabled in the :ref:`main Sign settings
-  <sign/configuration/settings>`
+- the avatar of the template's :guilabel:`Owner`; by default, the owner is the user who created the
+  template. This can be changed in the :ref:`configuration settings of the template
```
mcsi-odoo: Just 'template settings' rather than 'configuration settings of the template'?

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
-- :ref:`Edit a template <sign/request-signatures/edit-template>` that has *not yet been used* to send a signature
-  request. Alternatively, click on a template to open it for editing.
-- :guilabel:`Duplicate` a template. The duplicated template can then be
-  :ref:`edited <sign/request-signatures/edit-template>` as needed.
+- :ref:`Edit a template <sign/request-signatures/edit-template>` that has *not yet been used* to
+  send a signature request. Alternatively, click on a template to open it for editing.
+- :guilabel:`Duplicate` a template. The duplicated template can then be :ref:`edited
+  <sign/request-signatures/edit-template>` as needed.
 - :guilabel:`Archive` a template that is not currently needed.
+- Modify a template's :ref:`configuration settings <sign/request-signatures/template-form>`.
```
mcsi-odoo: As before, just 'settings' rather than 'configuration settings'?

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
 Hovering over the template's card reveals a :icon:`fa-ellipsis-v` :guilabel:`(vertical ellipsis)`
 icon in the top-right corner. Click the icon to:
 
-- :ref:`Edit a template <sign/request-signatures/edit-template>` that has *not yet been used* to send a signature
-  request. Alternatively, click on a template to open it for editing.
-- :guilabel:`Duplicate` a template. The duplicated template can then be
-  :ref:`edited <sign/request-signatures/edit-template>` as needed.
+- :ref:`Edit a template <sign/request-signatures/edit-template>` that has *not yet been used* to
+  send a signature request. Alternatively, click on a template to open it for editing.
```
mcsi-odoo, line 303: Maybe after this point we should add a point about the 'Stop Sharing' option, which I guess only appears if there is a template for which a sharing link has been created. 

Something like:

```suggestion
- :guilabel:`Stop Sharing` the link to the templated document or document envelope, if relevant; after sharing is stopped, the link can no longer be opened.
```

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
+- :guilabel:`Available in`: Select the :doc:`Odoo model <../../studio/models_modules_apps>` to
+  which the template is linked to restrict the template to :ref:`signature requests made from
+  records of that model <sign/request-signatures/template-odoo-record>`. Leave the field empty to
+  make the template available from any model.
+- :guilabel:`Authorized Users`: Specify the users authorized to view and use the template.
+- :guilabel:`Authorized Groups`: If :guilabel:`Manage template access` is enabled in the
+  :ref:`main Sign settings <sign/configuration/settings>`, define the user groups whose members are
+  authorized to view and use this template.
+- :guilabel:`Owner`: The user authorized to modify the template. By default, this is the user
+  who created it. Select a different user to transfer ownership.
```
mcsi-odoo: I don't think it's only the Owner who can modify the template. In my runbot, I set Marc Demo as a second Authorized User and it seems he can do anything (add and remove fields and signers, update the template settings) except link the template to a model. Maybe you had a different experience? Or maybe it _should_ only be the Owner who can modify the template?

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +297,77 @@ For each template, the following information is shown:
...
+- :guilabel:`Redirect Link`: Add a URL to redirect signers after signing the document, such as a
+  *Thank you* page, or the website's homepage.
+- In the :guilabel:`Communication` tab, type a default text to include when a signature request
+  using this template is sent via email. The text can still be edited before sending.
+- In the :guilabel:`Signers` tab, view and edit any :ref:`signer settings
+  <sign/request-signatures/signer-settings>` defined for the template.
+
+.. tip::
+   - Access all signature requests linked to the template by clicking the :guilabel:`Signed
+     Document` smart button.
```
mcsi-odoo: I guess that, like before, these aren't actually the _signature requests_ that appear, but rather the fully signed documents (e.g., if you share a link to a templated document with 50 people and 25 sign the document, those 25 'documents' will show when you click 'Signed Document'; or if you individually send signature requests using that template to 5 people, it's only the signed documents - and not the signature requests - that appear)

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -482,10 +530,11 @@ Signer settings
...
 ellipsis)` icon on the row of the signer. For each signer, the following options are available:
 
-- :guilabel:`Authentication`: Require the signer to :ref:`authenticate using a specific
+- :guilabel:`Extra Authentication Step`: Require the signer to :ref:`authenticate using a specific
   authentication method <sign/security/authentication>`.
-- :guilabel:`Can delegate`: Allow :ref:`the signer to be replaced
-  <sign/request-signatures/change-signers>` after the signature request has been sent.
+- :guilabel:`Can delegate`: Allow the signer to :ref:`delegate their signature
+  <sign/sign-document/delegate>` and :ref:`be replaced after the signature request has been sent
+  <sign/request-signatures/change-signers>`.
```
mcsi-odoo, line 544:
```suggestion
- :guilabel:`Can delegate`: Allow the signer to :ref:`delegate their signature
  <sign/sign-document/delegate>` or :ref:`be replaced after the signature request has been sent
  <sign/request-signatures/change-signers>`.
```

I think the 'and' should be 'or', no? Becuase in my mind we're talking about two different things here: me delegating my signature and thereby nominating someone else to replace me ... OR ... the sender of a signature request replacing a signer, e.g., because they were informed that someone left the company. To me, with the 'and' it reads like the two parts of the sentence are part of the same thing.

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -522,10 +571,11 @@ It is also possible to:
 - :ref:`set a Valid Until date <sign/request-signatures/validity>` after which the signature request
   expires
 - :ref:`set Reminders <sign/request-signatures/validity>` to be sent at fixed intervals
-- enable :guilabel:`Add certificate on each page` to add the reference of the :ref:`certificate of
-  completion <sign/security/certificate-of-completion>` to each page of the document(s)
-- have one or more parties notified when the document is fully signed by adding the relevant
-  contact(s) or email address(es) in the :guilabel:`CC` field
+- enable :guilabel:`Hide certificate key pages` to hide the :ref:`reference of the certificate of
+  completion <sign/security/hash>` displayed by default on each page of the document(s)
```
mcsi-odoo:
```suggestion
- enable :guilabel:`Hide certificate key on pages` to hide the :ref:`reference of the certificate of
```

Also, I think the ref is to slightly the wrong place. It should be to 'sign/security/certificate-of-completion'.

The signatory hash is related to each individual signer, if I understand right, while the certificate reference is really at the level of the fully signed document.  

In the screenshots, the red box is the (same) certificate reference that can be hidden on the signed doc and that appears on the certificate of completion, while the green boxes are the signatory hashes that only appear on the certificate of completion.

_Screenshot 1_
Image

_Screenshot 2_
Image

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -620,16 +670,16 @@ In the Kanban view:
 
   - When a signer has fully signed the document(s), the checkbox is ticked, the signer's name
     appears in green, and the date of the signature is added after the name.
-  - When a signature request is cancelled, the signers name appears in red with a :icon:`fa-times`
+  - When a signature request is canceled, the signers name appears in red with a :icon:`fa-times`
```
mcsi-odoo:
```suggestion
  - When a signature request is canceled, the signer's name appears in red with a :icon:`fa-times`
```

content/applications/productivity/sign/sign_document.rst
```diff
@@ -118,6 +118,29 @@ To define how your initials/signature should appear after clicking the relevant
 If there are multiple documents to sign within the same document envelope, the defined
 initials/signature can thereafter be added directly by clicking the relevant field.
 
+.. _sign/sign-document/delegate:
+
+Delegate signing
```
mcsi-odoo, line 123: My comments are similar for these 'Delegating' and 'Declining' sections, so I'll group them here.

1) The second sentence each time (i.e., 'The signature request is then ...') kind of mixes information that's relevant for the signer (e.g., the fact that the signature request is then sent to the new signer or that any other signers are notified of a refusal) with other info that's only really relevant for the sender.

   Could we somehow be more explicit about the distinction between signer actions/consequences here and the consequences for the sender/in Odoo Sign, if you see what I mean? And also maybe harmonise how we explain what is updated and where after the delegation/refusal? (e.g., about it being in the signature request details)

   Maybe we can discuss as I'm not convinced I'm explaining myself well 😅 

2) We talk about changes being recorded in the chatter for both, but I guess that's only in instances where a signature request was initiated from an Odoo record?

3) About the notes: I guess delegating also delegates the entire signature request (i.e., not possible to delegate for only one doc of an envelope), so both notes are relevant to both sections. Could we (each time) group the two notes in one and put it at the start of the sections?

mcsi-odoo: And I FINALLY discovered how to indent paragraphs in numbered lists here 🥳  ☝️

content/applications/productivity/sign/configuration.rst
mcsi-odoo: FWIW, I always write 'checkbox' as one word (and it's what my online dictionary gives, even for US English), and have the impression that's what I usually see.

content/applications/productivity/sign/configuration.rst
```diff
@@ -132,30 +132,28 @@ Field types
   which adds a visible line through the text.
 - :guilabel:`Stamp`: Signers can add a company stamp (defined when the
   field is :ref:`added to the document <sign/request-signatures/add-fields>`).
+- :guilabel:`Date`: When signers click a date field, today's date is filled in by default. They can
+  click it again to select a different date in the date picker. This field type can also be
+  :ref:`configured to be auto-completed <sign/configuration/auto-fill>` during the signing process.
```
mcsi-odoo: 'They can ...' just sounds a little odd to me. I'd suggest:

```suggestion
- :guilabel:`Date`: When signers click a date field, today's date is filled in by default. Clicking the field again allows a different date to be selected in the date picker. This field type can also be
  :ref:`configured to be auto-completed <sign/configuration/auto-fill>` during the signing process.
```

content/applications/productivity/sign/configuration.rst
```diff
@@ -132,30 +132,28 @@ Field types
...
 
 To enable auto-completion for a Sign field:
 
 #. Go to :menuselection:`Sign --> Configuration --> Fields`.
 #. Open the relevant field.
-#. From the dropdown beside :guilabel:`Linked to` select the model, e.g., *Contact*, that contains
-   the field you want to link to.
+#. From the :guilabel:`Linked to` dropdown, select the model that contains the field to link to,
+   e.g., *Contact*.
 #. Click beside :guilabel:`Linked Field`, then use the field selector to select the relevant field,
```
mcsi-odoo: On your new screenshot, I just see 'Field' rather than 'Linked field'.

content/applications/productivity/sign/configuration/auto-complete.png
mcsi-odoo, line 1: Just nitpicking here, but if not too much hassle, I'd redo the screenshot to avoid showing the very bottom of the field above 'Door'. (And ... 'door'?? 🤷‍♀️ )

mcsi-odoo (changes requested):
Hi @auva-odoo , when you said there were a lot of new things to add in 19.2, you meant it!! Take a look at my comments and suggestions and let me know if you want to discuss any!

P.S. I have a niggling feeling there was one more thing I wanted to double-check in relation to these updates, but can't for the life of me remember what it was ... so, if that comes back to me, I'll let you know! _(Edited to add: I had managed to forget to review the Configuration page, so maybe that was the niggle! But it's done now)_

content/applications/productivity/sign/request_signatures.rst
auva-odoo: Yes exactly, hence why I originally chose to link these two items with an *and* to try and imply that while those are two different things, they are linked (you can't allow to delegate without allowing to replace, and vice versa).

content/applications/productivity/sign/request_signatures.rst
auva-odoo: Good catch, thanks!

content/applications/productivity/sign/request_signatures.rst
mcsi-odoo: The first time I read it, I felt it _could_ have been interpreted as '... allowing a signer to delegate their signature and [therefore] to be replaced', rather than these being things that are initiated by different people.

content/applications/productivity/sign/request_signatures.rst
auva-odoo: I get your point, but I guess this overlaps with [this](https://github.com/odoo/documentation/pull/17436/changes#r3187319884)
Technically, you could also do both (the signer can delegate and then I can replace the person they delegated to; it probably doesn't make a lot of sense flow-wise, but it is technically possible). The two features are tied together. Maybe we should discuss this?

content/applications/productivity/sign/sign_document.rst
auva-odoo: 1. Sure, let's discuss this, I'm not sure I understand you last point (*And also maybe harmonise how we explain what is updated and where after the delegation/refusal? (e.g., about it being in the signature request details*)
2. Ah, I meant the *signature request's* chatter (which is why I added that reference to the *Signature request details* section in request_signatures.rst). I'll try and clarify this when I rework the paragraphs!
3. Yes, but it feels like an odd use case though? I'll move the note!

content/applications/productivity/sign/request_signatures.rst
auva-odoo: And while checking this, I realized that if one signer has already signed the signature request before it was canceled, their name is still in green with a check mark; only signers who haven't signed yet appear in red but that kinda makes sense to me so I'm not sure I should describe this more?
EDIT: I changed this

content/applications/productivity/sign/request_signatures.rst
auva-odoo: You're totally right (and the way this works really makes no sense to me but 🤐 )
Apparently it will be changed in 19.4 ... Anyways, I changed the desription of the Owner and Authorized users fields and will keep it very simple until they make this (slightly) more logical, if that's OK with you?

content/applications/productivity/sign/request_signatures.rst
auva-odoo: Ohhh! Ah I understand better now ... what about using and/or? it's not the prettiest but do you think that'd help?

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -132,8 +132,8 @@ open it. Alternatively, hover over the template, click the :icon:`fa-ellipsis-v`
 
 .. note::
    When you open a template that has already been used to send a signature request, the template is
-   in *visualization mode*. Clicking :icon:`fa-pencil` :guilabel:`Edit Template` in this case
-   creates a copy of the template, which can then be edited.
+   in *visualization mode*. Click :icon:`fa-pencil` :guilabel:`Duplicate & Edit` to create a copy
```
mcsi-odoo: I missed this before, and only spotted it when I looked more closely at the image again. I guess it would be more correct now to write:


```suggestion
   in :guilabel:`View only mode`. Click :icon:`fa-pencil` :guilabel:`Duplicate & Edit` to create a copy
```

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +299,79 @@ For each template, the following information is shown:
...
-- :guilabel:`Duplicate` a template. The duplicated template can then be
-  :ref:`edited <sign/request-signatures/edit-template>` as needed.
+- :ref:`Edit a template <sign/request-signatures/edit-template>` that has *not yet been used* to
+  send a signature request. Alternatively, click on a template to open it for editing.
+- :guilabel:`Stop Sharing` the link to the templated document or document envelope, if relevant;
+  after sharing is stopped, the link can no longer be opened.
+- :guilabel:`Duplicate` a template. The duplicated template can then be :ref:`edited
+  <sign/request-signatures/edit-template>` as needed.
 - :guilabel:`Archive` a template that is not currently needed.
+- Modify a template's :ref:`settings <sign/request-signatures/template-form>`.
```
mcsi-odoo: Sorry, my suggestion here overlooked the fact that you're actually clicking on **Configuration** to modify the template's settings. So maybe:

```suggestion
- Modify a template's :ref:`settings <sign/request-signatures/template-form>` by clicking :guilabel:`Configuration`.
```

Or there's maybe a better way ...

content/applications/productivity/sign/request_signatures.rst
```diff
@@ -312,14 +299,79 @@ For each template, the following information is shown:
...
+and select :guilabel:`Configuration`.
+
+The following fields are available on the form:
+
+- the template's name
+- :guilabel:`Available in`: Select the :doc:`Odoo model <../../studio/models_modules_apps>` to
+  which the template is linked to restrict the template to :ref:`signature requests made from
+  records of that model <sign/request-signatures/template-odoo-record>`. Leave the field empty to
+  make the template available from any model.
+- :guilabel:`Authorized Users`: Specify the users authorized to modify the template.
```
mcsi-odoo: Sorry, but I'm not remembering now why you were only mentioning 'modify' here, and not something like 'view, use, and modify' ...? (because you need to be an authorized user to even see the template, right?)

mcsi-odoo (commented):
Thanks for taking on board so much of my feedback!
I think the delegate and decline parts work well with this new approach.

Just a few small things here ... otherwise, looks great!

content/applications/productivity/sign/request_signatures.rst
auva-odoo: Whoops, I went to fast, sorry!

mcsi-odoo:
LGTM!

@robodoo delegate+

auva-odoo:
Thank you so much again for the review @mcsi-odoo :pray: 

@robodoo r+

PR #17456, opened 2026-04-20
[IMP] POS: one-click payment and max difference

LoredanaLrpz:
task-6117744

### DIFF
--- a/content/applications/sales/point_of_sale/payment_methods.rst
+++ b/content/applications/sales/point_of_sale/payment_methods.rst
@@ -9,6 +9,35 @@ options, including cash, card payments through a :ref:`configured payment termin
 <pos/terminals/configuration>`, :doc:`online payments <../../finance/payment_providers>`, or
 :doc:`customer accounts <payment_methods/customer_credit>`.
 
+.. cards::
+
+   .. card:: Cash machines
+      :target: payment_methods/cash_machines
+
+      Integrate cash machines to automate cash handling and transaction validation.
+
+   .. card:: Customer account
+      :target: payment_methods/customer_credit
+
+      Allow customers to pay on credit, deposit funds, and manage their debt directly from the POS.
+
+   .. card:: QR code payments
+      :target: payment_methods/qr_code_payment
+
+      Generate QR codes for customers to scan and pay using mobile banking apps.
+
+   .. card:: Payment terminals
+      :target: payment_methods/terminals
+
+      Connect and configure supported card terminals (e.g., Adyen, Stripe, Ingenico) to process card
+      payments.
+
+.. seealso::
+   `Payment methods (video tutorial) <https://www.youtube.com/watch?v=eHr4tS8Wmss>`_
+
+Configuration
+-------------
+
 To create a payment method, go to :menuselection:`Point of Sale --> Configuration --> Payment
 Methods`, click :guilabel:`New`, and follow the next steps:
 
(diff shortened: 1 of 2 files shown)


content/applications/sales/point_of_sale/payment_methods.rst
```diff
@@ -53,6 +53,28 @@ Methods`, click :guilabel:`New`, and follow the next steps:
...
+-----------------
+
+The one-click payment feature allows you to bypass the payment screen for a faster checkout. To
+activate this setting, navigate to the :ref:`POS settings <pos/use/settings>`, scroll down to the
+:guilabel:`Payment` section, and enable :guilabel:`One-click Payment`. Then, select one or multiple
+payment methods in the :guilabel:`Payment Methods` field.
+
+.. note::
+   Online payment, delivery payment, terminal payment, and methods that require customer
+   identification are not supported for fast payment validation.
```
xpl-odoo: I think mentioning methods that _do work_ would be clearer :)

content/applications/sales/point_of_sale/payment_methods.rst
```diff
@@ -53,6 +53,28 @@ Methods`, click :guilabel:`New`, and follow the next steps:
    - Always set the :guilabel:`Journal` to :guilabel:`Bank` when :ref:`configuring a payment
      terminal <pos/terminals/configuration>`.
 
+One-click payment
+-----------------
```
xpl-odoo, line 86: It feels weird to have a dedicated H2 for a single (and simple) option at this level.
Could it be worth describing the other options, if any, under a section called _Configuration_ or something similar, and start with the part about payment method creation? (i.e., "To create a payment method...")

content/applications/sales/point_of_sale/payment_methods.rst
```diff
@@ -53,6 +53,28 @@ Methods`, click :guilabel:`New`, and follow the next steps:
...
+to the cart. The previously selected payment methods are displayed as shortcut buttons next to the
+:guilabel:`Payment` button. Clicking a payment method validates the order immediately and brings
+you to the receipt screen.
+
+.. important::
+   While cash payments are supported, the one-click payment feature assumes the amount received is
+   exact. If the customer requires change, use the standard :guilabel:`Payment` button to enter the
+   amount tendered manually.
+
 .. cards::
```
xpl-odoo: I would move the cards right before the intro. Then have a section about general payment settings?

xpl-odoo (changes requested):

content/applications/sales/point_of_sale/payment_methods.rst
LoredanaLrpz: I mixed two of your comments: I split the beginning to create an introduction, then added a "configuration" h2, (On other pages, the cards appear right after a short introduction)
This way, we have a short intro, the cards, and two h2 titles

content/applications/sales/point_of_sale/payment_methods.rst
LoredanaLrpz: I'm resolving this because I answered with what I did just above =D

content/applications/sales/point_of_sale/payment_methods.rst
LoredanaLrpz: Technically, everything that does not require customer identification, I added this because it's a disclaimer when you enable the feature. I can just remove the note =)

xpl-odoo: The information fits nicely in a note; we can keep it. I was thinking it would be easier/clearer to list the options it works with :sweat_smile: 
But which options don't require some sort of identification, apart from cash?

xpl-odoo (approved):
Thanks for addressing my suggestions!
You can keep the note about one-click payment "restrictions" as is, rather than removing it, but it would be nice if the note was clearer.

@robodoo delegate+

LoredanaLrpz:
@robodoo r+

PR #17513, opened 2026-04-23
[REM] Subscriptions: Scheduled Actions removal

theRealThagomizer:
This 19.0 PR removes the Scheduled Actions documentation from the Subscriptions section. After careful consideration, I believe this documentation covers an edge-case too esoteric for the average Odoo user to get value from its inclusion.


theRealThagomizer:
Hiya, @Felicious! After our chat in the office, I've decided that this document doesn't really add anything of value to the average user and is more likely to do harm via leading them to break something for no benefit. I didn't see anything in the KPI documentation about what this is worth, so I assigned myself two points, reasoning that the initial work to revise the document, the realization that it no longer serves our needs, and the effort to remove it correctly represents a commensurate amount of work. But feel free to adjust as you see fit! Thank you!

Felicious (changes requested):
There seem to be merge conflicts in this?? not sure where it's coming from. 

After fixing that and renaming the PR to use the tag [REM], and editing the commit message to include [REM], I will approve and merge this (:

theRealThagomizer:
@Felicious I think I fixed the merge conflict without breaking anything else! I'll watch robodoo for the rest of the afternoon to ensure it doesn't come back with an exciting new issue.

Felicious (approved):
Looks good to me !

@robodoo r+

PR #17517, opened 2026-04-24
[FIX] tutorials: change link to search view example to be in 19.0

Mathilde411:
The group in the search view still had expand and string which don't work in 19.0


AntoineVDV (approved):
@robodoo r+

PR #17533, opened 2026-04-24
[FIX] rental: restructure rental image folder

meval1006:
Created an image folder for the rental landing page and updated the image path within the RST file


Felicious:
@robodoo r+

PR #17553, opened 2026-04-27
[IMP] conf.py: label "extended support" versions in the version switcher

AntoineVDV:
Versions 14.0, 15.0, and 16.0 have been re-added to the version switcher but were missed the usual "Odoo <version>" label.


AntoineVDV:
@robodoo r+

PR #17611, opened 2026-04-28
[REF] rental: rental section restructuring

meval1006:
Added two new subsections for the Rental section: Configure products and Rental orders. The new structure is the following:

- **Configure products**
  - Products
  - service products
- **Rental orders**
  - Manage deposits 

Also, replaced the old rental setting image with the correct one.

This saas-19.1 PR can be FWP up to master.


Felicious (approved):
LGTM @meval1006 (: 

I'm upping the pts to 2 because there was quite a few places where you had to be careful about getting the content right (: Appreciate your effort!

@robodoo r+

PR #17545, opened 2026-04-25
[IMP] MRP: Basic subcontracting rewrite

stepperpig:
This PR rewrites the basic subcontracting workflow doc. It removes reduplicated and redundant information, corrects the process for configuring/initializing basic subcontracting, and merges content from a soon-to-be-axed doc page.

[Task card](https://www.odoo.com/odoo/my-tasks/6157263/project.task/6157268)

**Changelog**
- Rewrite 'Basic subcontracting' doc for correctness
- Update screenshots
- Merge essential information from 'Lead times' doc

**Planned**
- An additional PR will be opened to remove the 'Basic subcontracting lead times' page

**Scope**
This PR will be FWP up to master. Any differences in functionality between 18.0 and 19.0 will be addressed in future PRs.

PR #17631, opened 2026-04-29
[REM] MRP: Remove Subcontracting Lead Times

stepperpig:
This PR removes extraneous lead times information across all subcontracting workflows docs. These docs both (1) reduplicate information and (2) don't add any new subcontracting-specific info about lead times.

**Changelog**
- Removed 'Basic subcontracting lead times' doc and its images folder
- Removed 'Resupply subcontracting lead times' doc and its images folder
- Removed 'Dropship subcontracting lead times' doc and its images folder
- Updated redirect rules

**Additional Notes**
No content has been added or modified. Only the subcontracting lead times docs have been removed. Existing info about basic subcontracting lead times has already been merged into a [rewrite](https://github.com/odoo/documentation/pull/17545) of the 'Basic' subcontracting route doc. The same will be done for the 'Resupply' and 'Dropship' routes in additional PRs.

**Version Scope**
This 17.0 PR should be FWP to master.


stepperpig:
Hi @Felicious, got a quick PR ready for your review - thank you!

Felicious (approved):
Looks good to me, @stepperpig !

Typically, when we move files, we don't need to add redirect rules, but since the lead times articles were a little repetitive, I agree that we can redirect them to the subcontracting minimum setup page, as they'll find the same info (and more) there (: 

Also, we typically handle one article edit/removal in a single PR, but since you did all 3 here, I'll up the scope to 5 (: 

@robodoo r+

PR #17704, opened 2026-05-04
[IMP] accounting: replace undistributed profit/loss

dade-odoo:
taskid-6123526

Replace the "undistributed profit/loss" account that's been renamed "result brought forward". Also update US and Canadian localization pages that refer to the "undistributed profit/loss" account.


auva-odoo (approved):
Thank you @dade-odoo!

@robodoo r+

PR #17706, opened 2026-05-04
[FIX] consolidation: broken ref link to appropriations

dade-odoo:
Fix the broken ref link to the appropriations account section of the year-end page.


xpl-odoo (approved):
@robodoo r+

PR #17715, opened 2026-05-05
[REM] accounting: remove bank rec tolerance

dade-odoo:
taskid-6005605

FP rolled back the RD task adding a tolerance setting so we'll just remove any mention of a general tolerance and the currency exchange tolerance (OLMA confirmed that it's currently 0% but it will be updated to be 3% in [this task](https://www.odoo.com/odoo/project/967/tasks/6143815)) then we'll update the doc when the RD task is deployed.


xpl-odoo (approved):
@robodoo r+

PR #17726, opened 2026-05-06
[IMP] user_interface: update Gantt documentation

snrav-odoo:
Remove consolidation, consolidation_max and consolidation_exclude attributes as they are no longer used in gantt view.

task-6031386
  
For enterprise PR see: 
https://github.com/odoo/enterprise/pull/112363


Feyensv (approved):
robodoo r+

PR #17771, opened 2026-05-08
[FIX] howtos/website_themes: Typo (Python Model link)

bram-odoo:
This PR fixes a broken link into the layout page (Python Model).

Task-6197487


AntoineVDV (approved):
The diff is correct, but you should target at least 17.0 before merging, as 16.0 is no longer maintained or built.

@robodoo delegate+

bram-odoo:
@robodoo r+

bram-odoo:
> The diff is correct, but you should target at least 17.0 before merging, as 16.0 is no longer maintained or built.
> 
> @robodoo delegate+

Good to know (did I see the line after my r+? Obviously). Thank you, @AntoineVDV

PR #17784, opened 2026-05-08
[IMP] developer/user_interface: remove reference to groups_id

danych2:
Since https://github.com/odoo/odoo/pull/98551, groups_id no longer works in ir.ui.view, so it should be removed from the documentation.


AntoineVDV (approved):
@robodoo r+

PR #17827, opened 2026-05-11
[IMP] odoo_theme: add rotation classes

nikibrown:
For 19.0 and up there is a need to adjust the rotation of the font awesome icons (specifically fa-phone) to match the ui in the Odoo phone app. 

Rather than creating a one off class for the phone icon, I used a sass loop to generate rotation classes in 5 degree increments so future icon adjustments can happen easily. 

Also removed some old CSS that was for internet explorer 9 with rotation issues. 

image

### DIFF
--- a/extensions/odoo_theme/static/scss/_font_awesome.scss
+++ b/extensions/odoo_theme/static/scss/_font_awesome.scss
@@ -82,27 +82,17 @@
     transform: rotate(359deg);
   }
 }
-.fa-rotate-90 {
-  transform: rotate(90deg);
-}
-.fa-rotate-180 {
-  transform: rotate(180deg);
-}
-.fa-rotate-270 {
-  transform: rotate(270deg);
+/* Generate rotation classes from 0 to 360 degrees in 5 degree increments */
+@for $i from 0 through 72 {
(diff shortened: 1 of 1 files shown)


huisit (approved):
@nikibrown works like a charm! This is exactly what we need to replicate the in-product **Hang Up** button, and it'll be great to rotate icons ourselves instead of asking for help every time there's a custom rotation. Thanks for your help on this :)

AntoineVDV (commented):
Hi @nikibrown, you must check that all three CIs are green before marking the PR as ready for review. Then, reviewers will be automatically assigned depending on impacted files, without you needing to manually pick one.

nikibrown:
Thanks @AntoineVDV should be all set now.

nikibrown:
@AntoineVDV Can you take a look at this (or whoever is supposed to review?) We have another PR that needs to use some of these rotation classes to make the phone icons accurate. Thanks!

edi-odoo (approved):
Great job, thanks! LGTM ;)

AntoineVDV (approved):
@robodoo r+

AntoineVDV:
@robodoo squash

PR #17843, opened 2026-05-12
[IMP] website_themes/howtos: Animation on Hover

bram-odoo:
This PR adds some explanation about how the hover effect is applied (image conversion to base64, attachment) and warn the reader about the required manual action to make the effect visible.


AntoineVDV (approved):
@robodoo delegate+

bram-odoo:
@robodoo r+

PR #17913, opened 2026-05-13
[IMP] Phone: Fix wording and add note for OnSIP credentials

fhe-odoo:
Corrected a sentence structure and added an important note about OnSIP credentials.


Felicious:
Closing because the identical PR merged (:

PR #17965, opened 2026-05-18
[FIX] administration/deploy: invalid nginx docs links

Feyensv:
Some links to the nginx documentation aren't working anymore and redirect to the main commercial entity owning nginx.

Nginx doc is now hosted under https://docs.nginx.com, and the links are now adapted accordingly.


Feyensv:
robodoo r+

PR #18000, opened 2026-05-19
[IMP] developer/api/external_api: document programmatic API key management

madprog:
Document the RPC methods introduced by odoo/odoo#246118 to generate and revoke API keys programmatically, and add recommendations about API key rotation best practices.

### DIFF
--- a/content/developer/reference/external_api.rst
+++ b/content/developer/reference/external_api.rst
@@ -136,8 +136,11 @@ Configuration
 
 .. _reference/external_api/api_key:
 
-API Key
--------
+API Keys
+--------
+
+Manual Key Generation
+~~~~~~~~~~~~~~~~~~~~~
 
 An API key must be set in the ``Authorization`` request header, as a bearer token.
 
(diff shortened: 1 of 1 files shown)


content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+Odoo provides RPC methods to generate and revoke API keys programmatically.
+
+Prerequisites
+~~~~~~~~~~~~~
+
+By default, programmatic API key management is restricted to users with the
+*Settings* administration access right.
+
+To allow other users to manage API keys programmatically, enable the system
+parameter ``base.enable_programmatic_api_keys``.
```
Julien00859: How?

You don't *enable* system parameters, you *set them a value*, so what value should be set?

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+~~~~~~~~~~~~~~
+
+API keys can be generated with :meth:`res.users.apikeys.generate`.
+
+The method accepts the following parameters:
+
+- ``key``: an existing valid API key
+- ``scope``: the scope assigned to the new key
+- ``name``: a human-readable label
+- ``expiration_date``: the expiration date of the key
```
Julien00859: add the json datatype of each value, for `expiration_data` is likely is a ISO8601 encoded string.

```suggestion
- ``key`` (string): an existing valid API key
- ``scope`` (string or null): the scope assigned to the new key
- ``name`` (string): a human-readable label
- ``expiration_date`` (string): the expiration date of the key, encoded as a ISO8601datetime
```

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+
+The new key should generally use the same scope as the existing key.
+However, it is possible to create a scoped key from an unscoped key when needed.
+Unscoped keys use ``null`` as their scope value (``None`` in Python).
```
Julien00859: What's a `scope`? The text doesn't explain what it is. The users are likely *not* gonna invent their own and instead use existing ones. As this documentation page is about the RPC APIs, we should at least document the `"rpc"` scope which is the generic one used for all `auth='bearer'` controllers

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+
+The expiration date is validated against the maximum API key duration allowed
+by the roles assigned to the user.
```
Julien00859: We should add recommendations about what lifetime we recommend for what usage. Bonus point if those recommendations are based on existing NIST (US)/ENISA (EU) ones.

content/developer/reference/external_api.rst
Julien00859: alternatively give a (relative) link to the `/doc` page, anchored on that method. Users will have to change the relative URL to add their domain name, but at least the info is precise.

Talking of which, please improve the docstring of the public methods of `res.users.apikeys`. At least please document every single parameter via `:param`.

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+
+By default, a user can generate up to 10 API keys programmatically.
+This limit can be configured with the system parameter
+``base.programmatic_api_keys_limit``.
+
+Attempting to exceed the configured limit raises an error.
+
+The ``key`` argument can seem redundant as it is already present in the Authorization header.
+However, the method itself requires an existing valid API key independently of the transport mechanism.
+For example, the method can also be called from an Odoo shell session.
```
Julien00859: A bit useless information, I wouldn't care to write it.

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+          if (!resSearch.ok) throw new Error(resSearch.json());
+      })();
+
+   .. code-tab:: bash
+
+      set -eu
+
+      API_KEY=
+
+      payload=$(jq -n --arg key "$API_KEY" '{key: $key}')
```
Julien00859: The payload is simple enough to be inlined in the `curl` command, no need to use `jq` here.

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
 <https://cheatsheetseries.owasp.org/cheatsheets/Secrets_Management_Cheat_Sheet.html#secrets-management-cheat-sheet>`_
 for further guidance on the management of API keys.
 
+.. _reference/external_api/api_key_management:
+
+API Key Management
```
AntoineVDV: There is another "API Key" section right above this one that makes the two sections sound like duplicates. It might be better to salvage the first section into this one, and maybe structure the instructions under "Manual key generation" and "Programmatic key generation"?

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+      payload=$(jq -n \
+          --arg key "$API_KEY" \
+          --arg name "Some service" \
+          --arg expiration "2026-05-19" \
+          '{
+            key: $key,
+            scope: null,
+            name: $name,
+            expiration_date: $expiration
+          }')
```
Julien00859: This is the first place where we use `jq`. I would prefer we do not. The `curl` examples are more about giving runnable HTTP example than using curl.

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+- ``expiration_date``: the expiration date of the key
+
+The new key should generally use the same scope as the existing key.
+However, it is possible to create a scoped key from an unscoped key when needed.
+Unscoped keys use ``null`` as their scope value (``None`` in Python).
+
+The expiration date is validated against the maximum API key duration allowed
+by the roles assigned to the user.
+
+Example:
```
AntoineVDV:
```suggestion
.. example::
```

AntoineVDV (approved):
LGTM! Just a few comments on the structure.

@robodoo delegate+

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+      curl https://mycompany.example.com/json/2/res.users.apikeys/revoke \
+          -X POST \
+          --oauth2-bearer $API_KEY \
+          -H "Content-Type: application/json" \
+          -d "$payload" \
+          --silent \
+          --fail
+
+If the key is valid, it is revoked immediately and becomes unusable for subsequent requests.
+Otherwise, an ``AccessDenied`` error is raised.
```
Julien00859: AccessDenied are translated to 403 Forbidden http status code, we better document the status code as users of the API typically will rely on that

content/developer/reference/external_api.rst
```diff
@@ -171,6 +171,212 @@ Please refer to `OWASP's Secrets Management Cheat Sheet
...
+          --fail \
+          | jq -r .
+      )
+      # store the new key securely
+
+The method returns the newly generated API key as a string that cannot be retrieved otherwise.
+
+By default, a user can generate up to 10 API keys programmatically.
+This limit can be configured with the system parameter
+``base.programmatic_api_keys_limit``.
```
Julien00859: For consistency. You're adding a newline after every sentence but the rest of this RST file doesn't do that. I don't have a problem *new* files use their own style, but I *do care a lot* about consistency *within* a file.

Julien00859 (changes requested):

content/developer/reference/external_api.rst
Julien00859: The relative link: `/doc/res.users.apikeys#generate`

Test it: http://localhost:8069/doc/res.users.apikeys#generate

Julien00859 (approved):
Thanks!  cc @AntoineVDV ready for me :)
I did not verified the rendering

AntoineVDV:
@Julien00859 damn you went ballistic on this one :grin: 

Nothing to add from my side, as I only focused on the structure and RST, trusting you with the content itself.

I already delegated the merge to PMO; adding you too in case you're reviewing the task.

@robodoo delegate=Julien00859

madprog:
Thanks guys for the reviews! :+1: 
@robodoo r+

[EDIT] the rendered result is available [here](https://runbot202.odoo.com/runbot/static/build/111363430-19-0/logs/build/html/developer/reference/external_api.html#programmatic-key-management).

PR #18001, opened 2026-05-19
[ADD] l10n_uz - add new uzbek localization

dade-odoo:
taskid-6178882

### DIFF
--- a/content/applications/finance/fiscal_localizations/uzbekistan.rst
+++ b/content/applications/finance/fiscal_localizations/uzbekistan.rst
@@ -1,0 +1,116 @@
+==========
+Uzbekistan
+==========
+
+.. |IFRS| replace:: :abbr:`IFRS (international financial reporting standards)`
+
+.. _localizations/uz/modules:
+
+Modules
+=======
+
+The following modules are installed automatically with the Uzbek localization:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 25 50
+
+   * - Name
+     - Technical name
+     - Description
+   * - Uzbekistan - Accounting
+     - l10n_uz
+     - Default fiscal localization package for Uzbekistan. Includes the chart of accounts and core
+       accounting configuration aligned with local requirements.
+   * - Uzbekistan - Accounting Reports
+     - l10n_uz_reports
+     - Provides localized financial reports, including the balance sheet and profit and loss
+       statement based on an |IFRS|-style structure.
+
+.. _localizations/uz/overview:
+
+Localization overview
+=====================
+
+The Uzbekistan localization assists companies in meeting local accounting requirements by providing
+a structured and compliant accounting setup. It contains the following core components:
+
+- :ref:`Taxes <localizations/uz/taxes>`: preconfigured tax rates, including standard 12% VAT,
+  zero-rated export, and exempt options
+- :ref:`Chart of accounts <localizations/uz/coa>`: preconfigured structure aligned with Uzbek
+  standards
+- :ref:`Financial reports <localizations/uz/taxes>`: |IFRS|-style balance sheet and
+  profit and loss reports with a tag-based linking of accounts to report lines
+
+.. _localizations/uz/taxes:
+
+Taxes
+-----
+
+The Uzbek localization includes preconfigured sales and purchase taxes:
+
+- **Standard VAT (12%):** This standard VAT is applied to most goods and services. The supplier
+  charges 12% VAT on the sale (VAT output) and is generally entitled to deduct the VAT
+  paid on business purchases (VAT input).
+
+  .. example::
+     Most commercial sales and general supply of goods and services within Uzbekistan use the
+     **Standard VAT (12%)** tax.
+
+- **Export VAT (0%):** The export tax is applied on taxable supplies where the VAT rate applied is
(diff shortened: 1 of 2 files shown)


content/applications/finance/fiscal_localizations/uzbekistan.rst
```diff
@@ -0,0 +1,118 @@
...
+  .. example::
+     Exports of goods and services to customers outside Uzbekistan use the **Export VAT (0%)** tax.
+
+- **Exempt transactions (0%):** The exempt tax applies to transactions that fall outside the scope
+  of VAT and are not subject to tax.
+
+  .. example::
+     Transactions that are legally exempt from VAT under local regulations such as essential food
+     products (meat, fish, potatoes, rice), pharmaceuticals, and financial/insurance services use
+     the **Exempt transactions (0%)** tax.
```
mcsi-odoo:
```suggestion
     Transactions that are legally exempt from VAT under local regulations, such as essential food
     products (meat, fish, potatoes, rice), pharmaceuticals, and financial/insurance services, use
     the **Exempt transactions (0%)** tax.
```

content/applications/finance/fiscal_localizations/uzbekistan.rst
```diff
@@ -0,0 +1,118 @@
...
+     - Technical name
+     - Description
+   * - Uzbekistan - Accounting
+     - l10n_uz
+     - Default fiscal localization package for Uzbekistan. Includes the chart of accounts and core
+       accounting configuration aligned with local requirements.
+   * - Uzbekistan - Accounting Reports
+     - l10n_uz_reports
+     - Provides localized financial reports, including the balance sheet and profit and loss
+       statement based on |IFRS|-style structure.
```
mcsi-odoo:
```suggestion
       statement based on an |IFRS|-style structure.
```

(as you write later)

content/applications/finance/fiscal_localizations/uzbekistan.rst
```diff
@@ -0,0 +1,118 @@
...
+
+.. example::
+   Accounts that use the tag :guilabel:`BS Line 0120: Right-of-Use Assets` appear on the balance
+   sheet on the :guilabel:`[0120] - Right-of-Use Assets` line.
+
+.. important::
+   If no tag is assigned:
+
+   - The account behaves as an off-balance sheet account.
+   - The account is not included in financial reports.
```
mcsi-odoo: Not super clear to me how this works (but maybe obvious to the intended audience!), and maybe you're familiar enough with the concept to help me understand?

In the Financial reports section, you mention that 'each account is assigned a tag'. Are these standardized (non-Odoo?) tags? Are they automatically assigned upon installation or does the user have to set them up? If automatic, how would it happen that an account has no tag assigned?

content/applications/finance/fiscal_localizations/uzbekistan.rst
```diff
@@ -0,0 +1,118 @@
...
+
+The Uzbekistan localization assists companies in meeting local accounting requirements by providing
+a structured and compliant accounting setup. It contains the following core components:
+
+- :ref:`Taxes <localizations/uz/taxes>`: preconfigured tax rates, including standard 12% VAT,
+  zero-rated export, and exempt options
+- :ref:`Chart of accounts <localizations/uz/coa>`: preconfigured structure aligned with Uzbek
+  standards
+- :ref:`Financial reports <localizations/uz/taxes>` (|IFRS|-style): |IFRS|-style balance sheet and
+  profit and loss reports with tag-based structure linking accounts to report lines
```
mcsi-odoo:
```suggestion
- :ref:`Financial reports <localizations/uz/taxes>`: |IFRS|-style balance sheet and
  profit and loss reports with tag-based linking of accounts to report lines
```

- I think we could drop the first 'IFRS-style', no?
- something about 'tag-based structure' feels odd, especially when we're also used 'IFRS-style _structure_' elsewhere. See what you think of my suggestion.

mcsi-odoo (changes requested):
Hi @dade-odoo , a few little nitpicks and suggestions, and some questions (from an accounting noob) about the account tags 🙃

content/applications/finance/fiscal_localizations/uzbekistan.rst
dade-odoo: Hi @mcsi-odoo , all valid questions!

First of all, there's an account type in Odoo called "Off-Balance Sheet" that hides the account from the Balance Sheet. I don't know of any other differences for Off-Balance Sheet accounts, so maybe we should just merge these two bullet points into something like

> If no tag is assigned, the account behaves as an off-balance sheet account and is not included in financial reports.

This tag-based structure of building reports is not standard in Odoo - Uzbekistan has the only localization that I know of that uses it. But to answer your question, it is standard in the Uzbek localization and the accounts do come pre-configured with these tags. That being said, the chart of accounts that comes with the localization is just a recommended starting point, and most businesses will remove accounts they don't need and add new ones (for example, they might want multiple accounts receivable, or multiple income accounts, or multiple expense accounts, etc.). When users create these new accounts, they will need to add the appropriate tags to them so they appear in the correct sections of their financial reports.

mcsi-odoo (approved):
Looks good to me, @dade-odoo !

@robodoo r+

PR #18014, opened 2026-05-20
[IMP] ecommerce: add orders overview

lman-odoo:
task-6047402


content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
...
+- :guilabel:`To Confirm`: unpaid and unconfirmed orders (i.e., :ref:`sent but unconfirmed quotations
+  <ecommerce/handling/sales/quotation-sent>`)
+- :guilabel:`To Invoice`: confirmed orders that are ready for invoicing but have not necessarily
+  been fulfilled (i.e., :ref:`sales orders <ecommerce/handling/sales/sales-order>`)
+
+.. image:: order_handling/order-status-buttons.png
+   :alt: Orders overview.
+
+Select a specific order from the orders list to view its details and order status. The bar at the
+top of the order shows its current status:
```
mcsi-odoo:
```suggestion
Select an order from the list to view its details. The status bar at the
top of the order shows its current status:
```

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
...
+
+  .. _ecommerce/handling/sales/quotation:
```
mcsi-odoo:
```suggestion
  .. _ecommerce/handling/sales-quotation:
```

As far as I know, we should only ever have three levels, so I do like this in my own doc

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
...
 
 - :guilabel:`Quotation`: A product has been :ref:`added to the cart
   <ecommerce/checkout/add-to-cart>`, but the customer has not yet completed the :ref:`checkout
   process <ecommerce/checkout/steps>`. If the customer is not logged in, the quotation is assigned
   to the default *Public User* contact.
+
+  .. _ecommerce/handling/sales/quotation-sent:
```
mcsi-odoo:
```suggestion
  .. _ecommerce/handling/sales-quotation-sent:
```

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
...
+
 - :guilabel:`Quotation Sent`: The customer has completed the :ref:`checkout process
   <ecommerce/checkout/steps>` and confirmed the order, but the :ref:`payment
   <ecommerce/checkout/payment>` has not yet been confirmed.
+
+  .. _ecommerce/handling/sales/sales-order:
```
mcsi-odoo:
```suggestion
  .. _ecommerce/handling/sales-sales-order:
```

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
...
-   - To view only orders in the :guilabel:`Quotation Sent` stage, go to :menuselection:`Website -->
-     eCommerce --> Orders` and remove the :guilabel:`Confirmed` filter. Then, open the
-     :icon:`fa-caret-down` dropdown menu and, under :guilabel:`Filters`, select :guilabel:`Unpaid`.
+   - To view only orders with the :guilabel:`Quotation Sent` status, click the :guilabel:`To
+     Confirm` button or go to :menuselection:`Website --> eCommerce --> Orders` and remove the
+     :guilabel:`Confirmed` filter. Then, open the :icon:`fa-caret-down` dropdown menu and, under
+     :guilabel:`Filters`, select :guilabel:`Unpaid`.
+   - Reporting data, such as the number of :guilabel:`Visitors`, :guilabel:`Orders` and
+     the :guilabel:`Sales` amount of a given period, is displayed on the left side of the
+     :guilabel:`Orders` page.
```
mcsi-odoo:
```suggestion
   - The number of :guilabel:`Visitors`, :guilabel:`Orders`, and the :guilabel:`Sales` amount for a given period is displayed on the right side of the :guilabel:`Orders` page dashboard.
```

- Since there are only these three elements, I wouldn't use 'Reporting data, _such as_ ...'
- I'm seeing them on the right, no?

Image

- Since you see this when on the Orders page, I think we need to move it up (well, see all other comments 🙃 )

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -185,15 +211,15 @@ To automate invoicing:
 #. Go to :menuselection:`Website --> Configuration --> Settings`.
 #. In the :guilabel:`eCommerce` section, enable :guilabel:`Automatic Invoice`.
 #. Select the relevant :guilabel:`Email Template`, then click :guilabel:`Save`.
-#. If needed, click the :icon:`fa-arrow-right` icon next to the template name to customize the
-   :doc:`email template </applications/general/companies/email_template>`.
+#. If needed, click the :icon:`fa-arrow-right` (:guilabel:`Internal link`) icon next to the template
+   name to customize the :doc:`email template </applications/general/companies/email_template>`.
```
mcsi-odoo, line 217: [Feel free to disregard this and the next comment if you don't want to deal with them here/now ... I'm only commenting 'cause you touched them 😉]

I know you were just adding an 'Internal link' here ... but ... it's strange to have this as a step 4 _after_ having saved, no? Could we imagine it as a tip instead? Or would that be too many admonitions in a row?

Let me know your thoughts!

(Same comment for the Abandoned carts section actually. But again, feel free not to act on this now!)

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
 ====
```
mcsi-odoo, line 20: A few thoughts about the updates:

- I think we should be referring to that new panel as a dashboard; I feel that's what we usually call this kind of this (e.g., in the Purchase app, there's similar). In the Purchase app, they refer to 'buttons', but I was thinking something like KPI widgets ... or maybe 'scorecards', as we call that kind of 'chart' in the Dashboards doc. To be discussed.

- Shouldn't we also cover the time period selector? I don't see that mentioned here. And let's keep the 'reporting data' (visitors, orders, sales stats) together with the rest about the dashboard. right now it feels lost/confusing to have it _after_ the order form page.

- Right now, I feel we're kind of jumping around a bit between the new dashboard section, the form view of the record, and the possibility to filter using the search bar as on any other Odoo screen. I'm wondering if maybe you can just start with saying 1) the Orders page is by defult filtered to show confirmed orders, then 2) there's also a handy dashboard where they can use one of the available buttons/widgets and select the time period, then 3) go on to explain briefly what you see when you open a sales order record, including the status bar.

- These tips are a bit confusing now, since they come after we've talked about the dashboard and after we've talked about the individual order forms. Maybe we can find a better way to structure this whole part.

Image

Let's talk about this one day, as I'm not being very clear or organised here!

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,25 +20,51 @@ Sale
...
+
 - :guilabel:`Quotation Sent`: The customer has completed the :ref:`checkout process
   <ecommerce/checkout/steps>` and confirmed the order, but the :ref:`payment
   <ecommerce/checkout/payment>` has not yet been confirmed.
+
+  .. _ecommerce/handling/sales/sales-order:
+
 - :guilabel:`Sales Order`: The customer has completed the :ref:`checkout process
   <ecommerce/checkout/steps>`, confirmed the order, and the :ref:`payment
   <ecommerce/checkout/payment>` has been successfully received.
```
mcsi-odoo: I think I'm generally struggling with the fact that we're using different language to explain the filters and to explain these three stages of an order. Is there any way we could make this clearer to the ecommerce user?

I'm also generally a bit confused about some wording, like:

- what do we mean by 'has completed the checkout process'? If I click the link, I'm seeing payment and order confirmation as steps
- what do we mean by 'the customer has ... confirmed the order'? I'm only really seeing the 'order confirmation' as the final step of the checkout steps. (Also, in 'To Confirm' we're mentioning 'unconfirmed orders' ... so I guess we're using this term to mean two different things?) 

_#confused #ecommercenoob_

mcsi-odoo (changes requested):
Hi Larissa! Nice that they've added this overview thing at the top of the orders page, but it's not the easiest to explain clearly. I think it'll be good for us to chat a bit about this in person, but I still wanted to share my ramblings with you in case it can give you a little headstart on understand some of my thoughts ...

content/applications/websites/ecommerce/order_handling.rst
lman-odoo: Oh I didn't know that :) Thank you

content/applications/websites/ecommerce/order_handling.rst
lman-odoo: Ok, thank you for the detailed comment.

What I plan to do now:

1. Use the wording "Dashboard" and "buttons."
2. Do you mean the time period selector of the reporting data? As it just changes the reporting data, not the filter buttons
3. I restructured: main view - dashboard and reporting - form view
4. Restructured the tips

content/applications/websites/ecommerce/order_handling.rst
lman-odoo: Hahah, this is a case of left-right disorientation 😆 Sorry, you are right, that is what I meant

content/applications/websites/ecommerce/order_handling.rst
lman-odoo: Hmm, actually, I would leave it like this. For me, it makes sense to add it as a final step, as you probably want to have a look at the template. And you can only click once you've saved the setting's activation. You cannot click it without saving first.

content/applications/websites/ecommerce/order_handling.rst
lman-odoo: I know what you mean. Especially because the filters do not exactly match the order document status. 

You can confirm the order without paying for it (mostly when using a wire transfer). Unconfirmed orders are orders in the Quotation and Quotation Sent stages. They are not confirmed in Odoo when the payment has not been received. Even if the customer completed the entire checkout process, they still appear as Unconfirmed if the payment has not been received.

lman-odoo: EDIT: I think we could remove the "confirmed the order" wording, as it could be misleading.

content/applications/websites/ecommerce/order_handling.rst
lman-odoo: Changed this sentence to create a better transition :)

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,29 +20,51 @@ Sale
...
 
 To view all quotations and orders, navigate to :menuselection:`Website --> eCommerce -->
-Orders`, and select one to open it. The status bar at the top of the order shows its current status:
+Orders`. By default, the :guilabel:`Confirmed` filter is enabled. Remove this filter to display all
+records.
+
+.. _ecommerce/handling/sales-dashboard:
+
+The dashboard at the top of the :guilabel:`Orders` page provides an overview of all orders and
+includes the following filter buttons:
```
mcsi-odoo:
```suggestion
The dashboard at the top of the :guilabel:`Orders` page provides a summary of the orders and
includes the following filter buttons:
```

Maybe 'summary' is more appropriate than 'overview'? And maybe we leave out 'all' since it doesn't include orders with the status 'quotation'.

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,29 +20,51 @@ Sale
...
+
 - :guilabel:`Quotation Sent`: The customer has completed the :ref:`checkout process
-  <ecommerce/checkout/steps>` and confirmed the order, but the :ref:`payment
-  <ecommerce/checkout/payment>` has not yet been confirmed.
-- :guilabel:`Sales Order`: The customer has completed the :ref:`checkout process
-  <ecommerce/checkout/steps>`, confirmed the order, and the :ref:`payment
-  <ecommerce/checkout/payment>` has been successfully received.
+  <ecommerce/checkout/steps>`, but the :ref:`payment <ecommerce/checkout/payment>` has not yet been
+  confirmed (Orders with this status are displayed when the :guilabel:`To Confirm` :ref:`dashboard
+  button <ecommerce/handling/sales-dashboard>` is selected).
```
mcsi-odoo: I definitely think it's good to give the 'Orders with this status ...' info here, but maybe just lower case 'orders'?

content/applications/websites/ecommerce/order_handling.rst
```diff
@@ -20,29 +20,51 @@ Sale
...
-     :icon:`fa-caret-down` dropdown menu and, under :guilabel:`Filters`, select :guilabel:`Unpaid`.
-   - E-commerce orders can be automatically assigned to a specific sales team or salesperson. To do
-     so, go to :menuselection:`Website --> Configuration --> Settings`. In the :guilabel:`eCommerce`
-     section, under :guilabel:`Orders Assignment`, select the relevant :guilabel:`Sales Team` or
-     :guilabel:`Salesperson`.
+  .. _ecommerce/handling/sales-sales-order:
+
+- :guilabel:`Sales Order`: The customer has completed the :ref:`checkout process
+  <ecommerce/checkout/steps>`, and the :ref:`payment <ecommerce/checkout/payment>` has been
+  successfully received (Orders with this status are displayed when the :guilabel:`To Fulfill` or
```
mcsi-odoo: As above

mcsi-odoo (approved):
Hey @lman-odoo , I really think it works better like this. Nice job. And yep, I was overthinking things on Thursday 🙃  Just a last few thoughts/suggestions ... otherwise, looks good to me!

@robodoo delegate+

lman-odoo:
@robodoo r+

PR #18017, opened 2026-05-20
[FIX] developer/tutorials: Button example without deprecated attrs

Megaaaaaa:
Using attrs to make a button invisible is deprecated since 17.0 but the button example hyperlink sends to a commit that used it so it induces people following the tutorial to use attrs which shouldn't be done anymore.

This commit replaces the link with the commit that replaced the `attrs=""` to the plain `invisible=""` use on the exact same button. (And adjusts the size of the delimiter to respect documentation guidelines.)


AntoineVDV (approved):
@robodoo r+

PR #18032, opened 2026-05-21
[IMP] iot: remove mentions to device subtype.

loouis-t:
In odoo/enterprise#113305, we removed the subtype field to simplify the IoT Device model.

We now remove occurrences of "subtype" in doc.


LoredanaLrpz:
@robodoo r+

PR #18036, opened 2026-05-21
[IMP] accounting: oman invoicing language

tohor-odoo:
taskid-6209590

Up to master.

### DIFF
--- a/content/applications/finance/fiscal_localizations/oman.rst
+++ b/content/applications/finance/fiscal_localizations/oman.rst
@@ -96,3 +96,22 @@ reporting period.
      <https://tms.taxoman.gov.om/portal/web/taxportal/home>`_.
    - `VAT Taxpayer Guide
      <https://tms.taxoman.gov.om/portal/documents/20126/1414820/VAT+Taxpayer+Guide+-+VAT+Return+Filing.pdf/fae31d0c-e7e7-0014-9f28-a4c506b36614?t=1733169613409>`_.
+
+Invoicing language
+==================
+
+Invoices can be issued in different languages according to your customer or business needs. By
+default, billing documents are sent to customers or vendors in the language set on their
+:doc:`contact form <../../essentials/contacts>`.
+
+It is also possible to add Arabic as a second language to your company's billing documents if
+desired or required. To do so, go to :menuselection:`Configuration --> Settings` in the
+**Accounting** app, then, in the :guilabel:`Customer Invoices` section, enable :guilabel:`Gulf
+Cooperation Council Format`.
+
+When enabled, all invoices, credit/debit notes, vendor bills, and refund bills feature both the
+language set on your customer or vendor contact form and an Arabic translation.
+
+.. tip::
+   Click :guilabel:`Preview` after confirming an invoice to see exactly how the invoice will appear
+   to the customer or vendor.


content/applications/finance/fiscal_localizations/oman.rst
```diff
@@ -96,3 +96,24 @@ reporting period.
...
+
+- Bilingual Arabic invoices: Add Arabic as a secondary language to your company's billing documents.
+  When enabled, all invoices, credit/debit notes, vendor bills, and refund bills will feature both
+  your company's primary language and an Arabic translation. To configure this, open the
+  **Accounting** app, go to :menuselection:`Configuration --> Settings`, and enable :guilabel:`Gulf
+  Cooperation Council Format` in the :guilabel:`Customer Invoices` section.
+
+.. tip::
+   To test these features, create and confirm an invoice. Then, click :guilabel:`Preview` on that
+   invoice to see exactly what the document will look like for your customer or vendor.
```
mcsi-odoo: Maybe it's just me, but it almost sounds like we're suggesting they create and confirm some kind of test/dummy invoice, whereas I guess you want to encourage them to check a preview before sending to see the changes they've made. So I think we can be more concise and general with our tip.

How about something like:

```suggestion
.. tip::
   Click :guilabel:`Preview` after confirming an invoice to see exactly how the invoice will appear to the customer or vendor.
```

content/applications/finance/fiscal_localizations/oman.rst
```diff
@@ -96,3 +96,24 @@ reporting period.
...
+- Per-customer language: Set a preferred langauge for individual customers. For example, if a
+  customer's language is set to Turkish, all documents sent to them are automatically generated in
+  Turkish. To configure this, open the customer's :doc:`contact form <../../essentials/contacts>`
+  and set the :guilabel:`Language` field to the desired language.
+
+- Bilingual Arabic invoices: Add Arabic as a secondary language to your company's billing documents.
+  When enabled, all invoices, credit/debit notes, vendor bills, and refund bills will feature both
+  your company's primary language and an Arabic translation. To configure this, open the
+  **Accounting** app, go to :menuselection:`Configuration --> Settings`, and enable :guilabel:`Gulf
+  Cooperation Council Format` in the :guilabel:`Customer Invoices` section.
```
mcsi-odoo: Since the first method mentioned is the default Odoo setup, i.e., documents sent in the language set for that contact on their contact form, I don't think it makes sense to spell it out so explicitly. (This would also be relevant for Saudi Arabia and UAE if we would make this change here.)

I'd suggest something like:

```suggestion
By default, billing documents are sent to customers or vendors in the language set on their :doc:`contact form <../../essentials/contacts>`. 

It is also possible to add Arabic as a second language to your company's billing documents if desired, or if required under XYZ. To do so, go to :menuselection:`Accounting --> Configuration --> Settings`, then, in the :guilabel:`Customer Invoices` section, enable :guilabel:`Gulf Cooperation Council Format`.

When enabled, all invoices, credit/debit notes, vendor bills, and refund bills feature both your company's primary language and an Arabic translation. 
```

Actually, in the Saudi Arabia and UAE 'Invoicing language' sections, they don't specify that after enabling Gulf Cooperation Council Format, documents would be in _the company's primary langauge_ + Arabic (rather than per-customer language + Arabic). Are you sure here that it would be company (and what do we mean by that ... DB language?) language + Arabic?

mcsi-odoo (changes requested):
Hi @tohor-odoo ! Let's have a quick chat about this next time we're both in the office 🇴🇲

content/applications/finance/fiscal_localizations/oman.rst
tohor-odoo: I had a another look and the correct info is indeed the language set on the contact form + Arabic, rather than the company's DB language. I corrected the info. Thanks for pointing it out!

mcsi-odoo (changes requested):
Looks good! Thanks @tohor-odoo !

@robodoo r+

mcsi-odoo (approved):

PR #18065, opened 2026-05-21
[REM] inventory: remove "sell from multiple/virtual warehouses" article

huisit:
remove doc from all versions, per BE experts' request


Felicious:
@robodoo r+

PR #18073, opened 2026-05-22
[IMP] Inventory: RFID: CipherLab Multi-Tag Output public link

adr-odoo:
CipherLab now provides a public link where users can easily download the APK without having to contact their supplier or CipherLab directly.


Felicious:
@robodoo r+

PR #18134, opened 2026-05-28
[FIX] Update the first tip in Keyboard Shortcut to include Alt key

SedjamTan:
Windows/Linux hold the Alt key to view the keyboard shortcuts assigned to each element on the interface


SedjamTan:
I think this behavior is present 17.0 and onward, but I have only tried 19.0

PR #18184, opened 2026-05-30
[FIX] mail:move mailbox part domains to incoming mail doc

jupe-odoo:
TO DO


Abridbus:
handled here: https://github.com/odoo/documentation/pull/18187

PR #18235, opened 2026-06-03
[FIX] payment_providers/nuvei: Eliminate Ecuador as supported country

masi-odoo:
Although our agreement with Nuvei included Ecuador as a supported country, issues have arised from Nuvei not working for the country.

We want to remove this from the official documentation so clients are not misleaded.


clbr-odoo:
@robodoo help

auva-odoo (approved):
@robodoo r+

PR #18307, opened 2026-06-08
[IMP] developer/orm: 19.4 changelog update

kmagusiak:


kmagusiak:
@odoo/doc-review Could you delegate+?

Feyensv:
Merging the changelog before some of those PRs are even merged seems strange to me :thinking: 
Also, please simply assign us as reviewers once it's ready, we'll take a look asap then.

robodoo delegate+ delegate=rco-odoo

kmagusiak:
> Merging the changelog before some of those PRs are even merged seems strange to me 🤔

The list is from what is planned (only the first one wait for merging and I removed one which got reverted).

rco-odoo (approved):
@robodoo r+

PR #18357, opened 2026-06-10
[IMP] odoo_theme: add exact phrase matching for quoted search terms

aksi-odoo:
Previously, all searches were fuzzy/partial.
This patch adds exact phrase matching:
i.e. terms enclosed in " " or ' ' are treated as exact phrases.

How it works:
- Quoted phrases are extracted before tokenization so they bypass stemming
- The remaining query runs through the normal index-based search to get candidates
- Each candidate page is fetched and checked for the exact phrase verbatim
- Non-matching pages are filtered out before display
- Highlights and summary excerpts are scoped to phrase words only, not stemmed variants

`searchtools.js` is patched directly and placed in `_static/` so Sphinx picks it up over the default bundled version.

see diff : https://www.diffchecker.com/Nk59cgho/

Before fix:

image

After fix

image

### DIFF
--- a/extensions/odoo_theme/static/searchtools.js
+++ b/extensions/odoo_theme/static/searchtools.js
@@ -1,0 +1,571 @@
+/*
+ * searchtools.js
+ * ~~~~~~~~~~~~~~~~
+ *
+ * Sphinx JavaScript utilities for the full-text search.
+ *
+ * :copyright: Copyright 2007-2021 by the Sphinx team, see AUTHORS.
+ * :license: BSD, see LICENSE for details.
+ *
+ * Modified to support exact phrase matching when search terms are enclosed in quotes.
+ */
+
+if (!Scorer) {
+  /**
+   * Simple result scoring code.
+   */
+  var Scorer = {
+    // Implement the following function to further tweak the score for each result
+    // The function takes a result array [filename, title, anchor, descr, score]
+    // and returns the new score.
+    /*
+    score: function(result) {
+      return result[4];
+    },
+    */
+
+    // query matches the full name of an object
+    objNameMatch: 11,
+    // or matches in the last dotted part of the object name
+    objPartialMatch: 6,
+    // Additive scores depending on the priority of the object
+    objPrio: {0:  15,   // used to be importantResults
+              1:  5,   // used to be objectResults
+              2: -5},  // used to be unimportantResults
+    //  Used when the priority is not in the mapping.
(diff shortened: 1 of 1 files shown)


aksi-odoo:
Snippets added by me are marked with a` [PATCH]` comment above them to make review easier.

AntoineVDV (changes requested):
Hi @aksi-odoo, thanks for putting in the effort to tackle a frustrating issue.

After review, I cannot merge this as-is, the main issue being the approach: I'm not convinced that overwriting Sphinx's `searchtools.js` by dropping an entire 570-line copy with scattered patches is a good idea. That would mean owning the entire Sphinx search implementation from now on. Every Sphinx upgrade would be a manual diff exercise to reapply the patches at the right, possibly rewritten, place into the new file. That effort is not proportional to the added value.

On the functional side, this does address one of the known pain points, but not the several other issues we've identified.

I'm closing this for now, but I'd be happy to discuss a cleaner approach.

aksi-odoo:
Hello @AntoineVDV  Thanks for review :)

You're right that owning a full copy of `searchtools.js` has a maintenance cost, and i am aware of that trade-off. The reason i went this route is that Sphinx's search is not designed to be extended from the outside. I was not able to find any  hook or override point/plugin interface. 
The exact phrase filtering has to happen inside the `query()` and `displayNextItem()` functions, which means touching the file directly is the only viable option short of replacing the entire search system.

On the upgrade concern: we're currently on Sphinx [4.3.2 ](https://github.com/odoo/documentation/blob/19.0/requirements.txt#L5)and haven't bumped it since. The latest is 9.x.x. 
we're far from the bleeding edge here. So i thought a Sphinx upgrade is not something we do frequently. The diff exercise you're describing is a real but infrequent cost.

If there's a preferred approach for patching Sphinx internals in this repo, happy to follow it. Otherwise we think the trade-off is acceptable given the constraints.

AntoineVDV:
I understand the constraints, but sometimes it's better to avoid going the hard route than living with the real consequences for a small added value.

Sphinx hasn't been upgraded in a while because it's such a painful exercise that we only do it when required; we also avoid upgrading to the bleeding-edge release and stick to battle-tested ones. So, it's still important not to consider that it won't happen again.

Rather than patching Sphinx to fix a single search issue, a better, easier-to-maintain approach would be to use OWL 3 to revamp the search experience completely and fill all gaps at once.

PR #18414, opened 2026-06-12
[IMP] pos: auto course allocation new section

LoredanaLrpz:
task-6289913

### DIFF
--- a/content/applications/sales/point_of_sale/restaurant.rst
+++ b/content/applications/sales/point_of_sale/restaurant.rst
@@ -314,23 +314,76 @@ the target table in the :ref:`Floor plan <pos/restaurant/floors>` view:
 Courses
 -------
 
-The :guilabel:`Course` button allows for splitting orders into multiple courses, sending each course
-to the kitchen sequentially.
+The **courses** feature allows restaurants to organize orders into distinct serving phases (such as
+starters, mains, and desserts) and manage when they are sent to the kitchen.
 
-To split an order into courses from the :ref:`register <pos/restaurant/orders>`, click
-:guilabel:`Course` and add products. Repeat the action as many times as needed, then click
-:guilabel:`Send` to send the order to the kitchen, which also fires the first course.
+Manual course allocation
+~~~~~~~~~~~~~~~~~~~~~~~~
 
-When ready for the second course, retrieve the order from the :ref:`Floor plan
-<pos/restaurant/floors>` view or the :guilabel:`Orders` overview, and click :guilabel:`Fire Course
-2`. Repeat the action as many times as needed.
+To split an order into courses directly from the :ref:`register <pos/restaurant/orders>`:
+
+- Click the :guilabel:`Course` button, then add the products for that course. Repeat this process as
+  needed.
+- Alternatively, click :guilabel:`Course` multiple times to generate the desired number of courses
+  in the cart first. Then, click each course individually to select it, and add its respective
+  products.
+
+Automatic course allocation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The automatic course allocation feature automatically sorts products by course depending on their
+:ref:`POS product category <pos/products/categories>`.
+
+To activate the feature, go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll
+down to the :guilabel:`Preparation` section, enable :guilabel:`Automatic course allocation`, then
+click :guilabel:`Save`.
+
+To configure courses and assign them to POS product categories, follow these steps:
+
+#. Go to :menuselection:`Point of Sale --> Configuration --> POS Product Categories` and select a
+   product category (e.g., *Burgers*).
+#. In the :guilabel:`Course` field, select an existing course or type a new one (e.g., *Main*) and
+   click :guilabel:`Create`.
+
+Once configured, adding a product from the *Burgers* category to an order automatically assigns it
+to the *Main* course.
+
+.. note::
+   - A single course can be assigned to multiple POS product categories. For example, the *Main*
+     course can be used for the *Burgers*, *Pizzas*, and *Meat* categories.
+   - You can create as many courses as necessary.
 
 .. tip::
-   - Alternatively, click :guilabel:`Course` as often as needed to display the desired number of
-     courses in the cart. Then, click each course, add products, and click :guilabel:`Send`.
-   - To transfer a product or an entire course into another course, select it in the cart, click
-     the :icon:`fa-ellipsis-v` (:guilabel:`Actions`) icon, then :icon:`fa-arrow-down`
-     :guilabel:`Transfer course`, and select the preferred course.
+   By default, courses appear in the cart in the order they were created. To rearrange the course
+   order to reflect your service flow, activate the :ref:`developer mode
+   <developer-mode/activation>`, go to :menuselection:`Point of Sale --> Configuration --> Courses`,
+   then use the :icon:`oi-draggable` :guilabel:`(draggable)` icon to move courses to the desired
+   position.
+
+Fire courses
+~~~~~~~~~~~~
+
+Once an order is complete, click :guilabel:`Send` at the bottom left of the register to transmit it
+to the kitchen. The first course appears on the :doc:`preparation display <extra/preparation>`
+without a label, indicating it is waiting to be prepared. Subsequent courses are labeled as
+:guilabel:`Pending` until they are ready to be fired.
+
+.. note::
+   The text on the :guilabel:`Send` button dynamically includes the POS categories of the items
+   being sent to the kitchen. If there are too many categories, the word *Send* is hidden, and
+   the button displays only the POS categories themselves.
+
+To fire a pending course, open the order from either the :ref:`Floor plan <pos/restaurant/floors>`
+or the :guilabel:`Orders` overview (select an order and click :guilabel:`Load Order`), then click
+the :guilabel:`Fire [Course]` button for the next sequence. Repeat this process as each course
+becomes ready for preparation.
+
+Transfer courses
+~~~~~~~~~~~~~~~~
+
+To move items to a different course, select either a single product or the entire course header in
+the cart. Click the :icon:`fa-ellipsis-v` (:guilabel:`Actions`) icon, select :icon:`fa-arrow-down`
+:guilabel:`Transfer course`, and choose the destination course.
 
 .. seealso::
    `Courses (video tutorial) <https://youtu.be/5W2S9HwSrDQ?si=Pauk9gyuypl4NU-M>`_


content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -335,6 +335,32 @@ When ready for the second course, retrieve the order from the :ref:`Floor plan
 .. seealso::
    `Courses (video tutorial) <https://youtu.be/5W2S9HwSrDQ?si=Pauk9gyuypl4NU-M>`_
 
+Automatic course allocation
+---------------------------
+
+The automatic course allocation feature creates courses and automatically sorts products by course.
```
mcsi-odoo: What do you mean by 'creates courses'? Becuase it seems like the actual creation of defined courses is not dependent on the feature being activated, right?

Maybe:

```suggestion
The automatic course allocation feature allows products added to an order to be automatically assigned to a default course.
```

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -335,6 +335,32 @@ When ready for the second course, retrieve the order from the :ref:`Floor plan
...
+To activate and configure the feature:
+
+#. Go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll down to the
+   :guilabel:`Preparation` section, and enable :guilabel:`Automatic course allocation`.
+#. Go to :menuselection:`Point of Sale --> Configuration --> POS Product Categories` and select a
+   product category.
```
mcsi-odoo: Could it help to give an example here? I know I'm new to POS, but I find it a bit abstract the way it is. Maybe something like:

```suggestion
#. Go to :menuselection:`Point of Sale --> Configuration --> POS Product Categories` and select a
   product category, e.g., Burgers.
```

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -335,6 +335,32 @@ When ready for the second course, retrieve the order from the :ref:`Floor plan
...
    `Courses (video tutorial) <https://youtu.be/5W2S9HwSrDQ?si=Pauk9gyuypl4NU-M>`_
 
+Automatic course allocation
+---------------------------
+
+The automatic course allocation feature creates courses and automatically sorts products by course.
+To activate and configure the feature:
+
+#. Go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll down to the
+   :guilabel:`Preparation` section, and enable :guilabel:`Automatic course allocation`.
```
mcsi-odoo: I'm wondering if it would be better to separate the 'activate' and 'configure'? Right now, both are bundled together, so the sentence that comes after the steps feels a bit odd.

```suggestion
To activate the feature, go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll down to the :guilabel:`Preparation` section, enable :guilabel:`Automatic course allocation`, then click :guilabel:`Save`.
```
Then, I would start the next part with something like:

```suggestion
   To configure courses and assign product categories to courses, follow these steps:.
```

Question though: is this considered the main way to configure courses? (i.e., not  to do it with debug mode activated, via Configuration > Courses)

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -335,6 +335,32 @@ When ready for the second course, retrieve the order from the :ref:`Floor plan
...
+#. Go to :menuselection:`Point of Sale --> Configuration --> POS Product Categories` and select a
+   product category.
+#. Enter a course in the :guilabel:`Course` field and click :guilabel:`Create`, or select an
+   existing course.
```
mcsi-odoo:
```suggestion
#. Enter a course in the :guilabel:`Course` field, e.g., :guilabel:`Main`, and click :guilabel:`Create`, or select an existing course.
```

Do we need to say something about repeating as many time as needed?

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -335,6 +335,32 @@ When ready for the second course, retrieve the order from the :ref:`Floor plan
...
+
+When you select any products from that POS category, they are automatically sorted into the
+defined courses.
```
mcsi-odoo:
```suggestion
Thereafter, when a product from that POS category is added to an order, it is automatically allocated to the defined course.
```

What happens if a customer wants to receive a product with a different course to the one it's automatically assigned to? Like Jane Doe prefers to have the shrimp croquettes (configured as starter) as her main? _Edited to add: just like when manually assigning courses. Which definitely confirms to me that we need to find a way to have a comprehensive Courses section that inclues manual and automatic options_

Also, what happens if there are product categories that do not have a course assigned? Just tried, and they seem to just get added to the previous course 'triggered' ... e.g., if you ordered a burger (Main) and a chocolate mousse (Dessert) and then another customer in the group orders a pizza (no course), then the pizza would end up in the Dessert course.

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -335,6 +335,32 @@ When ready for the second course, retrieve the order from the :ref:`Floor plan
...
+By default, courses are organized by creation sequence (the first created appears first). If you
+create a new course later but want it to appear before existing ones, you must manually change the
+sequence order.
+
+To rearrange the course order:
+
+#. Activate :ref:`developer mode <developer-mode/activation>`.
+#. Navigate to :menuselection:`Point of Sale --> Configuration --> Courses`.
+#. Rearrange the courses by dragging and dropping them into the desired order.
+
```
mcsi-odoo: Could this be better as tip? And if yes, condensed a bit? Something like:

```suggestion
By default, courses appear in the register in the order in which they were created. To rearrange the course order to reflect your service flow, activate :ref:`developer mode <developer-mode/activation>`, go to :menuselection:`Point of Sale --> Configuration --> Courses`, then use the :icon:`oi-draggable` :guilabel:`(draggable)` to move courses to the desired position.

```

mcsi-odoo (changes requested):
Hi @LoredanaLrpz , I'm happy it was lunchtime as I was finishing my review, 'cause all this talk of restaurants and courses was making me hungry 🤣 

My main comment at this stage would be (though I came to that conclusion a little after going into the details below): wouldn't it be better to slightly re-do the Courses section to first introduce the concept, and then explain the two ways a restaurant can work with courses? (The 'transferring a product between courses' is common to both anyway.)

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: I meant something like what you understood. It's just that instead of having to click the "Course" button to split an order into courses, it creates them in the cart and automatically sorts products. Without the feature, you have to click the button for courses to appear in the cart.

LoredanaLrpz: I'll try to be clearer

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: Configuration > Courses is only available in debug mode, unfortunately, so yes, it is the main way to configure courses (to assign a course to a pos product category).
I mention debug mode because it's the only way we could find to reorganize courses if the aperitif was created after the dessert and we need it first (for example).

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: I'm not sure it was necessary, but I've added it as a note along with something I originally didn't think about: the fact that one course can be set for multiple pos product categories, e.g., *Main* could be used for *Burgers*, *Pizzas*, and *Meat*.

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: That's right, they are added to the selected/triggered course. Drinks are a good example because one same drink could be ordered at any moment.
They can transfer any product from one course to another. I'll record a video: https://docs.google.com/videos/d/11b_q_GNegi7NnalW7jwpZ8ZWC5mgByJroYIgqtBm-6I/edit?usp=drive_link

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: It's something that made me quite mad when I couldn't find how to rearrange them last time, which is why I thought it could be important to mention outside of a tip, but it actually could be just a tip

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: Better yet, click "Main" to trigger it before selecting a pizza, but if the waiter forgot to trigger another course, he can transfer the product to the main course

LoredanaLrpz:
Hey @mcsi-odoo , I force-pushed a new version that is quite different from the first one. Following your suggestion, I changed the whole structure of the courses section. I don't know if that's what you had in mind? Let me know

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -314,23 +314,70 @@ the target table in the :ref:`Floor plan <pos/restaurant/floors>` view:
...
+   By default, courses appear in the cart in the order they were created. To rearrange the course
+   order to reflect your service flow, activate the :ref:`developer mode
+   <developer-mode/activation>`, go to :menuselection:`Point of Sale --> Configuration --> Courses`,
+   then use the :icon:`oi-draggable` :guilabel:`(draggable)` icon to move courses to the desired
+   position.
+
+Fire courses
+~~~~~~~~~~~~
+
+Once the order is complete, click :guilabel:`Send` to transmit it to the kitchen. The first course
```
mcsi-odoo: I was struggling to find the Send button, and then I realised that sometimes, the 'Send' doesn't appear:

Image


Is there a better way we could express this to avoid confusion?

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -314,23 +314,70 @@ the target table in the :ref:`Floor plan <pos/restaurant/floors>` view:
...
+appears on the :doc:`preparation display <extra/preparation>` as waiting to be prepared, while
+subsequent courses are marked as :guilabel:`Pending`.
```
mcsi-odoo: I think I need your help to understand this. For example, on this screenshot, I just sent orders for table 10 to the kitchen. I see the 'pending' for course 2, but I'm not seeing any kind of label on course 1 ... just the little clock icon:

Image


Is that what you mean by 'waiting to be prepared'? Or am I missing something?

mcsi-odoo (changes requested):
Hey @LoredanaLrpz , I think this looks great ... are you happy with it too?
I just need your help understanding the part about firing courses ... see my screenshot and questions 🙂

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: Exactly, the tag appears only when the order is on hold, not when it needs to be prepared. I'll try to make it clearer

content/applications/sales/point_of_sale/restaurant.rst
LoredanaLrpz: I could maybe add a note (?) I'll see what I can do :D

content/applications/sales/point_of_sale/restaurant.rst
```diff
@@ -314,23 +314,76 @@ the target table in the :ref:`Floor plan <pos/restaurant/floors>` view:
...
+   By default, courses appear in the cart in the order they were created. To rearrange the course
+   order to reflect your service flow, activate the :ref:`developer mode
+   <developer-mode/activation>`, go to :menuselection:`Point of Sale --> Configuration --> Courses`,
+   then use the :icon:`oi-draggable` :guilabel:`(draggable)` icon to move courses to the desired
+   position.
+
+Fire courses
+~~~~~~~~~~~~
+
+Once an order is complete, click :guilabel:`Send` to transmit it to the kitchen. The first course
```
mcsi-odoo: Since the _Send_ label can be hidden (and thank you for the note, that makes it super clear!!), maybe it would be good to also mention where it is? e.g.: 

```suggestion
Once an order is complete, click :guilabel:`Send` at the bottom left of the register to transmit it to the kitchen. The first course
```

mcsi-odoo (approved):
Thanks so much for those last changes, @LoredanaLrpz . It's super clear now. Once small suggestion below, but I leave that totally up to you, and I'll already delegate r+.

@robodoo delegate+

LoredanaLrpz:
@robodoo r+

PR #18420, opened 2026-06-12
[FIX] links to owl

djoewie:
All the links to the owl github are broken after the introduction of owl v3

making sure the links point back to the documentation that is now at a different location

will need to be ported to the newest version, and depending on the rules for backporting to older versions, this may also be needed

### DIFF
--- a/content/developer/reference/frontend/hooks.rst
+++ b/content/developer/reference/frontend/hooks.rst
@@ -357,5 +357,5 @@ API
      - Description
    * - `refName`
      - string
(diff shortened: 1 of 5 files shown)


AntoineVDV (approved):
Hi @djoewie, as you pointed out, this would indeed need to be backported to earlier versions (minimum 17.0) if the links are also broken there. It'd be easier to retarget this PR to 17.0 and forward-port it to master than to do it through multiple PRs.

djoewie:
ok, I will remake the  request on 17, and this can be ignored

PR #18460, opened 2026-06-15
[FIX] developer/*: links to owl v2

djoewie:
All the links to the owl github are broken after the introduction of owl v3

making sure the links point back to the documentation that is now at a different location

will NOT need to be ported forward, as there are more files to edit in newer versions
there is a separate pull request for those https://github.com/odoo/documentation/pull/18463

replacement of https://github.com/odoo/documentation/pull/18420

PR #18463, opened 2026-06-15
[FIX] developer/*: fix links to owl

djoewie:
All the links to the owl github are broken after the introduction of owl v3

making sure the links point back to the documentation that is now at a different location

will need to be ported to the newest version

replacement of https://github.com/odoo/documentation/pull/18420

contains all changes of https://github.com/odoo/documentation/pull/18460 + extra


AntoineVDV (approved):
@robodoo squash delegate+

djoewie:
@robodoo r+

PR #18504, opened 2026-06-17
[IMP] accounting: snailmail page limit

svfu-odoo:
Currently we only allow sending documents that are a maximum of 8 physical pages (regardless of the limits set by the snailmail provider / Pingen).

Every now and then we get tickets because of that.

To avoid confusion due to the page limits given in the "Pingen's layout requirements" linked below we add a line stating the current limits.

The limit has been in place for 4 years and is unlikely to be (just) increased since we charge the same number of credits per document while Pingen charges dependent on the length of the document.

opw-6200661


tohor-odoo (approved):
Hey @svfu-odoo! Thanks for the investigation, looks good to me :)

auva-odoo (commented):
@robodoo r+

PR #18505, opened 2026-06-17
[IMP] developer/backend: switch Warning to UserError

danych2:
Since https://github.com/odoo/odoo/pull/98138, Warning has been removed from the action's evaluation context, UserError is used instead


AntoineVDV (approved):
@robodoo r+

PR #18554, opened 2026-06-19
[IMP] developer: document populate write job domains

pivi-odoo:
Document the new top-level `domain` attribute for populate write jobs.

The update explains that write jobs can target records by `ref`, by `domain`, by both, or by neither, and clarifies that domains are evaluated once to select target records. It also notes that create jobs cannot define a top-level domain.

task-6303786

Related PR: https://github.com/odoo/odoo/pull/271112


kmagusiak (approved):

AntoineVDV (approved):
@robodoo r+ delegate+ delegate=kmagusiak

PR #18556, opened 2026-06-22
[IMP] howtos/website_themes: Bundles description

bram-odoo:
This commit improves the description of the different assets bundles and adds some context instead of just giving specific examples.

Task-6316683

### DIFF
--- a/content/developer/howtos/website_themes/theming.rst
+++ b/content/developer/howtos/website_themes/theming.rst
@@ -824,17 +824,21 @@ This is a non-exhaustive list of the frequently used bundles for a website:
    * - Bundle
      - Description
    * - web._assets_primary_variables
-     - Mainly used for the `primary_variables.scss` file
+     - Defines the primary SCSS variables (colors, fonts, spacing, etc) used as the main theming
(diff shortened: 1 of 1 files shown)


content/developer/howtos/website_themes/theming.rst
```diff
@@ -824,17 +824,21 @@ This is a non-exhaustive list of the frequently used bundles for a website:
    * - Bundle
      - Description
    * - web._assets_primary_variables
-     - Mainly used for the `primary_variables.scss` file
+     - Defines the primary SCSS variables (colors, fonts, spacing, etc) used as the main theming
+       layer across all bundles. Mainly used for the `primary_variables.scss` file.
```
AntoineVDV: Not 100% required, but you could use the `:file:` role for consistency in case we style it differently from regular highlighted terms.

AntoineVDV (approved):
@robodoo delegate+

content/developer/howtos/website_themes/theming.rst
bram-odoo: Good point, thanks for the input!

bram-odoo:
@robodoo r+

PR #18561, opened 2026-06-22
[FIX] legal: update license info for 19.0

niyasraphy:


odony (approved):
Thanks 👍 

@robodoo r+

(I guess no need to backport in older versions of the doc, it makes sense that they don't refer to the license of future versions)

PR #18564, opened 2026-06-22
[REM] accounting: remove tax scope

dade-odoo:
The Tax Scope (goods vs services) feature apparently does not have any effect. POs will discuss if the feature itself should change or be removed entirely.

To be merged up to and including master


auva-odoo (approved):
@robodoo r+

PR #18576, opened 2026-06-23
[FIX] sale_shopee: wrong synchronization status

dado-odoo:
The status of the orders that will be synchronized were incorrect.

### DIFF
--- a/content/applications/sales/sales/shopee_connector/manage.rst
+++ b/content/applications/sales/sales/shopee_connector/manage.rst
@@ -59,8 +59,11 @@ For every synchronized order, Odoo creates a sales order and a customer (contact
 customer hasn't been previously imported from Shopee or doesn't already exist in the database.
 
 .. note::
-   The principal of the synchronization is to *only fetch orders that needs to be shipped*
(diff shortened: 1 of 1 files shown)


lipiraux (approved):
Hello @dado-odoo ,

The content is correct for me 👍. thx for noticing and fixing it. As for the form, I'll leave it for the documentation team to decide and r+.

lipiraux:
@dado-odoo I just noticed it now, but you might have to recreate your branch and create a new PR. The runbot is not happy because it assumed the fix was targeting master because the branch name wasn't prefixed with any known target. The format is usually `<target>-<anything you want>-<trigram>`.

I think these commands should do the trick.
```bash
git switch dado-odoo-patch-1
git branch -m 19.0-fix-shopee-sync-status-dado
git push -u origin
```

You might also want to update your commit message to follow [Odoo guidelines](https://www.odoo.com/documentation/19.0/contributing/development/git_guidelines.html#commit-message-structure). For example: 

```
[FIX] sale_shopee: wrong synchronization status

The status of the orders that will be synchronized were incorrect.
```

I hope this doesn't make you too afraid to open a PR next time 😄

PR #18585, opened 2026-06-23
[FIX] Recruitment: Add required module

larm-odoo:
Adding information about a necessary module to show a specific button when offering a job.

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/6317855) for this PR.


larm-odoo:
Hi @Felicious - this is a 1-point fix, so sending right to you for approval. I have NOT done the master edit for this doc (new images, combing through), just adding the needed info so it is usable for the readers immediately.  I'll do an overhaul later.

Felicious (approved):
@robodoo r+

PR #18592, opened 2026-06-24
[IMP] essentials: sharing enriched view url

mcsi-odoo:
task-6310608

### DIFF
--- a/content/applications/essentials/search.rst
+++ b/content/applications/essentials/search.rst
@@ -238,3 +238,29 @@ the list of favorites.
 In the :guilabel:`Shared with` field, select the relevant users from the drop-down menu, then click
 :icon:`fa-cloud-upload` :guilabel:`(Save manually)`. The favorite is now visible for all selected
 users under :icon:`fa-star` :guilabel:`Favorites`.
+
+.. note::
+   Differences in access rights may mean that users of a shared search do not see the same results
+   as the user who created the favorited search.
+
+Share search results
+====================
+
+To copy a link to the current view that preserves any applied filters or groupings, open the command
+palette by pressing `CTRL + K` (or `Cmd ⌘ + K`), type `Share`, then select the :guilabel:`Share`
+command.
+
+The link is automatically copied to the clipboard and can be shared with internal users via the
+desired channel.
+
+.. tip::
+   To copy the link directly to the clipboard without first opening the control panel, press
+   `ALT + SHIFT + H` (or `CTRL + SHIFT + H`).
+
+.. note::
+   - When the link is opened, any original filters or groupings are automatically applied. Any
+     groupings appear in the search bar, while clicking the :icon:`fa-filter` :guilabel:`(filter)`
+     icon beside :guilabel:`Shared` in the search bar reveals the filters in the :guilabel:`Custom
+     Filter` window.
+   - Differences in access rights may mean that a user opening a shared link does not see the same
+     results as the user who created the link.


content/applications/essentials/search.rst
```diff
@@ -238,3 +238,33 @@ the list of favorites.
...
+Share search results
+====================
+
+To copy a link to the current view that preserves any applied filters or grouping, follow
+these steps:
+
+#. Open the command palette by pressing `CTRL + K` (or `Cmd ⌘ + K`).
+#. Type `Share`, then select it to copy the link to the clipboard.
+
+The link can then be shared internal users via the desired channel.
```
auva-odoo:
```suggestion
To copy a link to the current view that preserves any applied filters or grouping, open the command palette by pressing `CTRL + K` (or `Cmd ⌘ + K`), type `Share`, then select the :guilabel:`Share` command. The link is automatically copied to the clipboard and can be shared with internal users via the desired channel.
```
- "to copy a link" is already in the intro sentence, so it feels redundant in the instructions too.
- Dropping the ⬆️ trims things down a bit more, and even as-is, this is simple enough to just be text.
- Added guilabel + command since I first thought I needed to select the link, which threw me off (since it's not shown and copies automatically).
- Fixed a missing "with"
- Maybe the second sentence can go on an individual line for readability?

content/applications/essentials/search.rst
```diff
@@ -238,3 +238,33 @@ the list of favorites.
...
+
+.. tip::
+   To copy the link directly to the clipboard without first opening the control panel, press
+   `CTRL + ALT + H`.
+
+When the link is opened, any original filters or grouping are automatically applied. Any grouping
```
auva-odoo:
```suggestion
When the link is opened, any original filters or groupings are automatically applied. Any grouping
```
since grouping can be multi-level? (i.e., multiple groupings?)
and I think this should all be in a note/tip (I'd add the info to the existing one, as a bullet list item), and the screenshot is not necessary imo

content/applications/essentials/search.rst
```diff
@@ -238,3 +238,33 @@ the list of favorites.
...
+these steps:
+
+#. Open the command palette by pressing `CTRL + K` (or `Cmd ⌘ + K`).
+#. Type `Share`, then select it to copy the link to the clipboard.
+
+The link can then be shared internal users via the desired channel.
+
+.. tip::
+   To copy the link directly to the clipboard without first opening the control panel, press
+   `CTRL + ALT + H`.
```
auva-odoo: From what I remember (tested in 19.2 back when it was first released, maybe they changed/fixed it since), the shortcut changes based on the OS

auva-odoo (changes requested):

content/applications/essentials/search.rst
mcsi-odoo: You're right, much better! Thanks.

content/applications/essentials/search.rst
mcsi-odoo: You're totally right. And somehow I even got the Linux/Windows shortcut wrong 🙈 

It's `ALT + SHIFT + H` or `CTRL + SHIFT + H`

mcsi-odoo (commented):
Thanks for the review Audrey. 
Feedback integreated in the new version I'll push shortly.

auva-odoo (approved):
Thank you for taking my feedback into account @mcsi-odoo, and for updating this doc!

@robodoo r+

PR #18593, opened 2026-06-24
[IMP] howtos/website_themes: Bundles description + header/footer templates

bram-odoo:
This commits pursues #18556 and refines the description of the `website.website_builder_assets` bundle.

It also adds explicit information about the bundle within which the header/footer template option has to be loaded.

Task-6316683


AntoineVDV (approved):
@robodoo delegate+

bram-odoo:
@robodoo r+

PR #18636, opened 2026-06-26
[IMP] localization: specify use of branches in different fiscal loc

afma-odoo:
task-6129738


auva-odoo (changes requested):
Thank you @afma-odoo !

I think I'd move the note to the end of the Configuration section, it seems a bit odd to me where it is right now, especially since we don't mention branches at all in the paragraph above? Could we maybe briefly discuss this? 

And should we also maybe add something somwhere in the [branch doc](https://www.odoo.com/documentation/master/applications/general/companies.html#branches)?

auva-odoo (approved):
Thanks for the clarification @afma-odoo  and sorry I missed the Branches doc in the Accounting section 🙈 

@robodoo r+

PR #18647, opened 2026-06-26
[IMP] website/ecommerce: update tutorial links

masp-odoo:
Task-6320209


auva-odoo (approved):
Thank you for thinking of this @masp-odoo and for the thorough update! 🙂 

@robodoo r+

PR #18765, opened 2026-07-02
Quick POC sitemap.xml file using sphinx-sitemap

nikibrown:


requirements.txt
```diff
@@ -9,3 +9,4 @@ sphinxcontrib-htmlhelp==2.0.1
 sphinxcontrib-serializinghtml==1.1.5
 sphinxcontrib-qthelp==1.0.3
 sphinx-tabs==3.4.5  # Compatibility with docutils==0.17.0
+sphinx-sitemap==2.6.0
```
nikibrown, line 12: Hi @AntoineVDV @Felicious 

TLDR:
@AntoineVDV Can you take a look at Sphinx Sitemap and let me know if this will work for our needs? Or if there is a better alternative?

As part of an effort to improve SEO and AEO in the documentation we want to add a sitemap.xml file to the documentation website.  I did a quick proof of concept with `sphinx-sitemap` and my local baseurl and it looks like the file was generated correctly.

There are 1,137 RST files in the 19.0 branch and the generated sitemap has 1139 URLs (looks like theres an index and search that were added). 

Generated sitemap.xml file:
[https://gist.github.com/nikibrown/50a6668e5f46b61276b83959a4325e5c](https://gist.github.com/nikibrown/50a6668e5f46b61276b83959a4325e5c)

It looks like this supports versions and languages/locales
[https://sphinx-sitemap.readthedocs.io/en/latest/advanced-configuration.html#version-support](https://sphinx-sitemap.readthedocs.io/en/latest/advanced-configuration.html#version-support)

Let me know if there are other issues to consider or if I'm missing something. 

Thanks!

Feyensv:
Hello @nikibrown, thanks for this quick POC :wave:

It was already considered in the past and we concluded the only potential gains would be too costly to develop and handle: https://github.com/odoo/documentation/pull/9441#pullrequestreview-2452015708 

As far as I remember, `sphinx-sitemap` didn't correctly handle our custom sphinx configuration and it would have required further tinkering from the internal team to combine it with the existing sitemap of `odoo.com`. They also concluded after further investigation that it wouldn't effectively improve the SEO (without handling `lastmod` sitemap values, which isn't worth the work).

nikibrown:
@Feyensv  Thanks for the additional context. I was surprised we didn't have a sitemap file, probably should have searched for previous PR's :)

Re: combining with the main odoo.com sitemap - From what understand separate sitemaps are actually standard practice. So the sitemap could live here just for documentation: odoo.com/documentation/sitemap.xml. I don't think this is actually an issue.

RE: SEO - sitemaps without lastmod are still beneficial and standard practice. We are also trying to encourage LLMs to crawl our documentation to display correct version information and prevent hallucinations in answers. Sitemap.xml, llms.txt and schema data play a very important role in this.

Can revisit adding sitemap.xml with lastmod to the documentation repo?

PR #18779, opened 2026-07-06
[IMP] l10n_jo: change UBL format

afma-odoo:
task-6364972


auva-odoo (approved):
Thank you for taking care of this so quickly @afma-odoo! 🤗 

@robodoo r+

PR #18465, opened 2026-06-15
[IMP] accounting: restructure avalara doc and add avalara direct

dade-odoo:
taskid-6222754

Restructure Avalara documentation, including:
- rename AvaTax Integration page to Avalara Integration
- be more consistent in distinguishing Avalara (company/subscription/service/account) and AvaTax (engine/workflow)
- merge Avalara Included content into parent Avalara Integration page
- add Avalara Direct content on Avalara Integration page

PR #18856, opened 2026-07-08
[FIX] avalara: add redirections from old avalara pages

dade-odoo:
The [restructure of avalara PR](https://github.com/odoo/documentation/pull/18465) didn't include redirections. This PR adds them to avoid broken links.


auva-odoo (approved):
@dade-odoo I changed the last one as it didn't use the old file name and therefore didn't work 😉 

@robodoo r+

PR #18857, opened 2026-07-08
[ADD] attendances: document reminder feature

ksin-odoo:
Add documentation for the attendance and timesheet reminder system. The reminder popup appears after 15 minutes when an employee is scheduled to work but has not checked in or started a session.

Covers:
- Check-in reminder (attendance-based tracking)
- Session reminder (timesheet-based tracking)
- Merged icon behavior when both apps are installed
- Configuration steps for Attendance Based option
- Timing, working schedule, cross-tab, and state reset behavior

task-6171500


ksin-odoo:
PR is not required

PR #18860, opened 2026-07-08
[IMP] web: drop date filter range attributes

jeanschoenlaub:
[IMP] developer/view_architectures: drop date filter range attributes

The `start_month`, `end_month`, `start_year` and `end_year` attributes
let a `date` filter customize the range of month/year period options
shown in the search menu.

With the new relative date filters and their prev/next navigation, this
per-view customization is no longer needed: the period option range is
now fixed and users navigate relatively instead.

Remove the documentation for these attributes and update the
`default_period` valid values to reflect the fixed range.

see odoo/odoo#267980, odoo/enterprise#122986

task-6260555

### DIFF
--- a/content/developer/reference/user_interface/view_architectures.rst
+++ b/content/developer/reference/user_interface/view_architectures.rst
@@ -1976,74 +1976,6 @@ The `filter` element can have the following attributes:
    :type: str
    :default: `''`
 
-.. attribute:: start_month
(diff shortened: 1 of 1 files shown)


jeanschoenlaub:
Wrong PR, see: https://github.com/odoo/documentation/pull/18866

PR #18866, opened 2026-07-08
[IMP] developer/view_architectures: drop date filter range attributes

jeanschoenlaub:
The `start_month`, `end_month`, `start_year` and `end_year` attributes let a `date` filter customize the range of month/year period options shown in the search menu.

With the new relative date filters and their prev/next navigation, this per-view customization is no longer needed: the period option range is now fixed and users navigate relatively instead.

Remove the documentation for these attributes and update the `default_period` valid values to reflect the fixed range.

see odoo/odoo#267980, odoo/enterprise#122986

task-6260555

### DIFF
--- a/content/developer/reference/user_interface/view_architectures.rst
+++ b/content/developer/reference/user_interface/view_architectures.rst
@@ -2041,74 +2041,6 @@ The `filter` element can have the following attributes:
    :type: str
    :default: `''`
 
-.. attribute:: start_month
-   :noindex:
-
-   The earliest month that will show up in the dropdown of a date filter, as an offset relative to the current month.
-
-   .. example::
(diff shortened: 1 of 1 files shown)


AntoineVDV (approved):
@robodoo squash delegate+ delegate=aab-odoo

AntoineVDV:
Hi @jeanschoenlaub, I reverted your PR title change because I already set the merge method to `squash`, which means that the mergebot will use the PR title and description as the commit title and body. Since `web` doesn't refer to anything in the documentation repo (commits should adapt to the repo they'll land in), I kept `view_architectures`.

aab-odoo (approved):
robodoo r+

PR #18884, opened 2026-07-09
[FIX] *: make cheat sheet examples translatable

dylankiss:
Currently the accounting interactive cheat sheet examples are constructed in JavaScript, including the English strings used in them. This results in the examples not being translatable, which is a problem for non-English users.

In order to make these examples translatable, we created a new extension called ``odoo_dynamic_examples``. This extension allows to define the examples in HTML, and to use the `_()` function to mark strings for translation. An accompanying JavaScript file handles the dynamic behavior of the examples. The examples are then rendered in the documentation using the ``odoo-dynamic-example`` directive.

Examples can contain HTML, JS, or both. The HTML and JS files are linked by their name, which is the name of the example. For example, the ``chart_of_accounts`` example is defined in ``chart_of_accounts.html`` and ``chart_of_accounts.js``.

Additionally, we found that the dynamic JS code used React, which was probably a bit overkill for what we needed. We replaced the React code with vanilla JS code and got rid of the dependencies on React and Immutable.

Finally, we also removed inventory-related JS files that didn't seem used anymore.

[task-5173959](https://www.odoo.com/odoo/project.task/5173959)


AntoineVDV (commented):
Hello there, nice work!

Let's be honest, we (doc-review) are not going to "review" accounting-specific JS/HTML mementos, but we can at least comment on a few things:
- Using a dedicated Sphinx extension sounds like a good approach.
- It's great to finally get rid of React in the documentation.
- The balance sheet example seems to work, but other examples appear to be currently broken in your branch.

So let's conclude with a traditional "LGTM" and trust you to fix what must be before, *and after* the merge 😄 

@robodoo delegate+

dylankiss:
@AntoineVDV Thanks for the feedback! When testing it locally, it worked in the end, but I forgot to test with a clean slate. 🙈 It was a matter of the final JS files not being found in this case. Should be fixed now. I'll test again when the new build is finished.

dylankiss:
It works, but I'm still checking a slightly more flexible approach.

PR #18917, opened 2026-07-13
[IMP] Purchase: Reordering rules example block

jero-odoo:
Removing example admonition block based on customer feedback
Docs task : https://www.odoo.com/odoo/project.task/4987026


jero-odoo:
@erjer-odoo quick edit, sending right to you!

erjer-odoo (approved):
@robodoo r+

PR #18926, opened 2026-07-14
[REF] developer/../view_architecture: factorize popover doc

aab-odoo:
Task~6385983


AntoineVDV (approved):
LGTM 👍 

I checked whether using the `include` directive would help reduce the number of refs. It does, but it's not clear that it's an improvement either... It's typically used to render single-attribute pieces of doc, while here, popovers are documented in an entire section, which doesn't cut-and-paste as well.

So let's go with your version. The important thing is to have the information written somewhere. We can always restructure this page and others in the future if needed.

@robodoo delegate+

juliusc2066:
@robodoo rebase-ff delegate+ r+

juliusc2066:
@robodoo r+

aab-odoo:
robodoo r+

PR #18936, opened 2026-07-15
[IMP] pos/ingenico: update Ingenico documentation 19.0

emmi-odoo:
task-5504275


content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
 .. important::
    - Ingenico payment terminals require an :doc:`IoT system </applications/general/iot>`.
    - Ingenico is currently only available in Belgium, the Netherlands and Luxembourg.
-   - Odoo works with the Ingenico Lane/, Desk/, and Move/ payment terminals as they support the TLV
+   - Odoo works with the Ingenico Lane/Desk/Move 5000 payment terminals, as they support the TLV
```
mcsi-odoo: Would it be better to stick to the actual device names Ingenico uses on their website, rather than try to condense like this, i.e.,  

```suggestion
   - Odoo works with Ingenico's Lane/5000, Desk/5000, and Move/5000 payment terminals, as they support the TLV
```

- https://ingenico.com/en/products-services/payment-terminals/tetra/lane5000-le
- https://ingenico.com/en/products-services/payment-terminals/tetra/desk5000
- https://ingenico.com/en/products-services/payment-terminals/tetra/move5000

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
      communication protocol through TCP/IP.
 
-Configuration
-=============
+.. _pos/ingenico/terminal-configuration:
 
-Connect an IoT system
----------------------
+Lane/Desk/Move 5000 terminal configuration
```
mcsi-odoo: To stay consistent with other payment terminal doc, shouldn't this just be:

```suggestion
Ingenico configuration
```

After all, you already mentioned in the 'Important' note above which terminals are supported. Plus, this formulation isn't very future-proof if, for example, more devices are supported in the future.

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
 
-Configure the Lane/Desk/Move 5000 terminals for Ingenico BENELUX
-----------------------------------------------------------------
+#. Go to :menuselection:`Point of Sale --> Configuration --> Settings`.
+#. Scroll down to the :guilabel:`Payment Terminals` section, enable the :guilabel:`Ingenico
+   (BENELUX)` payment terminal.
+#. Click :guilabel:`Save`.
+
+Then :doc:`connect the IoT system to Odoo </applications/general/iot/connect>` and follow
+these steps on the terminal to configure the Ingenico Lane/Desk/Move 5000 terminal:
```
mcsi-odoo: I find the way this is done in the Worldline doc works well ... is there a reason you're not using the same approach here? i.e.,

```suggestion
First, enable the Ingenico payment terminal in the :ref:`POS settings <pos/use/settings>` under :guilabel:`Payment Terminals`. Then :doc:`connect the IoT system to Odoo </applications/general/iot/connect>` and follow these steps on the terminal:
```

Also, if you do feel the need to refer to the supported models, then I think it would be better to use, e.g., 

"... configure the Lane/5000, Desk/5000, or Move/5000 terminal:

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
 
 #. Press the function button (:guilabel:`F` on Lane/5000, :guilabel:`⦿` on Desk/5000 and
    Move/5000).
```
mcsi-odoo: [I know you didn't change it, but ... a suggestion]

```suggestion
#. Press the function button (:guilabel:`F` on the Lane/5000 terminal, :guilabel:`⦿` on the Desk/5000 and
Move/5000 terminals).
```

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
-#. Select :guilabel:`Change Connection` and press :guilabel:`OK` on the next screen.
-#. Select :guilabel:`TCP/IP` and :guilabel:`IP-address`.
-#. On the next screen, enter the IP address of your IoT system.
-#. Enter `9000` as port number and press :guilabel:`OK` on the next screen.
-
-At this point, the terminal restarts and should be displayed on the IoT system's form in Odoo.
+#. Go to :menuselection:`Kassa Menu --> Settings Menu`, enter the settings password (default:
+   `2009`), and press :guilabel:`OK`.
+#. Select :guilabel:`Protocol` and press :guilabel:`OK`.
+#. Select `CTEP` and press :guilabel:`OK`.
```
mcsi-odoo: Should `CTEP` really be literal markup if it's something the user is seeing and selecting on the screen, rather than entering themselves?

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
+#. Select :guilabel:`Change Connection` and press :guilabel:`OK`.
+#. Select :guilabel:`TCP/IP` and press :guilabel:`OK`. Then, select :guilabel:`IP-address` and press
+   :guilabel:`OK`.
+#. Enter the IoT's IP address and press :guilabel:`OK`.
```
mcsi-odoo:
```suggestion
#. Enter the IoT system's IP address and press :guilabel:`OK`.
```

I guess we're always talking about an IoT box's/system's XYZ ... right? And actually, which is preferred ... IoT box or IoT system? (I think I've seen both across the doc.)

_Edited to add:_ OK, so I guess we use 'IoT system' to cover either the Odoo IoT box or the Windows virtual IoT. So here, IoT system would be most appropriate, right?

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
+#. Enter port number `9001` (or `9050` if using Windows IoT).
```
mcsi-odoo: Is it much more likely a user will be using an Odoo IoT box rather than the Windows virtual IoT? If not, maybe we should phrase this more like:

```suggestion
#. Enter port number `9001` if using an Odoo IoT box or `9050` if using the Windows virtual IoT.
```

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
```diff
@@ -8,46 +8,54 @@ the work of your cashiers.
...
+#. Select :guilabel:`No SSL` and press :guilabel:`OK`.
+
+The terminal restarts and should be displayed on the IoT system's form in Odoo. Continue the
+configuration and create a :ref:`payment method <pos/ingenico/payment-method>`.
```
mcsi-odoo: If by 'Continue the configuration ...' you just mean 'Do the Odoo configuration part that's explained below', then I think you should remove it, as it's confusing to see it here. It's already clear enough that you have the Ingenico configuration (especially when you rename it) and then Odoo configuration sections.

mcsi-odoo (changes requested):
Hi Emmanuël, just a few comments here ... happy to discuss any. Also, I took a look at the task and am a bit confused by the following:

Image

Maybe I'm misunderstanding, but now that these terminals are being configured 'in Worldline mode', doesn't that require selecting Worldline in the POS settings and when configuring the payment method? (Just going by what I'm seeing in the task.)

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
emmi-odoo: Right on! Thanks ;)

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
emmi-odoo: The thing is, it got brought down to these devices only, but there were more before. I agree with you, though; it's been updated in PyCharm.

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
emmi-odoo: The reason is: we wondered if saying "go to the POS settings" was still appropriate, while we could just write the menu selection, and naturally, I started turning it into steps, but a paragraph could work too.

content/applications/sales/point_of_sale/payment_methods/terminals/ingenico.rst
emmi-odoo: Suggesting a guilabel instead?

mcsi-odoo: Yep, indeed, I meant to use a guilabel instead.

mcsi-odoo (approved):
As discussed ... 

@robodoo delegate+

emmi-odoo:
@robodoo r+

PR #18939, opened 2026-07-15
[IMP] pos/mercado-pago: add important banner to Mercado Pago configur…

emmi-odoo:
task-6373598

### DIFF
--- a/content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
+++ b/content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
@@ -11,8 +11,9 @@ customer transactions.
      <https://www.mercadopago.com.ar/herramientas-para-vender/lectores-point>`_, `Brazil
      <https://www.mercadopago.com.br/ferramentas-para-vender/maquininhas-point>`_, and `Mexico
      <https://www.mercadopago.com.mx/herramientas-para-vender/lectores-point>`_.
-   - Mercado Pago payment terminals do not require an :doc:`IoT Box </applications/general/iot>` to
-     operate.
+   - Mercado Pago payment terminals do not require :doc:`an IoT system </applications/general/iot>`
+     to operate.
+   - Odoo does not support the :guilabel:`STANDALONE` operation mode.
 
 .. seealso::
    - :doc:`Mercado Pago online payments </applications/finance/payment_providers/mercado_pago>`
@@ -94,12 +95,6 @@ There are three credentials to generate and collect from the `Mercado Pago Devel
 #. Locate the terminal serial number on the back of the terminal and write it down for the
    :ref:`pos/mercado-pago/payment-method` step.
 
-Once the Mercado Pago configuration is complete, log in to your Odoo database to create a
-:ref:`payment method <pos/mercado-pago/payment-method>` using the collected credentials.
-
-.. warning::
-   Odoo does not support the :guilabel:`Standalone` operation mode.
-
 .. note::
    All terminals purchased with your Mercado Pago account are automatically displayed on the
    Mercado Pago dashboard.
@@ -108,6 +103,11 @@ Once the Mercado Pago configuration is complete, log in to your Odoo database to
 
 Odoo POS configuration
 ======================
+
+.. important::
+   For the Mercado Pago terminal to work correctly with Odoo, make sure to :ref:`enable Mercado Pago
+   as a payment provider <payment_providers/mercado_pago/odoo-configuration>` and add the associated
+   account credentials.
 
 #. Go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll down to the
    :guilabel:`Payment Terminals` section, enable :guilabel:`Mercado Pago`, and click
--- a/content/applications/finance/payment_providers/mercado_pago.rst
+++ b/content/applications/finance/payment_providers/mercado_pago.rst
@@ -31,6 +31,8 @@ Configuration on Mercado Pago dashboard
    The testing environment is not supported; all transactions are generated in the production
    environment.
 
+.. _payment_providers/mercado_pago/odoo-configuration:
+
 Odoo configuration
 ==================
 


content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
```diff
@@ -11,8 +11,7 @@ customer transactions.
      <https://www.mercadopago.com.ar/herramientas-para-vender/lectores-point>`_, `Brazil
      <https://www.mercadopago.com.br/ferramentas-para-vender/maquininhas-point>`_, and `Mexico
      <https://www.mercadopago.com.mx/herramientas-para-vender/lectores-point>`_.
-   - Mercado Pago payment terminals do not require an :doc:`IoT Box </applications/general/iot>` to
-     operate.
+   - Mercado Pago payment terminals do not require an IoT Box to operate.
```
mcsi-odoo: Should we be saying that these payment terminals "do not require an IoT _system_ to operate"? 

And (so I understand) you're removing the link because you figure they don't need to be able to access that info anyway, since it's not needed?

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
```diff
@@ -28,6 +27,11 @@ to your payment terminal and store <pos/mercado-pago/terminal>`. Then, create a
 application <pos/mercado-pago/point-smart>` for technical setup and acquiring :ref:`credentials
 <pos/mercado-pago/credentials>` to complete the integration.
 
+.. important::
+   For the Mercado Pago terminal to work correctly, make sure to enable :ref:`Mercado Pago
+   <payment_providers/mercado_pago/odoo-configuration>` as a :ref:`payment provider
+   <payment_providers/add_new>` and add the associated account credentials.
```
mcsi-odoo: Isn't it overkill, and confusing, to add both these links here? Can't we just do like this:

```suggestion
.. important::
   For the Mercado Pago terminal to work correctly, make sure to :ref:`enable Mercado Pago as a payment provider <payment_providers/mercado_pago/odoo-configuration>` and add the associated
   account credentials.
```

But ... a bigger question I have is why are we adding this info in the Mercado Pago configuration section? Shouldn't this count as part of the Odoo configuration? (And is this 'enabling it as an online payment provider' only relevant for Mercado or also for other terminal setups? I'm feeling a bit confused about all this 🙃 )

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
```diff
@@ -73,6 +77,9 @@ Point Smart application
 Credentials generation
 ----------------------
 
+.. warning::
+   Odoo does not support the :guilabel:`Standalone` operation mode.
```
mcsi-odoo: I notice you moved this from its former position at the end of the Mercado Pago configuration section. Now it feels very much part of the Credentials section ... but is that the best place for it? Could it make more sense just to have it in the first 'Important' note at the top of the page?

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
```diff
@@ -97,9 +104,6 @@ There are three credentials to generate and collect from the `Mercado Pago Devel
 Once the Mercado Pago configuration is complete, log in to your Odoo database to create a
 :ref:`payment method <pos/mercado-pago/payment-method>` using the collected credentials.
```
mcsi-odoo: [I know you didn't change this, so, feel free to ignore ...]

I made a comment on your Ingenico PR questioning whether it was a good idea to have a sentence like this at the end of the [insert provider name] configuration section. Also, here, it's not just about creating the payment method, it's also enabling the terminal, right? (And maybe even the part about enabling the payment provider in this case?) So I would either make it more generic, or not have such a sentence at all.

FWIW, I don't see a sentence like this on the recently updated Adyen page, or on the Worldline page, for example.

mcsi-odoo (changes requested):
Hey Emmanuël, a few comments/questions here, but mainly I'm a bit confused about _where_ the new important banner is being added. Maybe we can discuss so I can understand the reasoning better?

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
emmi-odoo: Yes, we discussed that with Loredana a few weeks ago. I don't find it relevant to have a ref if it's not required, but at the same time, a ref could be useful as a form of information. What do you think?

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
emmi-odoo: True!

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
emmi-odoo: You're right! This poor sentence has been tossed around enough

mcsi-odoo: Hopefully this will be its final resting place :headstone: :rofl:

content/applications/sales/point_of_sale/payment_methods/terminals/mercado_pago.rst
mcsi-odoo: Hmm, I think my gut instinct would be to keep the ref. Maybe for some, it would be enough to know that it's not needed, but maybe others would like to understand better.

mcsi-odoo (approved):
Thanks for the updates, @emmi-odoo !
I let you decide whether or not to keep that ref to the IoT system page (but I would 😉 ).

@robodoo delegate+

mcsi-odoo:
Oh, forgot to say, you have a little extra something here: 

image

:green_salad: :tooth:

emmi-odoo:
@robodoo r+

PR #18952, opened 2026-07-16
[IMP] Payroll - Update US payroll localization

larm-odoo:
Added more states, retroactively, on [this task card](https://github.com/odoo/project/1251/tasks/5963020).

[Task card](https://www.odoo.com/odoo/project/3835/tasks/6396833) for this PR.

### DIFF
--- a/content/applications/hr/payroll/payroll_localizations/united_states.rst
+++ b/content/applications/hr/payroll/payroll_localizations/united_states.rst
@@ -7,22 +7,51 @@ employer payroll taxes. It accounts for federal and state regulations.
 
 The states included in this localization are:
 
-- California (CA)
-- New York (NY)
-- Texas (TX)
-- Florida (FL)
-- Colorado (CO)
-- Alabama (AL)
-- Washington (WA)
-- Nevada (NV)
-- Vermont (VT)
-- Illinois (IL)
-- Virginia (VA)
-- Washington, D.C (DC)
-- Arizona (AZ)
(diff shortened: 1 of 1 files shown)


larm-odoo:
Hi @erjer-odoo - this is a tiny 1-point PR, ready for your final approval!

content/applications/hr/payroll/payroll_localizations/united_states.rst
```diff
@@ -7,22 +7,51 @@ employer payroll taxes. It accounts for federal and state regulations.
...
+   * - Texas
+     - (TX)
+   * - Vermont
+     - (VT)
+   * - Virginia
+     - (VA)
+   * - Washington
+     - (WA)
+   * - Washington, D.C.
+     - (DC)
```
erjer-odoo:
```suggestion
     - \(DC)
```

So this is a weird case... when this is built, it's interpreted as a Roman numeral, so it formats without parentheses and adds a period at the end. Therefore, the (DC) sequence needs to be escaped.

The same will be true of Michigan (MI) and Maryland (MD), so that's just something to keep an eye out for in the future, should this table need to update.

erjer-odoo (approved):
LGTM -- just one minor issue where the build looks different than you'd expect, so you need to escape that sequence. Otherwise good to go!

@robodoo delegate+

larm-odoo:
@robodoo r+

PR #18954, opened 2026-07-17
[IMP] ecommerce: add new Products features

lman-odoo:
task-6229038
task-6229018

### DIFF
--- a/content/applications/websites/ecommerce/configuration/products.rst
+++ b/content/applications/websites/ecommerce/configuration/products.rst
@@ -18,24 +18,27 @@ Product creation
 ================
 
 To create a product from the frontend, go to the Website app, click :guilabel:`New`
-:icon:`fa-caret-down` in the top-right corner, then select :guilabel:`Product`. Enter the
-:guilabel:`Product Name`, and other relevant details such as the :guilabel:`Barcode`,
-:guilabel:`Sales Price`, default :guilabel:`Sales Taxes`, and a :ref:`Website Category
-<ecommerce/categories_variants/categories>`, then add an image. Once you click :guilabel:`Save`, you
-are redirected to the product page where you can :ref:`customize it
-<ecommerce_design/product_page/customization>` using the :doc:`website editor
+:icon:`fa-caret-down` in the top-right corner, then select :guilabel:`Product`. Enter the product
+name and other relevant details, such as the :guilabel:`Barcode`, :guilabel:`Sales Price`, default
+:guilabel:`Sales Taxes`, and a :ref:`Website Category <ecommerce/categories_variants/categories>`,
+then add an image. Once you click :guilabel:`Save`, you are redirected to the product page where you
+can :ref:`customize it <ecommerce_design/product_page/customization>` using the :doc:`website editor
 </applications/websites/website/web_design>`. Once you are finished, click :guilabel:`Save`. To
 configure additional advanced :ref:`product details <ecommerce/products/product-configuration>`,
 click :guilabel:`Product` in the top-right corner to access the backend product form.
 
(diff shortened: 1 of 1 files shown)


content/applications/websites/ecommerce/configuration/products.rst
```diff
@@ -199,6 +203,12 @@ corresponding :guilabel:`Optional Products`, :guilabel:`Accessory Products`, and
          <../../website/web_design/elements>` if necessary. Go to the :guilabel:`Style` tab, scroll
          to the :guilabel:`Alternative Products` section, and modify the options as needed.
 
+.. tip::
+   To automatically suggest cross-sell and upsell products, go to :menuselection:`Website -->
+   Configuration --> Settings`, scroll down to the :guilabel:`eCommerce` section, and enable
+   :guilabel:`Automate suggested products`. Once enabled, Odoo runs a recurring automated action.
+   Since suggestions are based on :ref:`e-commerce categories
+   <ecommerce/categories_variants/categories>`, ensure each product has a category assigned.
```
auva-odoo: - I'm not sure I tested this right but it looks like these aren't just *suggestions*, are they? like it looks like the scheduled action will fill the fields automatically and then voilà, saved and done. The user doesn't even have to review them?
- It think it makes more sense to refer the products types by their "usual" name (optional, accessory, alternative)?
- I'd like to discuss the scheduled action part; I suggested to remove that part but it seems the action is only run once a week, meaning if the user changes the configuration (e.g. to assign product categories or something), they'll either have to run the scheduled action manually or wait a week. Is that correct? If so, should we mention it?
- 
(The suggestion below has some issues + doesn't take the scheduled action into account, I'm just writing it down so we can discuss this)

```suggestion
   To automatically set optional, accessory, and alternative products, go to :menuselection:`Website -->
   Configuration --> Settings`, scroll down to the :guilabel:`eCommerce` section, and enable
   :guilabel:`Automate suggested products`. The products are added based on the :ref:`e-commerce category <ecommerce/categories_variants/categories>` assigned to the main product.
```
- not sure about *set*, do you have any other ideas?
- really not convinced with "main product"

content/applications/websites/ecommerce/configuration/products.rst
```diff
@@ -344,8 +354,12 @@ in the :guilabel:`Packagings` field. The available package types are displayed o
...
-   :guilabel:`Sales`, add as many package types as needed in the :guilabel:`Packagings` field.
+   - It is also possible to add packagings to a specific :ref:`product variant
+     <ecommerce/categories_variants/product-variants>`. To do so, go to the product form, click the
+     :icon:`fa-sitemap` :guilabel:`Variants` :ref:`smart button
+     <products/variants/variants-smart-button>`, and select the relevant product variant. Under
+     :guilabel:`Sales`, add as many package types as needed in the :guilabel:`Packagings` field.
+   - To prevent selected packaging units from being sold on the e-commerce, go to
+     :menuselection:`Website --> Configuration --> Settings`, and scroll down to the
+     :guilabel:`eCommerce` section. Then, under :guilabel:`Restrict Packagings`, select the units
+     that should not be available for sale.
```
auva-odoo, line 365: Do you know the purpose of this feature? I'm not sure I get, because this will make all products with this unit unavailable, correct?

auva-odoo (changes requested):
@lman-odoo thank you for the update! I spoke a bit too soon earlier, sorry, I do have a couple of questions 🙈 It will probably be easier to just discuss them IRL I think 🙂. Let me know when you're available!

auva-odoo (approved):
Thank you @lman-odoo! 🙂 

@robodoo r+

PR #19033, opened 2026-07-22
[IMP] Sales: Product Variants clarifications

theRealThagomizer:
This is a very minor PR that clarifies a line about where users can find Product Variants within the Sales app. It's closer to a fix than to an improvement, but the previous language wasn't technically wrong.

This 18.0 PR can be FWP up to saas-19.3.


theRealThagomizer:
@robodoo up to saas-19.3

erjer-odoo (approved):
@robodoo r+

PR #19045, opened 2026-07-22
[FIX] reference/user_interface: typo in xpath replace position example

jacw-odoo:
The `xpath` node has a feature in the `replace` position which copies copies the node matched by the xpath expression. This feature looks for an element with text exactly equal to `$0`; newlines, whitespace characters, and other nodes will prevent the characters from being recognized.

Thus, the example code for this behavior is incorrect, as it includes several newlines and spaces. This commit fixes the example code.

task-6408040 [Link](https://www.odoo.com/odoo/my-tasks/6408040)

### DIFF
--- a/content/developer/reference/user_interface/view_records.rst
+++ b/content/developer/reference/user_interface/view_records.rst
@@ -304,9 +304,7 @@ specifies how the matched node should be modified.
       .. code-block:: xml
 
          <xpath expr="//field[@name='x_field']" position="replace">
-             <div class="wrapper">
-                 $0
-             </div>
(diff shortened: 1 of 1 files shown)


jacw-odoo:
ERROR	Limit reached: 05115493a386d37aae01daee2619a62cbf7c1031 has more than 50 commit (1638) and will be skipped. Contact runbot team to increase your limit if it was intended.

Ahh, I think I see the problem. I need to rename this branch to have the 17.0 prefix. The github overview shows that there is only one commit, but the runbot builds probably guess from context clues like the branch name.

PR #19074, opened 2026-07-23
[FIX] developer/tutorials: remove remark for present fields in views

YassinWalid:
After odoo/odoo#137031, the views no longer require the fields used in them to be present in the view, as they are automatically added to the view by the framework. This commit removes the outdated remark in the tutorial that suggested otherwise.


AntoineVDV (approved):
@robodoo delegate+

Note that it's recommended to reference commits, and not PRs, in your commit message (in the PR description it's fine). That is because each of your or the mergebot's pushes will log a mention on the target PR, which quickly becomes annoying spam.

SaddemAmine (approved):

YassinWalid:
@robodoo r+

PR #19160, opened 2026-07-29
[IMP] MRP: clarify Apply on Variants matching logic

NeusAap:
The documentation explains that BoM lines can be assigned to multiple attribute values, but does not explain how those values are combined.

Clarify that values belonging to the same attribute are evaluated as OR conditions, while values belonging to different attributes are evaluated as AND conditions. Also clarify that leaving the field empty applies the line to every product variant.

This matches the BoM filtering logic implemented in _skip_for_no_variant (https://github.com/odoo/odoo/blob/19.0/addons/mrp/models/mrp_bom.py#L593-L630), which delegates instant and dynamic variant matching to _match_all_variant_values (from the product model).

I've added an example using the same attributes & colors as given in the supplied screenshots.

### DIFF
--- a/content/applications/inventory_and_mrp/manufacturing/advanced_configuration/product_variants.rst
+++ b/content/applications/inventory_and_mrp/manufacturing/advanced_configuration/product_variants.rst
@@ -109,8 +109,26 @@ columns. Then, choose the desired values in the :guilabel:`Apply on Variants` co
    :align: center
    :alt: "Apply on Variants" option on the additional options menu.
 
-Each component can be assigned to multiple variants. Components with no variants specified are used
-in every variant of the product. The same principle applies when configuring operations and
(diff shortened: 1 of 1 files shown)


stepperpig (approved):
Hello, thanks for the contribution! Tagging @erjer-odoo for a final check.

erjer-odoo (approved):
Hi @NeusAap ! Thank you for your contribution!

We're passing this along to merge into our documentation.

@robodoo r+

NeusAap:
No worries, glad I could help. Hopefully it saves the next person reading it, an hour or two ;)

erjer-odoo:
@robodoo r+

(Didn't get a reaction from the bot last time; seeing if this will spur it to do something.)

PR #19193, opened 2026-07-30
[FIX] developer/tutorials/discover_js_framework: fix typo

SaddemAmine:
No task ID


jeanbfly (approved):
I feel like I'm really making a difference 😆

SaddemAmine:
You are! Good catch!

AntoineVDV (approved):
Hello there, thanks for the fix 🙂 

Note that the 16.0 version is no longer built, but the forward-port to master will reach the maintained versions.

@robodoo r+ squash

PR #19194, opened 2026-07-30
[IMP] pos: add alternative Invoice method on feedback screen

emmi-odoo:
task-6164196


auva-odoo (approved):
Thanks @emmi-odoo!

@robodoo r+

PR #19213, opened 2026-07-31
[FIX] fiscal_localizations: correct UAE typo in localization docs

arshadzaynuddin:
The chart of accounts description read "UEA" instead of "UAE". This corrects the country abbreviation in the United Arab Emirates fiscal localization documentation.


dade-odoo (approved):
Thanks for the fix @arshadzaynuddin

dade-odoo:
@robodoo r+

PR #19308, opened 2026-08-05
[IMP] domain_names: CAA record

mege-odoo:
Occasionally customers use a CAA record to restrict which certificate authorities are allowed to provide certificates for their domain. If Let's Encrypt is not included in that, it will cause the certificate generation by Odoo Online or SH to fail.

Since this is an advanced topic, I opted to put a note about it in the section that gives more details about the SSL implementation. I expect that putting it in the main "Configure an existing domain name" section would cause confusion for most customers.

### DIFF
--- a/content/applications/websites/website/configuration/domain_names.rst
+++ b/content/applications/websites/website/configuration/domain_names.rst
@@ -362,7 +362,8 @@ protocol.
 
 Odoo generates a separate SSL certificate for each domain :ref:`mapped to a database
 <domain-name/db-map>` using `Let's Encrypt's certificate authority and ACME protocol
-<https://letsencrypt.org/how-it-works/>`_.
+<https://letsencrypt.org/how-it-works/>`_. Any CAA records configured for the domain must `allow
+Let's Encrypt <https://letsencrypt.org/docs/caa/>`_, otherwise certificate generation may fail.
 
 .. note::
    - Certificate generation may take up to 24 hours.


content/applications/websites/website/configuration/domain_names.rst
```diff
@@ -362,7 +362,9 @@ protocol.
 
 Odoo generates a separate SSL certificate for each domain :ref:`mapped to a database
 <domain-name/db-map>` using `Let's Encrypt's certificate authority and ACME protocol
-<https://letsencrypt.org/how-it-works/>`_.
+<https://letsencrypt.org/how-it-works/>`_. If you have a CAA record configured for your domain, make sure it `allows
+Let's Encrypt <https://letsencrypt.org/docs/caa/>`_, otherwise the certificate can fail to
+generate.
```
auva-odoo:
```suggestion
<https://letsencrypt.org/how-it-works/>`_. Any CAA record configured for the domain must
`allow Let's Encrypt <https://letsencrypt.org/docs/caa/>`_, otherwise certificate generation
may fail.
```
to avoid the overuse of *you* as per the doc guidelines and be a bit more direct in the process 🙂

auva-odoo (approved):
Thank you @mege-odoo, my suggestion for improvement is pretty minor so I'll go ahead and approve this already. Thanks!

@robodoo delegate+

content/applications/websites/website/configuration/domain_names.rst
mege-odoo: Thanks for tightening that up and the style tip! :-)

It also made me realize the original and commit description were inaccurate in the sense that it's not a singular CAA record or each individual CAA record that needs to be updated. If a domain has one or more CAA records, the user needs to add additional CAA records to allow Let's Encrypt, or remove the restrictive ones. So I kept your wording, but made "CAA records" plural.

mege-odoo:
@robodoo r+

PR #17469, opened 2026-04-20
[IMP] mrp: split and merge MOs

erjer-odoo:
[Task card](https://www.odoo.com/odoo/project.task/6126433).

* Set up replacements for "manufacturing order(s)" and "bill of materials" throughout article
* Update for functionality in 19.1
* New "Change Quantity to Produce" split button documentation
* Update screenshots according to most recent guidelines

This saas-19.1 PR **can** be FWP **up to saas-19.2**. There are further upcoming changes for the next release that will be documented in a separate PR.

PR #19344, opened 2026-08-06
[IMP] mrp: split and merge MOs

erjer-odoo:
## What this PR does and why it's needed
Updates the split and merge manufacturing orders documentation to reflect a UI improvement introduced in 19.3, where splitting an MO now uses a "Max Batch Size" field (with auto-calculated split count) instead of manually entering a fixed number of splits, and both the split and merge actions are accessed through a consolidated actions menu.

## Changes

**Split manufacturing orders**
- Documented the new two-path split workflow: from the *Manufacturing Orders* list view (via checkbox + `Actions --> Split`) and from an individual MO's actions menu
- Updated screenshots for the split production pop-up and the settings/actions menu to match the new UI

**Merge manufacturing orders**
- Clarified that MOs must use the same BoM to be merged
- Updated the select-orders screenshot to reflect the current UI

---
This `saas-19.3` PR can be FWP up to `master`.

**Note**: This looks like I touched the entire file, but the bulk of the work was already done for saas-19.1 and 19.2 in a [previous PR](https://github.com/odoo/documentation/pull/17469). I only added one section that is worth 2 points; that's why it's labeled as a `2`.


erjer-odoo:
Hi @stepperpig -- this PR is ready for your review when you get a moment. Thank you!

stepperpig (approved):
Hi @erjer-odoo, this looks good to me! Approved and ready for a final check.

Felicious (approved):
@robodoo r+

PR #19377, opened 2026-08-10
[FIX] conf: support more places for odoo source

Xavier-Do:
By default documentation will search in the current and parent directory for odoo/odoo-bin.

This can be a problem when working in a fork, or in any other case that would make the main odoo source dir not named odoo. (odoo-fork/odoo-bin)
```
~/src/odoo-fork
~/src/documentation[-fork]
```
The proposed solution allows to set the DOCUMENTATION_ODOO_SOURCE_DIR environment variable to make it
more flexible. A more common name could be used but wouldn't be stable friendly since variable name like ODOO_SOURCES_DIR, ODOO_SOURCESDIR, ... are already used in other situation and won't work in this case.

DOCUMENTATION_UPGRADE_UTIL_SOURCE_DIR can also be provide for upgrade-util sources.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -92,26 +92,32 @@ sys.path.insert(0, str(extension_dir.absolute()))
 
 # Search for the directory of odoo sources to know whether autodoc should be used on the dev doc
 odoo_sources_candidate_dirs = (Path('odoo'), Path('../odoo'))
-odoo_sources_dirs = [
-    d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()
-]
+odoo_source_dir = os.environ.get('DOCUMENTATION_ODOO_SOURCE_DIR')
+odoo_source_dir = Path(odoo_source_dir) if odoo_source_dir else None
+if not odoo_source_dir:
+    odoo_source_dir = next(
+    (d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()),
+    None,
+)
 odoo_dir_in_path = False
 
-if not odoo_sources_dirs:
+if not odoo_source_dir:
     _logger.warning(
         "Could not find Odoo sources directory in neither of the following folders:\n"
         "%(dir_list)s\n"
         "The 'Developer' documentation will be built but autodoc directives will be skipped.\n"
         "In order to fully build the 'Developer' documentation, clone the repository with "
-        "`git clone https://github.com/odoo/odoo` or create a symbolic link.",
+        "`git clone https://github.com/odoo/odoo` or create a symbolic link.\n"
+        "You can also provide the path to the Odoo sources directory with the environment "
+        "variable DOCUMENTATION_ODOO_SOURCE_DIR",
         {'dir_list': '\n'.join([f'\t- {d.resolve()}' for d in odoo_sources_candidate_dirs])},
     )
 else:
     if (3, 6) < sys.version_info < (3, 7):
         # Running odoo needs python 3.7 min but monkey patch version_info to be compatible with 3.6.
         sys.version_info = (3, 7, 0)
-    odoo_dir = odoo_sources_dirs[0].resolve()
-    source_read_replace_vals['ODOO_RELPATH'] = '/../' + str(odoo_sources_dirs[0])
+    odoo_dir = odoo_source_dir.resolve()
+    source_read_replace_vals['ODOO_RELPATH'] = '/../' + Path(os.path.relpath(odoo_dir, Path().resolve())).as_posix()
     sys.path.insert(0, str(odoo_dir))
     import odoo.addons
     odoo.addons.__path__.append(str(odoo_dir) + '/addons')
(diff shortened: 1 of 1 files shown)


AntoineVDV (changes requested):
Hello, IMHO we should go for the most straightforward solution, which is relying only on the environment variable (while keeping the current discovery system for setup stability).

The reason is that the primary contributors to the documentation are non-technical people, and the current discovery logic has already caused difficulties, so let's not add another layer of magic that comes with its own confusing behavior (from the POV of non-developers). For example, several contributors renamed their local `odoo/odoo` clone to avoid having the documentation build the developer documentation. If this PR were merged as-is, `odoo-bin` would suddenly be found, and their local setup would break.

Xavier-Do:
Works for me, I updated the pr to remove the additional discovery magic. 
I also change the env variable to be DOCUMENTATION_ODOO_SOURCE_DIR, we can change it if you want.

conf.py
```diff
@@ -92,26 +92,32 @@
...
-odoo_sources_dirs = [
-    d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()
-]
+odoo_sources_dir = os.environ.get('DOCUMENTATION_ODOO_SOURCE_DIR')
+odoo_sources_dir = Path(odoo_sources_dir) if odoo_sources_dir else None
+if not odoo_sources_dir:
+    odoo_sources_dir = next(
+    (d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()),
+    None,
+)
```
AntoineVDV, line 101:
```suggestion
        (d for d in odoo_sources_candidate_dirs if d.is_dir() and (d / 'odoo-bin').exists()),
        None,
    )
```

AntoineVDV, line 101: Why not consider `os.environ.get('DOCUMENTATION_ODOO_SOURCE_DIR')` as the first `odoo_sources_candidate_dirs`? This would keep the `d.is_dir()` and `(d / 'odoo-bin').exists()` checks on that candidate too, and shield non-technical contributors from ImportErrors if the variable is not correctly set/outdated.

conf.py
```diff
@@ -92,26 +92,32 @@
...
 odoo_dir_in_path = False
 
-if not odoo_sources_dirs:
+if not odoo_sources_dir:
     _logger.warning(
-        "Could not find Odoo sources directory in neither of the following folders:\n"
+        "Could not find Odoo sources directory in neither of the following folders\n"
```
AntoineVDV: Why remove the colon?

conf.py
```diff
@@ -92,26 +92,32 @@
...
         {'dir_list': '\n'.join([f'\t- {d.resolve()}' for d in odoo_sources_candidate_dirs])},
     )
 else:
     if (3, 6) < sys.version_info < (3, 7):
         # Running odoo needs python 3.7 min but monkey patch version_info to be compatible with 3.6.
         sys.version_info = (3, 7, 0)
-    odoo_dir = odoo_sources_dirs[0].resolve()
-    source_read_replace_vals['ODOO_RELPATH'] = '/../' + str(odoo_sources_dirs[0])
+    odoo_dir = odoo_sources_dir.resolve()
+    source_read_replace_vals['ODOO_RELPATH'] = '/../' + str(odoo_sources_dir)
```
AntoineVDV: This breaks if the env var points to a non-sibling directory.

conf.py
```diff
@@ -135,13 +143,19 @@
         odoo_dir_in_path = True
 
 if odoo_dir_in_path:
-    upgrade_util_dir = next(filter(Path.exists, [Path('upgrade-util'), Path('../upgrade-util')]), None)
+
+    upgrade_util_dir = os.environ.get('DOCUMENTATION_UPGRADE_UTIL_SOURCE_DIR')
+    upgrade_util_dir = Path(upgrade_util_dir) if upgrade_util_dir else None
+    if not upgrade_util_dir:
+        upgrade_util_dir = next(filter(Path.exists, [Path('upgrade-util'), Path('../upgrade-util')]), None)
```
AntoineVDV, line 150: Same as for the odoo sources

AntoineVDV (changes requested):
Don't forget to update README.md too, that's the place where you'd learn about the new env variables.

Nitpicking: "source dir" as you used in your env vars naming is more grammatically correct than the "source**s** dir" used in the Python variable names and warning logs; would you mind aligning them?

conf.py
Xavier-Do: Wouldn't this be a bugprone behaviour?
If DOCUMENTATION_ODOO_SOURCE_DIR is present, we should use it since it is explicitly requested, actually I even thought about allowing to set it empty to disable the usage of the odoo_source_dir, since you explained that some user are renaming odoo something else to remove the sources from the build.

conf.py
Xavier-Do: Nice catch, this is now transformed as a relative path

conf.py
AntoineVDV: I don't see why? It's the first candidate to let it be used if set, and it's skipped otherwise

Xavier-Do: Because if I set `DOCUMENTATION_ODOO_SOURCE_DIR` to something and it does not exists, it is potentially a typo or a misconfiguration, in your case it may silently continue with the fallback, i would prefer to have a warning instead of trying to debug why it didn't use my specific worktree or something. 
As you explained to me you proposed some devs to rename their odoo dir to something different to avoid having it used by the documentation. With the proposed solution the workaround would be to set `DOCUMENTATION_ODOO_SOURCE_DIR` to a random value that does not exist and it would work.

AntoineVDV (approved):
@robodoo delegate+

Xavier-Do:
robodoo r+

PR #19385, opened 2026-08-11
[IMP] pos: update the Receipts documentation

emmi-odoo:
task-6102710

### DIFF
--- a/content/applications/sales/point_of_sale/use/receipts.rst
+++ b/content/applications/sales/point_of_sale/use/receipts.rst
@@ -4,6 +4,10 @@
 Receipts
 ========
 
+Point of Sale allows users to :ref:`configure <pos/configuration/receipt-configuration>` receipts
+to be :ref:`printed <pos/configuration/receipt-printing>` or shared with customers via email
+or text message.
+
 +----------------------------------------------------------+---------------------------------+
 | POS receipts display the following elements:             |                                 |
 |                                                          |                                 |
@@ -24,20 +28,19 @@ Receipts
 Configuration
 =============
 
-To set up POS receipts, go to :menuselection:`Point of Sale --> Configuration --> Settings`, and
-scroll down to the :guilabel:`Bills & Receipts` section.
+To set up POS receipts, go to :menuselection:`Point of Sale --> Configuration --> Settings`, scroll
+down to the :guilabel:`Bills & Receipts` section, then:
 
-- To customize the header and footer, enable the :guilabel:`Custom Header & Footer` setting and
-  enter the information to be printed on the receipts in both fields.
-- To print receipts automatically upon payment confirmation, enable the :guilabel:`Automatic
-  Receipt Printing` setting.
-- To print receipts that don't display product prices, enable the :guilabel:`Basic Receipt` setting.
-- To allow for sending receipts via email, SMS, or WhatsApp from the :ref:`Receipt <pos/use/sell>`
-  screen, enable the :guilabel:`SMS Enabled` and/or :guilabel:`WhatsApp Enabled` setting(s).
-
-  .. note::
-     The :guilabel:`WhatsApp Enabled` setting is only available if the :guilabel:`WhatsApp
-     Messaging` module is :ref:`installed <general/install>`.
+- Enable :guilabel:`Custom header & Footer`, then enter the desired text in the :guilabel:`Header`
+  and :guilabel:`Footer` fields to customize receipts.
+- Enable :guilabel:`Automatic Receipt Printing` to print receipts automatically upon payment
+  confirmation.
+- Enable :guilabel:`Basic Receipt` to generate receipts that do not display product prices.
+- Enable :guilabel:`SMS Enabled` and/or, if the :ref:`WhatsApp <whatsapp/odoo-templates>` app is
+  installed, :guilabel:`WhatsApp Enabled` to allow users to share receipts with customers via SMS or
+  WhatsApp. Then, create or select a template from the :guilabel:`Receipt template` field.
+- Enable :guilabel:`Self-service invoicing` to add a QR code and URL on the receipt, allowing
+  customers to :ref:`request an invoice <pos/invoices/qr-codes>`.
 
 .. seealso::
    - :ref:`pos/restaurant/bills`
@@ -45,12 +48,19 @@ scroll down to the :guilabel:`Bills & Receipts` section.
    - :doc:`../hardware_network/receipt_printers`
    - `Receipts and Invoices (video tutorial) <https://youtu.be/w_DKgHcIV0U?si=Gnf6untzAz2zvNku>`_
 
-.. _pos/configuration/receipt-reprint:
+.. _pos/configuration/receipt-printing:
 
-Receipt reprint
-===============
+Receipt printing, sending, and reprinting
+=========================================
 
-To reprint a receipt, follow the next steps:
+To print the receipt from the :ref:`POS register <pos/use/open-register>`, click :icon:`fa-print`
+:guilabel:`Print Full Receipt` on the receipt screen after the payment is completed.
+
+To send the receipt from the :ref:`POS register <pos/use/open-register>` via email or text
+message/WhatsApp after the payment process, enter the customer's email address and/or phone number
+in the dedicated field, then click the relevant icon.
+
+To reprint a receipt, follow these steps:
 
 #. Access the :ref:`POS register <pos/use/open-register>`.
 #. Click :guilabel:`Orders` on the POS interface.
@@ -58,5 +68,12 @@ To reprint a receipt, follow the next steps:
 #. Select the desired order and click :icon:`fa-print` :guilabel:`Print Receipt`.
 
 .. tip::
-   Filter the list of orders using the search bar: type in your reference and select
+   Filter the list of orders using the search bar: type in your reference, then select
    :guilabel:`Receipt Number`, :guilabel:`Date`, or :guilabel:`Customer`.
+
+.. note::
+   - If a :ref:`receipt printer <pos/epos-printers/configuration>` is configured and the
+     :guilabel:`Automatic Receipt Printing` setting is enabled, the receipt is automatically printed
+     when the payment is completed.
+   - If a receipt printer is not configured, clicking :icon:`fa-print` :guilabel:`Print Full
+     Receipt` on the receipt screen triggers the browser's print function.


content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -4,6 +4,10 @@
 Receipts
 ========
 
+Point of Sale allows users to :ref:`configure <pos/configuration/receipt-configuration>` receipts
+that can be :ref:`printed and reprinted <pos/configuration/receipt-printing>`, or shared with
```
auva-odoo:
```suggestion
that can be :ref:`printed <pos/configuration/receipt-printing>`, or shared with
```
I don't think it's necessary to mention the reprint here in the intro

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -25,38 +29,49 @@ Configuration
 =============
 
 To set up POS receipts, go to :menuselection:`Point of Sale --> Configuration --> Settings`, and
-scroll down to the :guilabel:`Bills & Receipts` section.
+scroll down to the :guilabel:`Bills & Receipts` section:
```
auva-odoo:
```suggestion
To set up POS receipts, go to :menuselection:`Point of Sale --> Configuration --> Settings`,
scroll down to the :guilabel:`Bills & Receipts` section, then:
```
I would then change the wording of all list items in the list below following this example:
```suggestion
Enable :guilabel:`Custom header & Footer`, then enter the desired text in the :guilabel:`Header` and :guilabel:`Footer` fields to customize the receipts' header and footer.
```

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -25,38 +29,49 @@ Configuration
...
 
 - To customize the header and footer, enable the :guilabel:`Custom Header & Footer` setting and
   enter the information to be printed on the receipts in both fields.
 - To print receipts automatically upon payment confirmation, enable the :guilabel:`Automatic
   Receipt Printing` setting.
-- To print receipts that don't display product prices, enable the :guilabel:`Basic Receipt` setting.
-- To allow for sending receipts via email, SMS, or WhatsApp from the :ref:`Receipt <pos/use/sell>`
+- To print receipts that do not display product prices, enable the :guilabel:`Basic Receipt`
```
auva-odoo: I'd say *generate* instead of *print* since receipts aren't necessarily printed, unless this option is specific to printed receipts, and the same applies to any relevant settings described in this section (e.g., line 43)

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -25,38 +29,49 @@ Configuration
...
+  setting.
+- To allow users to share receipts with customers by email, SMS, or WhatsApp from the receipt
   screen, enable the :guilabel:`SMS Enabled` and/or :guilabel:`WhatsApp Enabled` setting(s).
```
auva-odoo: I'd mention (briefly) the additional configuration (fields) required for this (too bad both features don't work the same way though ...  🤐 )

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -25,38 +29,49 @@ Configuration
...
-.. _pos/configuration/receipt-reprint:
+.. _pos/configuration/receipt-printing:
+
+Receipt printing and reprint
+============================
 
-Receipt reprint
-===============
+To print the receipt in the :ref:`POS register <pos/use/open-register>`, click :icon:`fa-print`
+:guilabel:`Print Full Receipt` on the receipt screen.
```
auva-odoo: I would add something like: *after the payment process* or *after the payment is completed* in addition to *on the receipt screen* so we have a temporal reference as well as a UI one?

and should we mention the fact that the receipt is printed automatically if the related feature has been enabled?

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -25,38 +29,49 @@ Configuration
...
 #. Open the dropdown selection menu next to the search bar, and set the filter to :guilabel:`Paid`.
 #. Select the desired order and click :icon:`fa-print` :guilabel:`Print Receipt`.
 
 .. tip::
-   Filter the list of orders using the search bar: type in your reference and select
+   Filter the list of orders using the search bar: type in your reference, then select
    :guilabel:`Receipt Number`, :guilabel:`Date`, or :guilabel:`Customer`.
+
+.. note::
+   If a receipt printer is not connected to the database, clicking :icon:`fa-print` :guilabel:`Print
```
auva-odoo: Should we add a link to a receipt printer doc? or a seealso at the bottom of the section?

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -25,38 +29,49 @@ Configuration
...
 
-.. _pos/configuration/receipt-reprint:
+.. _pos/configuration/receipt-printing:
+
+Receipt printing and reprint
+============================
 
-Receipt reprint
-===============
+To print the receipt in the :ref:`POS register <pos/use/open-register>`, click :icon:`fa-print`
```
auva-odoo:
```suggestion
To print the receipt from the :ref:`POS register <pos/use/open-register>`, click :icon:`fa-print`
```

auva-odoo (changes requested):
Hi @emmi-odoo! In addition to my review comments, I'm curious to know why you mentioned receipt printing/reprinting (in the dedicated section), but not sending the receipt by email/SMS/WhatsApp? I'd suggest either:

- Updating the second h2 heading to include receipt sending, then covering those features in a short sentence (similar to what you did for printing), or
- Adding a separate h2 for receipt sending, though given how short those "instructions" would be, I'm not sure it's worth it. But let's try it if you prefer this option!

And could you please make sure the PR name + commit title have the following format: [TAG] app name: short description
(We rarely need to add a "module" for documentation so just the app name is enough)

content/applications/sales/point_of_sale/use/receipts.rst
emmi-odoo: How about something like this: 

"Enable :guilabel:`SMS Enabled` and/or :guilabel:`WhatsApp Enabled` to allow users to share receipts with customers by email, SMS, or WhatsApp from the receipt screen. Then, create or select a template from the :guilabel:`Receipt template`."

I'm wondering if we should mention the **Invoice template** for WhatsApp, even though we're more interested in the **Receipt template**.

content/applications/sales/point_of_sale/use/receipts.rst
emmi-odoo: I thought about it, it's worth mentioning the Automatic Receipt Printing here again, just in case someone only comes into this section and doesn't see it in the one above. 😄

content/applications/sales/point_of_sale/use/receipts.rst
emmi-odoo: Yes indeed

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -4,6 +4,10 @@
 Receipts
 ========
 
+Point of Sale allows users to :ref:`configure <pos/configuration/receipt-configuration>` receipts
+that can be :ref:`printed <pos/configuration/receipt-printing>`, or shared with customers by email
+or phone.
```
auva-odoo: I'm just catching this now but I fear that *sharing by phone* could be interpreted as *over the phone* (even though I do realize it makes no sense, but still). Could we find another way of saying this better?

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -24,39 +28,56 @@ Receipts
...
+  a template from the :guilabel:`Receipt template`.
+- Enable :guilabel:`Self-service invoicing` to allow customers to request an :ref:`invoice by
+  scanning the QR code or entering the URL <pos/invoices/qr-codes>` generated on the receipt.
 
-  .. note::
-     The :guilabel:`WhatsApp Enabled` setting is only available if the :guilabel:`WhatsApp
-     Messaging` module is :ref:`installed <general/install>`.
+.. note::
+   The :guilabel:`WhatsApp Enabled` setting is only available if the :ref:`WhatsApp Messaging
+   <whatsapp/odoo-templates>` module is :ref:`installed <general/install>`.
```
auva-odoo:
```suggestion
   The :guilabel:`WhatsApp Enabled` setting is only available if the :ref:`WhatsApp
   <whatsapp/odoo-templates>` app is :ref:`installed <general/install>`.
```
I know this is the name in the Apps menu, but actually Whatsapp seems to be an app, with its own icon and everything.
EDIT: but actually, i'd try to fit this in directly where we're mentioning the field since it's so short and minor (see my [comment on the related lines](https://github.com/odoo/documentation/pull/19385/changes#r3774795641))

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -24,39 +28,56 @@ Receipts
...
-- To allow for sending receipts via email, SMS, or WhatsApp from the :ref:`Receipt <pos/use/sell>`
-  screen, enable the :guilabel:`SMS Enabled` and/or :guilabel:`WhatsApp Enabled` setting(s).
+- Enable :guilabel:`Custom header & Footer`, then enter the desired text in the :guilabel:`Header`
+  and :guilabel:`Footer` fields to customize receipts.
+- Enable :guilabel:`Automatic Receipt Printing` to print receipts automatically upon payment
+  confirmation.
+- Enable :guilabel:`Basic Receipt` to generate receipts that do not display product prices.
+- Enable :guilabel:`SMS Enabled` and/or :guilabel:`WhatsApp Enabled` to allow users to share
+  receipts with customers by email, SMS, or WhatsApp from the receipt screen. Then, create or select
+  a template from the :guilabel:`Receipt template`.
```
auva-odoo: - This makes it sound like at least one of those features must be enabled to be able to send receipts by email and from what I tested, that's not the case.
- Missing word
- I'd try to fit in the contents of the note on line 46-47 directly here

Here's a suggestion taking all of this into account:

```suggestion
- Enable :guilabel:`SMS Enabled` and/or, if the :ref:`WhatsApp <whatsapp/odoo-templates>` app is installed, :guilabel:`WhatsApp Enabled` to allow users to share receipts with customers via SMS or WhatsApp. Then, create or select a template from the :guilabel:`Receipt template` field.
```

content/applications/sales/point_of_sale/use/receipts.rst
auva-odoo: If anything, I think the Invoice template mention would belong in the POS invoice doc

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -24,39 +28,56 @@ Receipts
...
+- Enable :guilabel:`Self-service invoicing` to allow customers to request an :ref:`invoice by
+  scanning the QR code or entering the URL <pos/invoices/qr-codes>` generated on the receipt.
```
auva-odoo: The QR code/url is not displayed on the receipt if this feature is not enabled, correct? In that case, i'd rephrase, this wording makes it sound like it is displayed anyways. It might just be me, but just in case:


```suggestion
- Enable :guilabel:`Self-service invoicing` to add a QR code and URL on the receipt allowing customers to :ref:`request an invoice <pos/invoices/qr-codes>`.
```

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -24,39 +28,56 @@ Receipts
...
+Receipt printing, sending, and reprint
+======================================
 
-Receipt reprint
-===============
+To print the receipt from the :ref:`POS register <pos/use/open-register>`, click :icon:`fa-print`
+:guilabel:`Print Full Receipt` on the receipt screen after the payment is completed.
 
-To reprint a receipt, follow the next steps:
+To send the receipt from the :ref:`POS register <pos/use/open-register>` by email or phone/WhatsApp
```
auva-odoo:
```suggestion
To send the receipt from the :ref:`POS register <pos/use/open-register>` by email or SMS/WhatsApp
```
if you don't want to use SMS we could use *text message* maybe? not phone though (see my comment as to why in the intro paragraph)

content/applications/sales/point_of_sale/use/receipts.rst
```diff
@@ -24,39 +28,56 @@ Receipts
...
+after the payment process, enter your email address and/or your phone number in the dedicated field,
+then click the relevant icon.
```
auva-odoo: you wouldn't be entering *your* email address or telephone number ...

auva-odoo (changes requested):

auva-odoo (approved):
Thanks @emmi-odoo, this LGTM 🙁 

@robodoo delegate+

emmi-odoo:
@robodoo r+

PR #19446, opened 2026-08-14
[IMP] accounting: clarify internal transfers process

dade-odoo:
taskid-6354695


auva-odoo (approved):
Thank you @dade-odoo, this LGTM!

@robodoo r+

PR #19504, opened 2026-08-16
[FIX] developer/actions - fix condition for calculating remaining records

sakesun:
We can infer remaining exhaustion only when the fetched records are less then limit provided.


kmagusiak (approved):

Feyensv (approved):
robodoo r+

Feyensv:
robodoo squash

PR #267, opened 2018-03-14
Odoo.sh detailed doc

dbo-odoo:
Detailed doc for odoo.sh instead of a small FAQ on the website.

@beledouxdenis 
I've opened the PR to allow a review to be made :)

PR #19530, opened 2026-08-18
[REM] odoo_sh: outdated import behavior

mege-odoo:
If you import a backup into an Odoo SH production branch, it does not disable scheduled actions like the documentation says, so this section is obsolete. It appears to have been unaltered since the original SH documentation was created in 2018: https://github.com/odoo/documentation/pull/267 .

If a customer imports a neutralized test backup into a production branch and wishes to use it as a production database, they do need to re-enable scheduled actions as well as manually change many other things. This process is complex and to be avoided, and better handled by support if there's no other way, so seems out of scope for this documentation.


xpl-odoo (approved):
Thanks @mege-odoo !

@robodoo r+

PR #19579, opened 2026-08-19
[IMP] point_of_sale: add QR code generator option

willbonde:
Adds JustMakeTheQRCode.com as another QR code generator example in the POS self-ordering customization tip.

Disclosure: I am affiliated with JustMakeTheQRCode.com.


LoredanaLrpz:
@robodoo r+

LoredanaLrpz (approved):

PR #19634, opened 2026-08-21
[ADD] accounting: document tax excl. / tax incl. toggle

dade-odoo:
taskid-6344726

Adds an explanation of how the new Tax Excl. / Tax Incl. toggle on quotes/SOs/POs/invoices works.
Also adds more context to the tax computation page in general to make the page more clear.

### DIFF
--- a/content/applications/finance/accounting/taxes/tax_computation.rst
+++ b/content/applications/finance/accounting/taxes/tax_computation.rst
@@ -2,6 +2,12 @@
 Tax computation
 ===============
 
+How taxes are computed depends on multiple different fields on the taxes form view. The :ref:`Tax
+Computation <taxes/computation>` field is located at the top of the taxes form view, but the
+:ref:`Included in Price <taxes/included-in-price>`, :ref:`Affect Base of Subsequent Taxes
+<taxes/base-subsequent>`, and :ref:`Base Affected by Preceding Taxes <taxes/base-affected>` fields
+are all located in the :guilabel:`Advanced Options` tab.
+
 .. _taxes/computation:
 
 Tax computation
@@ -200,11 +206,13 @@ defined in the :guilabel:`Formula` field. The Python expression may contain the 
 Included in price
 -----------------
 
+The :guilabel:`Included in Price` field determines if the price on products and invoices already
+includes this tax or if this tax should be applied on top of that value.
+
 .. tip::
    To set a company-wide default for this setting, go to :menuselection:`Accounting -->
    Configuration --> Settings`, find the :guilabel:`Taxes` section, and set the :guilabel:`Prices`
-   setting to :guilabel:`Tax Excluded` or :guilabel:`Tax Included`. This setting cannot be changed
-   once invoices have been created.
+   setting to :guilabel:`Tax Excluded` or :guilabel:`Tax Included`.
 
 :guilabel:`Default` indicates that the tax follows the company-wide default.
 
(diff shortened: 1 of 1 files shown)


content/applications/finance/accounting/taxes/tax_computation.rst
```diff
@@ -2,6 +2,12 @@
 Tax computation
 ===============
 
+How taxes are computed depends on multiple different fields on the taxes form view. The :ref:`tax
+computation <taxes/computation>` field is located at the top of the taxes form view, but the
+:ref:`Included in price <taxes/included-in-price>`, :ref:`Affect base of subsequent taxes
+<taxes/base-subsequent>`, and :ref:`Base affected by preceding taxes <taxes/base-affected>` fields
+are all located in the :guilabel:`Advanced Options` tab.
```
xpl-odoo:
```suggestion
How taxes are computed depends on multiple different fields on the taxes form view. The :ref:`Tax
Computation <taxes/computation>` field is located at the top of the taxes form view, but the
:ref:`Included in Price <taxes/included-in-price>`, :ref:`Affect Base of Subsequent Taxes
<taxes/base-subsequent>`, and :ref:`Base Affected by Preceding Taxes <taxes/base-affected>` fields
are all located in the :guilabel:`Advanced Options` tab.
```
I would capitalize the field labels as if these were inside guilabels, since you are giving instructions to find them on the UI :)

content/applications/finance/accounting/taxes/tax_computation.rst
```diff
@@ -215,16 +230,41 @@ computation will therefore compute a tax amount on top of the sales price.
 computation will therefore split the sales price into a tax-excluded base and the tax amount. This
 makes it suitable for B2C sales in most countries, where prices are quoted tax-inclusive.
 
+If a tax's :guilabel:`Included in Price` field is set to either :guilabel:`Tax Excluded` or
+:guilabel:`Tax Included`, the tax's behavior will not be changed by the :guilabel:`Tax Excl. / Tax
+Incl.` toggle on quotations, sales orders, purchase orders, and invoices.
+
```
xpl-odoo, line 246: What do you think about putting all the info about the toggle inside an alert block (could be a _note_ instead, or even a sub-heading?)
I gave it a try, and I think it's easier to follow that way. I didn't include your note about using the toggle as I felt it wasn't necessary here. I put the rst just in case you want to go for it, but feel free to rephrase/adapt as you see fit.

Image

```rst
:guilabel:`Default` indicates that the tax follows the company-wide default.

:guilabel:`Tax Excluded` indicates that the tax amount is not included in the sales price. The tax
computation will therefore compute a tax amount on top of the sales price.

:guilabel:`Tax Included` indicates that the tax amount is included in the sales price. The tax
computation will therefore split the sales price into a tax-excluded base and the tax amount. This
makes it suitable for B2C sales in most countries, where prices are quoted tax-inclusive.

.. important::
   If the :guilabel:`Default` option is selected, using the :guilabel:`Tax Excl. / Tax Incl.` toggle
   on a specific quotation, sales order, purchase order, or invoice changes this tax's behavior for
   that record only.

   However, if a tax's :guilabel:`Included in Price` field is set to either :guilabel:`Tax Excluded`
   or :guilabel:`Tax Included`, the tax's behavior will not be changed by the :guilabel:`Tax Excl. /
   Tax Incl.` toggle on quotations, sales orders, purchase orders, and invoices.

xpl-odoo (commented):
Thanks for the PR @dade-odoo !
Already delegating, even though I got a suggestion for you regarding the way we present the info about the toggle (feel free to use/adapt as you see fit)

@robodoo delegate+

content/applications/finance/accounting/taxes/tax_computation.rst
dade-odoo: Thanks for this recommendation - I do like the alert block idea, but I feel like that leaves out a little bit of context that would be nice to add (both an intro sentence explaining the use case and the note I had explaining that you should add products before using the toggle). I've tried a sub-section as you suggested, which I like, but it does mean that we have a single H3. I prefer the sub-section but I'm open to suggestions if you prefer the alert block! Let me know :)

xpl-odoo: The info about the toggle felt a bit out of place with the rest of the page so that's why I wasn't sure about having a separate section, but it's clearer that way so go for it :)

dade-odoo:
@robodoo r+

PR #19665, opened 2026-08-24
[IMP] POS: add section about ref price per uom

LoredanaLrpz:
task-6119076


content/applications/sales/point_of_sale/products.rst
```diff
@@ -149,10 +149,48 @@ the :guilabel:`Attributes & Variants` tab.
...
+-----------------------
+
+To display a reference price per :doc:`Unit of Measure (UoM)
+<../../inventory_and_mrp/inventory/product_management/configure/uom>` (e.g., `$20.00/kg`) alongside
+the sales price on product forms and printed :ref:`product labels
+<inventory/shipping_receiving/product-labels>`, go to :menuselection:`Point of Sale
+--> Configuration --> Settings`, scroll down to the :guilabel:`Pricing` section, and enable
+:guilabel:`Product Reference Price`.
+
+Then, to set up products, proceed as follows:
```
auva-odoo: I think we could be more specific here:
```suggestion
Then, specify the product's reference price as follows:
```
or maybe you wanted to be a bit more general because you also mention the sales price? If so, I'd mention something with prices still (maybe even *set up/configure product prices* or something along those lines)

content/applications/sales/point_of_sale/products.rst
```diff
@@ -149,10 +149,48 @@ the :guilabel:`Attributes & Variants` tab.
...
+
+#. Go to :menuselection:`Point of Sale --> Products --> Products` and open a product.
```
auva-odoo:
```suggestion
#. Go to :menuselection:`Point of Sale --> Products --> Products` and click the relevant product.
```

content/applications/sales/point_of_sale/products.rst
```diff
@@ -149,10 +149,48 @@ the :guilabel:`Attributes & Variants` tab.
...
+   :alt: Product Reference Price configuration on product forms
+
+.. example::
+
+   - **By weight:** A butter croissant weighs 60 g and sells for $1.20 per unit. To display its
+     reference price per kilogram, set the :guilabel:`Reference Unit` to `0.06 kg` (since 60 g =
+     0.06 kg). The reference price calculates automatically as `$20.00/kg`.
+   - **By pack quantity:** A 6-pack of soda sells for $5.00. Set the :guilabel:`Reference
+     Unit` to `6 cans`. The reference price calculates automatically as `$0.83/can`.
+
```
auva-odoo, line 193: Should we (briefly) mention the fact that if product variants are defined for the product, the reference unit must be defined for each individual variant?

auva-odoo (changes requested):
Thank you @LoredanaLrpz, this is really clear and complete (especially the examples, great idea to add 2 different use cases in there!) already as is, I just have a few suggestions 🙂

content/applications/sales/point_of_sale/products.rst
LoredanaLrpz: No, no, we could be more precise, as you suggested. I talk about the sales price just because it's necessary to compute the unit reference price, but it's not really the main subject at all.

auva-odoo (approved):
LGTM @LoredanaLrpz, thank you!!!

@robodoo r+

PR #19675, opened 2026-08-25
[IMP] payment_providers/payu: remove webhook configuration

shrm-odoo:
When the PayU payment provider was merged, merchants had to configure the webhook manually. However, this issue has now been fixed, so manual webhook configuration is no longer required. Since we already send the `webhook_url` in the request payload, PayU automatically configures the webhook using that URL.


AntoineVDV (approved):
@robodoo r+ squash

PR #19739, opened 2026-08-26
[IMP] Purchase: Control Policies update

jero-odoo:
Docs task: https://www.odoo.com/odoo/project/3835/tasks/6482454

Updating doc to remove task about error message when trying to create a vendor bill without having received the product.


jero-odoo:
@erjer-odoo sending this to you for final since it is just a small edit

erjer-odoo (approved):
@robodoo r+

PR #19799, opened 2026-08-27
[IMP] pos: update Scale documentation page

emmi-odoo:
task-6045371


emmi-odoo:
Hello @odoo/be-doc-review , 

From 19.4, it is possible to connect the scale using a USB cable. However, the connection is only direct with a Windows computer. For Linux, for example, we need to open the device terminal and type a command (we tested both operating systems). 

image

content/applications/general/iot/devices/scale.rst
```diff
@@ -55,5 +56,8 @@ Guide for Ariva S series scales <https://www.mt.com/dam/RET_DOCS/Ariv.pdf>`_ and
 
 #. Press **>0<**.
 #. Press **>0<** again to :guilabel:`SAVE`; the scale restarts.
-#. Reboot the IoT box or :ref:`restart the Windows virtual IoT service <iot/windows_iot/restart>`.
-   The scale should then appear as `Toledo 8217`.
+
+.. tip::
+   Reboot the :doc:`IoT box <../../../../applications/general/iot/iot_box>` or :ref:`restart the
+   Windows virtual IoT service <iot/windows_iot/restart>`. The scale should then appear as `Toledo
+   8217`.
```
auva-odoo: If this is a required step, I would leave it as it was, as part of the instructions.
And also, we can use a relative path here:

```suggestion
.. tip::
   Reboot the :doc:`IoT box <../iot_box>` or :ref:`restart the
   Windows virtual IoT service <iot/windows_iot/restart>`. The scale should then appear as `Toledo
   8217`.
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
-   <https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=uriserv%3AOJ.L_.2014.096.01.0107.01.ENG>`_
-   to use a scale as an integrated device.
+   - In EU member states, `certification is legally required
+     <https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=uriserv%3AOJ.L_.2014.096.01.0107.01.ENG>`_
+     to use a scale as an integrated device.
+   - The **Mettler Toledo scale (Ariva S)** using protocol 8217 is recommended with Odoo.
+   - Connecting a scale to Odoo requires setting up an :doc:`IoT system
+     </applications/general/iot>`. For optimal reliability, it is recommended to use a
+     :doc:`physical IoT system <../../../../applications/general/iot/iot_box>` rather than a
+     :doc:`virtual one <../../../../applications/general/iot/windows_iot>`.
```
auva-odoo: same as above about relative paths

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
 ----------------
 
-#. :ref:`Access the POS settings <pos/use/settings>`.
+To connect the scale with your Odoo database, follow these steps:
+
+#. Connect the :ref:`IoT system to Odoo <iot/iot-odoo/connection>`.
+#. Go to :menuselection:`Point of Sale --> Configuration --> Settings`.
 #. Scroll down to the :guilabel:`Connected Devices` section and enable :guilabel:`IoT Box`.
-#. Select the scale in the :guilabel:`Electronic Scale` field.
+#. Select the scale in :guilabel:`Electronic Scale` field.
```
auva-odoo: not sure why you removed the *the* here?

```suggestion
#. Select the scale in the :guilabel:`Electronic Scale` field.
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
+.. _pos/scale/config-connection:
+
 Scale connection
 ----------------
 
-#. :ref:`Access the POS settings <pos/use/settings>`.
+To connect the scale with your Odoo database, follow these steps:
+
+#. Connect the :ref:`IoT system to Odoo <iot/iot-odoo/connection>`.
+#. Go to :menuselection:`Point of Sale --> Configuration --> Settings`.
```
auva-odoo, line 31: From what I remember, you have a popup at the end of of the IoT setup process that allows you to select the POS or something like that (which enables the IoT box setting in the POS's settings as well), hence the previous description. Is that still the case? If it is, I guess it could be another way of connecting the IoT box to the POS and we should probabl 1) check which way is the most common and 2) have the alternative way in a tip or something.

Could you please check?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
 .. tip::
-   Alternatively, click the :icon:`fa-ellipsis-v` (:guilabel:`Dropdown menu`) icon on a POS card and
-   click :guilabel:`Edit` to access this setting.
-
-Product configuration
----------------------
+   - If the scale is not recognized, :ref:`follow the Ariva guidelines to reboot it
+     <iot/scale/ariva-s>`.
+   - Turn on the scale before starting the IoT system to ensure the scale is detected during the IoT
+     boot process. Alternatively, restart the IoT system to detect the scale.
```
auva-odoo: what do you mean by *boot* process? setup? reboot?

but then based on this, I believe this should actually be the first step in the instructions above (on line 27 of this file) then, shouldn't it?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
+#. Select the scale in :guilabel:`Electronic Scale` field.
 #. Click :guilabel:`Save`.
 
 .. tip::
-   Alternatively, click the :icon:`fa-ellipsis-v` (:guilabel:`Dropdown menu`) icon on a POS card and
-   click :guilabel:`Edit` to access this setting.
-
-Product configuration
----------------------
+   - If the scale is not recognized, :ref:`follow the Ariva guidelines to reboot it
```
auva-odoo: do you mean reconfigure the scale as outlined on lines 42-63 in scale.rst?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -81,20 +91,34 @@ turns green.
...
-#. Place the product on the scale and wait for the weight to be displayed in the popup window.
-#. Once the weight is determined, the price is automatically computed.
+#. On the :guilabel:`General Information` tab, set the :guilabel:`Sales Price` to the relevant
+   unit of mass.
+#. Click the :guilabel:`Point of Sale` tab and enable :guilabel:`To Weigh With Scale` to weigh the
+   product directly on the connected scale.
+#. Save the product, then access the :ref:`POS register <pos/use/open-register>`.
+#. Select the created product in the grid. A popover displays the product's name.
+#. Place the physical product on the scale. Once the weight is determined, the price is
+   automatically computed in the popover.
```
auva-odoo, line 118: I can't test this but I assume the product's quantity is also updated?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
-#. Ensure the :guilabel:`Point of Sale` checkbox is activated for the product to be available in
-   POS.
-#. On the :guilabel:`General Information` tab, define a :guilabel:`Sales Price` per :guilabel:`kg`.
+.. _pos/scale/config-product:
 
-   .. note::
-      This step requires to enable the :doc:`Units of Measure
-      </applications/inventory_and_mrp/inventory/product_management/configure/uom>` feature. To
-      activate it:
+Product creation
```
auva-odoo: Could we find a better heading? I think this is super confusing and product creation usually refers to actually creating products (in the backend for example) and configuring them.
Whereas this section just refers to the scale-related configuration, correct?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
 
 .. important::
-   The selected unit of measure for weighable products must be :guilabel:`kg` to ensure compliance
-   with **European regulations**.
+   To comply with :ref:`European regulations <pos/scale/eu-regulations>`, the unit of measure for
+   weighable products must be set to `kg`.
+
+.. tip::
+   Ensure the :guilabel:`Point of Sale` checkbox is activated on the product form to make the
+   product available in Point of Sale.
```
auva-odoo: I would rephrase this wole section to:

1) rework the section to focus on everything configuration-related: enabling the inventory setting + product configuration (move the product configuration info from the "product weighing" section here)
3) I'd remove the "summary"/intro, I don't think it's needed or relevant here
4) Write a short intro focusing on configuration (to use a scale/weigh products in POS, enable the inventory setting, then configure the product as follows --> this is just a general idea to explain what I mean)
5) check the paths: `<../../../../applications/inventory_and_mrp/inventory/product_management/configure/uom>` should be either `<../../../inventory_and_mrp/inventory/product_management/configure/uom>` or `</applications/inventory_and_mrp/inventory/product_management/configure/uom>` for example

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
+.. _pos/scale/eu-regulations:
+
 European regulations
 ====================
 
-When using scales in commercial transactions, the database integrated with a scale must be
-configured to meet specific European requirements. This includes supporting at least three decimal
-places for accuracy and using proper rounding for units of measure, such as `kg` instead of generic
-`units`.
+When using scales in commercial transactions, the database connected to the scale must be
```
auva-odoo: technically, the scale is connected to the database, not the other way around, is it?
maybe *with* would be more appropriate than *to* here?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -81,20 +91,34 @@ turns green.
    Both the :doc:`customer <customer_display>` and POS displays must have a minimum diagonal
    size of 6 inches. For optimal readability, larger screens are recommended.
 
-Using a scale in PoS
-====================
+.. _pos/scale/product-weighing:
+
+Product weighing
+================
+
```
auva-odoo, line 111: I'd remove the configuration-specific info and just focus on the "use" part in this section

I also wonder: how common is it that people are going to weight products *without* a connected scale? i.e., that they're going to manually enter the quantity?
and would that workflow be possible in the register? I tried setting the quantity manually but it just updated the number of "units" I suppose.
Depending on the answer, should we take this workflow into account as well in the "product weighing" section? in a tip or something? sorry if this is a stupid question

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -81,20 +91,34 @@ turns green.
...
+   unit of mass.
+#. Click the :guilabel:`Point of Sale` tab and enable :guilabel:`To Weigh With Scale` to weigh the
+   product directly on the connected scale.
+#. Save the product, then access the :ref:`POS register <pos/use/open-register>`.
+#. Select the created product in the grid. A popover displays the product's name.
+#. Place the physical product on the scale. Once the weight is determined, the price is
+   automatically computed in the popover.
 #. Click :guilabel:`Order` :icon:`fa-angle-double-right` to add the product to the cart.
-#. Remove the previous product from the scale.
+#. Remove the product from the scale and continue with the :ref:`payment process <pos/use/sell>`.
```
auva-odoo: not necessarily, you might want to add other products before you proceed to payment

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,79 @@ Scale
...
 Configuration
 =============
 
+.. _pos/scale/config-connection:
+
 Scale connection
 ----------------
 
-#. :ref:`Access the POS settings <pos/use/settings>`.
+To connect the scale with your Odoo database, follow these steps:
```
auva-odoo: technically we're just connecting it to the POS here, aren't we?
like it could already be connected to the database (through the iot box), but not to the POS?

auva-odoo (changes requested):
Thank you @emmi-odoo, let me know if it'd be helpful to discuss my comments IRL.
And could you also please make sure to use correct relative paths (I think I added comments every time I saw one, but it might be best to check them all based on those comments just in case!).

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: true

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: I wonder why too

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: Yes, shall I rephrase the sentence?

auva-odoo: Yes, but now that I think of it, this is only valid for that specific scale. What if they have a different scale? Do they have to reconfigure it too? or just restart it?

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: I've tried several paths, the only ones that work are the ones that I previously wrote 🥲. Do you know what other combination would work best? 

Image

Image

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: It's nice your brought this up, because every time I was testing it, I never got such popover. However, I just got back from another test session with Isabel and out of 5 times testing, we got the popover 2ice. We asked the devs and they said the popover is supposed to be there. If it doesn't appear, then it's a bug. So, I'll add it in the doc. But only the IoT checkbox is enabled in the settings, you still need to add the Scale in the settings after connecting it (see second image). 

Image

Image

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: I don't think they'd follow this documentation if they have another scale than the one that is recommended with Odoo, as the integration might not work (unless they have a scale that emulates the 8217 protocol - then, they'd follow this doc as if it was a Toledo scale). Would you like the rephrase to focus on Ariva, still?

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: Yes, the setup process, let's keep this formulation 😃 And I'll put part of the sentence as the first step, then keep the sentence "Alternatively, restart the IoT ..." in the tip, what do you think?

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: What about "Units of measure"?

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: Only the **kg** and **price** are updated, because in the product form, we change the "per Unit" to "per kg".

content/applications/sales/point_of_sale/hardware_network/scale.rst
auva-odoo: Yes, good idea, let's see what that would look like :slightly_smiling_face:

content/applications/sales/point_of_sale/hardware_network/scale.rst
auva-odoo: maybe this part, yes? i'd have to see what it looks like first though so let's try and reevaluate in the re-review if that's ok with you?

content/applications/sales/point_of_sale/hardware_network/scale.rst
auva-odoo: but where do you see the kg updated in the register? I couldn't find it but it might just be me :see_no_evil:

content/applications/sales/point_of_sale/hardware_network/scale.rst
auva-odoo: Given the fact that I suggested moving the actual product configuration here, I feel like we probably need a heading that would incude both? as you can't proceed with the product configuration (as far as UoM go at least- without first enabling the inventory feature, correct? so the 2 are definitely linked.

I think "Product configuration" would probably a good start (and you'd still mention the inventory feature activation as a first step of the scale/weighing-specific product configuration)

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: You can get a visual [here](https://youtu.be/vGNe-PmulAA?si=G2CDbqwbRiGD3VhK&t=287)

content/applications/general/iot/devices/scale.rst
```diff
@@ -40,13 +42,12 @@ specific setting must be modified, and a dedicated Mettler USB-to-proprietary RJ
 To configure the Ariva S scale for IoT system recognition, refer to page 17 of `Mettler's Setup
```
auva-odoo: stupid question but are we sure this is still "page 17"? 🙈

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+
 Scale connection
 ----------------
 
-#. :ref:`Access the POS settings <pos/use/settings>`.
-#. Scroll down to the :guilabel:`Connected Devices` section and enable :guilabel:`IoT Box`.
-#. Select the scale in the :guilabel:`Electronic Scale` field.
+To connect the scale with Point of Sale, follow these steps:
+
+#. Turn on the scale before starting the IoT system to ensure it is detected during setup.
```
auva-odoo, line 27: Should we add a step after this one to say that we need to connect the scale to the IoT system? and refer to this? /applications/general/iot/devices/scale

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+   :guilabel:`Electronic Scale` field.
 #. Click :guilabel:`Save`.
 
 .. tip::
-   Alternatively, click the :icon:`fa-ellipsis-v` (:guilabel:`Dropdown menu`) icon on a POS card and
-   click :guilabel:`Edit` to access this setting.
-
-Product configuration
----------------------
+   - If the scale is not recognized by the IoT system, :ref:`reconfigure it by following the
```
auva-odoo: sorry, this is probably my own suggestion but I would say this (to make sure we also take other scales into account):

```suggestion
   - If the scale is not recognized by the IoT system, restart it or :ref:`reconfigure it by following the
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+     setup guidelines <iot/scale/ariva-s>`. If the issue persists, restart the IoT system.
+   - Alternatively, to connect the IoT system to the database, go to the :ref:`POS settings
```
auva-odoo: This only connects the IoT system to the **POS** and to do that, it needs to be connected to the database already AFAIK. Hence:

```suggestion
   - Alternatively, to connect the IoT system to the POS after it is :ref:`connected to the database <iot/iot-odoo/connection>, go to the :ref:`POS settings
```
(the ref ensures we don't confuse the two)

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+     <pos/use/settings>`, scroll down to the :guilabel:`Connected Devices` section, and enable the
+     :guilabel:`IoT Box` setting.
+   - Alternatively, to enable and configure the :guilabel:`IoT Box` setting, click the
+     :icon:`fa-ellipsis-v` (:guilabel:`Dropdown menu`) icon on a POS card, then click
+     :guilabel:`Configure`.
```
auva-odoo: honestly I feel like we should remove this, I really don't see the point

but if you want to keep it, I'd change the "Alternatively" since we already have that wording in the list item above

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
 
-In order to weigh products using an integrated scale, go to :menuselection:`Point of Sale -->
-Products --> Products`, create a product or open an existing product form, and configure it as
-follows:
+.. seealso::
+   - :doc:`/applications/general/iot/connect`
+   - :doc:`/applications/general/iot/devices/scale`
```
auva-odoo: i'd add a label for the ref to avoid confusion (connect the scale to what?)


```suggestion
   - :doc:`Connect a scale to an IoT system </applications/general/iot/devices/scale>`
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
-      activate it:
+Product configuration
+---------------------
 
-      #. Go to :menuselection:`Inventory --> Configuration --> Settings`.
-      #. Scroll down to the :guilabel:`Products` section and activate :guilabel:`Units of Measure`.
-#. Go to the :guilabel:`Point of Sale` tab and activate :guilabel:`To Weigh With Scale`. This
-   enables the product to be weighed directly on the connected scale at the POS.
+To configure the scale to weigh products and display their weight in the :ref:`POS register
+<pos/use/open-register>`, follow these steps:
```
auva-odoo: you're not configuring the scale though ...
I'm not sure what I'd replace it with though so I think I would rephrase to avoid this completely. Something like:


```suggestion
The following configuration is required to be able to weigh products and display their weight in the :ref:`POS register <pos/use/open-register>`:
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+
+#. Go to :menuselection:`Inventory --> Configuration --> Settings`.
+#. Scroll down to the :guilabel:`Products` section, enable the :doc:`Units of Measure & Packagings
+   </applications/inventory_and_mrp/inventory/product_management/configure/uom>` setting, and click
+   :guilabel:`Save`.
+#. Go to :menuselection:`Point of Sale --> Products --> Products`.
+#. Click :guilabel:`New` to :ref:`create a product <pos/products/creation>` or click an existing
+   product form.
```
auva-odoo: we're not clicking a product form (Kanban view vs form view), and I'd actually have the reverse since we're probably going to be editing an existing product:


```suggestion
#. Click an exisiting product or click :guilabel:`New` to :ref:`create a new one <pos/products/creation>`.
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+#. On the :guilabel:`General Information` tab, set the :guilabel:`Sales Price` to the relevant
```
auva-odoo:
```suggestion
#. In the :guilabel:`General Information` tab, set the :guilabel:`Sales Price` to the relevant
```
and before we mention this, should we add that we need to make sure the Point of sale checkbox is enabled (because otherwise we won't see the POS tab)?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+   unit of mass.
+#. Click the :guilabel:`Point of Sale` tab and enable :guilabel:`To Weigh With Scale` to weigh the
+   product directly on the connected scale.
```
auva-odoo:
```suggestion
#. Click the :guilabel:`Point of Sale` tab and enable :guilabel:`To Weigh With Scale` to be able to weigh the
   product directly on the connected scale.
```
or something like "if you wish to weight ..." but the original phrasing makes it sound like this weighs the product or something, it's a bit weird

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+#. Save the product form.
```
auva-odoo: you don't have to save manually, it's saved automatically; I'd remove this?

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
 
 .. important::
-   The selected unit of measure for weighable products must be :guilabel:`kg` to ensure compliance
-   with **European regulations**.
+   To comply with :ref:`European regulations <pos/scale/eu-regulations>`, the unit of measure for
+   weighable products must be set to `kg`.
+
+.. tip::
+   Ensure the :guilabel:`Point of Sale` checkbox is activated on the product form to make the
+   product available in Point of Sale.
```
auva-odoo: ah yes I see but as mentionned in my previous comment, this isn't only to make it available in the POS; if it's not enabled you won't even be able to perform the config steps

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -3,69 +3,92 @@ Scale
...
+
 European regulations
 ====================
 
-When using scales in commercial transactions, the database integrated with a scale must be
-configured to meet specific European requirements. This includes supporting at least three decimal
-places for accuracy and using proper rounding for units of measure, such as `kg` instead of generic
-`units`.
+When using scales in commercial transactions, the database to which the scale is connected must be
+configured to comply with specific European requirements. This includes supporting at least three
```
auva-odoo:
```suggestion
configured to comply with European requirements. This includes supporting at least three
```

content/applications/sales/point_of_sale/hardware_network/scale.rst
```diff
@@ -81,20 +104,29 @@ turns green.
...
+================
+
+To weigh a product in the :ref:`POS register <pos/use/open-register>` using the :ref:`connected
+scale <pos/scale/config-connection>`, follow these steps:
 
-#. :ref:`Access the POS register <pos/use/open-register>`.
-#. Select the product to weigh on the order screen or scan its barcode.
-#. Place the product on the scale and wait for the weight to be displayed in the popup window.
-#. Once the weight is determined, the price is automatically computed.
+#. Select the :ref:`created product <pos/scale/config-product>` in the grid. A popover displays the
```
auva-odoo:
```suggestion
#. Select the relevant :ref:`product <pos/scale/config-product>` in the grid. A popover displays the
```

auva-odoo (changes requested):
Thanks for the changes @emmi-odoo, the structure looks much more logical to me! I have a ew more comments on the wording/content 🙂

content/applications/general/iot/devices/scale.rst
emmi-odoo: Yes, but in case it changes, I can adjust the sentence ;)

content/applications/sales/point_of_sale/hardware_network/scale.rst
emmi-odoo: It's another way of enabling the setting, so out of consistency with the doc, I think we can leave it, but remove "alternatively" indeed

auva-odoo (approved):
@robodoo r+

PR #19834, opened 2026-08-28
[IMP] pos: Gift card links

kop-odoo:
task-6245292

Updated gift card `:ref:`s to point to the newly created documentation section.


auva-odoo (approved):
Good catch @kop-odoo, thank you!

@robodoo r+

PR #19876, opened 2026-09-01
[IMP] accounting: Inventory val - variation account section clarif

afma-odoo:


xpl-odoo:
@robodoo delegate+

afma-odoo:
@xpl-odoo Thanks for the review ;)

@robodoo r+

PR #19889, opened 2026-09-02
[IMP] Payroll - Payslip printing clarification 17.0

larm-odoo:
Received a request to update the printing section of this doc, and all future docs. This is to be published only on 17.0. A new PR is needed for 18 and 19 since the docs have changed significantly between versions. 

This is not updating all formatting that changed over time, this is ONLY for the one sentence in this PR regarding printing. 

Original [task card](https://www.odoo.com/odoo/project/3835/tasks/5875693) for this PR.


larm-odoo:
Hi @erjer-odoo - this 1 point PR is going right to you (not Nikki since this is not a FIX/redirect). I believe that is ok - if not, please let me know!

erjer-odoo (approved):
@robodoo r+

larm-odoo:
@robodoo fw=no

PR #19892, opened 2026-09-02
[IMP] General: Inactivity timeout correction

jero-odoo:
Correcting broken link and typo


jero-odoo:
@erjer-odoo sending straight to you for final, thanks!

erjer-odoo (approved):
@robodoo r+

PR #19894, opened 2026-09-02
[IMP] Payroll - Saudi Arabia Salary Advance 19.2

larm-odoo:
Per this [task card](https://www.odoo.com/odoo/project/3835/tasks/6344461), the salary advance changed in 19.2, and there is no need to click a button - there is a new section on the payslip.


larm-odoo:
Another 1 point PR, @erjer-odoo!

erjer-odoo (approved):
@robodoo r+

PR #19935, opened 2026-09-04
[FIX] developer/reference/cli: -c/-s lookup order changed

Julien00859:
https://github.com/odoo/odoo/pull/251414 changed the lookup order for `--config` and `--save` to be compatible with XDG standard locations. We forgot to reflect the changes in the documentation.

### DIFF
--- a/content/developer/reference/cli.rst
+++ b/content/developer/reference/cli.rst
@@ -139,10 +139,22 @@ This command is the default one: you can omit it, and it will be chosen anyway.
 
 .. option:: -c <config>, --config <config>
 
-    path to an alternate :ref:`configuration file <reference/cmdline/config>`.
-    If not defined, Odoo checks ``ODOO_RC`` environmental variable
-    and default location :file:`{$HOME}/.odoorc`.
-    See configuration file section :ref:`below <reference/cmdline/config>`.
+    path to an alternate :ref:`configuration file <reference/cmdline/config>`,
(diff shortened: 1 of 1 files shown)


content/developer/reference/cli.rst
```diff
@@ -139,10 +139,22 @@ This command is the default one: you can omit it, and it will be chosen anyway.
...
+    * :file:`/home/$USER/.config/odoo/odoo.conf`
+    * :file:`/etc/xdg/xdg-cinnamon/odoo` (XDG site, can vary)
+    * :file:`/home/$USER/.odoorc`
+    
+    Windows:
+
+    * :file:`C:\Users\%USERNAME%\AppData\Local\odoo\odoo\odoo.conf`
+    * :file:`C:\ProgramData\odoo\odoo\odoo.conf`
+    * :file:`odoo.conf` file next to the ``odoo-bin`` executable
+    * :file:`C:\Users\%USERNAME%\.odoorc`
```
Feyensv, line 157: You could use a `..tabs::` with two separate `..tab::` for the two cases :eyes:

Feyensv (approved):
Minor suggestion otherwise lgtm 👌 

robodoo delegate+

content/developer/reference/cli.rst
Julien00859: I could, but I find tabs to be a worse UX in this case, thanks for suggesting tho

Julien00859:
@robodoo r+

PR #19984, opened 2026-09-08
[IMP] user_interface: add dialog_size view arch attribute

aab-odoo:
Views' archs now support a `dialog_size` attribute on their root node, used to determine the dialog's size when the view is rendered inside one. Document it once as a shared root attribute, referenced from every view type.

Task-6429925


AntoineVDV (approved):
@robodoo r+ delegate+

PR #19482, opened 2026-08-15
[IMP] phone: remove schedule later, clarify actions

huisit:
Updates the Phone app documentation (content/applications/productivity/phone.rst and content/applications/productivity/phone/phone_widget.rst) to match recent UI changes to the phone widget: the Schedule Later action is gone, the Contacts tab was dropped, and the call-actions menu now separates History from Log.

- phone.rst: reworded the intro paragraph and linked "link calls and messages to customer interactions" to the activities doc via :doc:.
- phone_widget.rst:
  - Removed documentation of the Contacts tab (no longer one of the widget's tabs) and reordered the remaining tab list to Recent → Keypad → Activities.
  - Split the old Log drop-down action into two distinct entries:
      - History (Recents tab, Keypad tab): view the call history for the selected contact.
    - Log (Recents tab only): log the call as a completed activity on a related record (clarifies it logs a completed activity, not a scheduled one).

This saas-19.4 branch can be FWP to master.

PR #19992, opened 2026-09-08
[FIX] phone: correct number of navigation tabs

huisit:
quick fix to decrement # of tabs in the softphone widget from 4 → 3

follow-up to https://github.com/odoo/documentation/pull/19482, which removed _contacts_ tab details since it was removed from the product


huisit:
@erjer-odoo this quick fix is ready for review!

erjer-odoo (approved):
@robodoo r+

PR #20013, opened 2026-09-10
[IMP] accounting: edi update

tohor-odoo:
taskid-6545739

### DIFF
--- a/content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
+++ b/content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
@@ -69,42 +69,52 @@ Registration
 ------------
 
 To register on Peppol, go to :menuselection:`Accounting --> Configuration --> Settings` and scroll
-to the :guilabel:`PEPPOL Electronic Invoicing` section. Then, follow these steps:
+down to the :guilabel:`Electronic Invoicing` section. Then, follow these steps:
 
-#. Click :guilabel:`Activate Electronic Invoicing` and fill in the following fields:
+#. Click :guilabel:`Activate Peppol` and fill in the following fields:
 
-   - Using the :icon:`fa-caret-down` :guilabel:`(down arrow)` icon, make sure the relevant
-     country-specific Peppol endpoint identifier is selected in the dropdown list, then enter your
-     Peppol endpoint (usually a Company Registry or VAT number).
-   - :guilabel:`Email`
-   - :guilabel:`Phone`, including the country code (e.g., `+32` in Belgium)
+   - Select an identifying method for your company (VAT, company registry number, or the relevant
+     option) and enter the related information.
+   - :guilabel:`Email`.
+   - :guilabel:`Phone`, including the country code (e.g., `+32` in Belgium).
 
-#. Click :guilabel:`Activate Peppol`. The registration is then pending activation and should be
-   automatically activated within a day.
+#. Click :guilabel:`Activate Peppol`. The registration is then pending activation and should
+   be automatically activated within a day.
 
    .. seealso::
       `Peppol endpoint - OpenPeppol eDEC Code Lists <https://docs.peppol.eu/edelivery/codelists/>`_
       (open the "Participant Identifier Schemes" as HTML page)
 
-#. Define where documents should be received:
+Configuration
+-------------
 
-   - :guilabel:`Receive in Journal`: If necessary, select another purchase journal in the
-     :guilabel:`Incoming Invoices Journal` field.
-   - :doc:`Receive in Documents <../../../productivity/documents>`: Select a folder in the
-     :guilabel:`Document Workspace` field if multiple purchase journals are used.
+#. Once Peppol is activated:
+
+#. Go to :menuselection:`Accounting --> Configuration --> Settings` and scroll down to the
+   :guilabel:`Electronic Invoicing` section.
+#. Modify the :guilabel:`Peppol Contact Email` if needed.
+#. Select the :guilabel:`Peppol Participation Role`:
+   - :guilabel:`Sending & Receiving`: to both send *and* receive Peppol documents.
+   - :guilabel:`Sending Only`: to send Peppol documents only.
+
+#. If :guilabel:`Sending & Receiving` is selected as the :guilabel:`Peppol Participation Role`,
+   define where Peppol documents should be received:
+
+   - :guilabel:`Receive as Bills`: If multiple purchase journals are used, select the journal where
+     the bills should be posted in the :guilabel:`Incoming Accounting Journal` field.
+   - :guilabel:`Receive as Documents`: Select the :guilabel:`Document Folder` where the Peppol
+     documents should be stored and the :guilabel:`Document Tags` to assign to them.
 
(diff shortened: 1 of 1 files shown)


content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
```diff
@@ -69,37 +69,39 @@ Registration
...
 
-#. Click :guilabel:`Activate Electronic Invoicing` and fill in the following fields:
+#. Click :guilabel:`Activate Peppol` and fill in the following fields:
 
-   - Using the :icon:`fa-caret-down` :guilabel:`(down arrow)` icon, make sure the relevant
-     country-specific Peppol endpoint identifier is selected in the dropdown list, then enter your
-     Peppol endpoint (usually a Company Registry or VAT number).
-   - :guilabel:`Email`
-   - :guilabel:`Phone`, including the country code (e.g., `+32` in Belgium)
+   - Select an identifying method for your company (VAT, company registry number, or your local
```
auva-odoo:
```suggestion
   - Select an identifying method for your company (VAT, company registry number, or the relevant local
```

content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
```diff
@@ -69,37 +69,39 @@ Registration
...
+     option) and enter the related information.
+   - :guilabel:`Email`.
+   - :guilabel:`Phone`, including the country code (e.g., `+32` in Belgium).
 
-#. Click :guilabel:`Activate Peppol`. The registration is then pending activation and should be
-   automatically activated within a day.
+#. Click :guilabel:`Activate Peppol` again. The registration is then pending activation and should
```
auva-odoo:
```suggestion
#. Click :guilabel:`Activate Peppol`. The registration is then pending activation and should
```
The label is the same but it's not the same button so technically you're not clicking the button again

content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
```diff
@@ -69,37 +69,39 @@ Registration
...
+   be automatically activated within a day.
 
    .. seealso::
       `Peppol endpoint - OpenPeppol eDEC Code Lists <https://docs.peppol.eu/edelivery/codelists/>`_
-      (open the "Participant Identifier Schemes" as HTML page)
+      (open the "Participant Identifier Schemes" as HTML page).
```
auva-odoo: Based on my next comment, I'd remove the indentation for this seealso

content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
```diff
@@ -69,37 +69,39 @@ Registration
...
+     documents.
 
-#. Click :guilabel:`Save`.
+     - Under :guilabel:`Peppol Reception Mode`, choose whether you want to receive Peppol documents
+       as bills (:guilabel:`Receive as Bills`), or as documents (:guilabel:`Receive as Documents`).
+     - Select an :guilabel:`Incoming Accounting Journal` to post the related journal entries.
+   - Select :guilabel:`Sending Only` if you wish to send Peppol documents only.
 
-All invoices and vendor bills can then be sent/received directly using Peppol.
+#. Click :guilabel:`Save`.
```
auva-odoo, line 108: 1) I'd make this a separate set of instructions (and therefore, a separate numbered list) since this isn't part of the registration process per say and this whole process is likely going to be a 2-part process anyways (i.e., people are probably not going to stay in the accounting settings while they wait for the registration to be activated)
2) For the Peppol Contact Email, the email defined on the company record is filled in by default, you don't need to enter it, hence why I suggest we change that step slightly. However, **can you please check what this field is used for exactly, and whether it can be changed freely, with no consequence on the Peppol side?**
3) The Incoming Accounting journal only applied to the Receive as Bills option; if you select Receive as Documents, you have to select a folder and tag; I added this in my suggestion
4) The Incoming Accounting Journal is filled in by default, and I assume selecting a journal is only relevant if you have multiple purchase journals, so I feel like we should revert to something similar to the previous version of this field description, which mentionned this aspect.
5) Given the above, I think it'd be easier and clearer to have one step per "field" (Peppol Participation Role and Peppol reception Mode)

Here's a suggestion for this section, taking the above into account:

```suggestion
Once Peppol is activated:

#. Go to :menuselection:`Accounting --> Configuration --> Settings` and scroll
down to the :guilabel:`Electronic Invoicing` section. 
#. Modify the :guilabel:`Peppol Contact Email` if needed.
#. Select the :guilabel:`Peppol Participation Role`:

   - :guilabel:`Sending & Receiving`: to both send *and* receive Peppol documents
   - :guilabel:`Sending Only`: to send Peppol documents only
    
#. If :guilabel:`Sending & Receiving` is selected as the :guilabel:`Peppol Participation Role`, define where Peppol documents should be received:

   - :guilabel:`Receive as Bills`: If multiple purchase journals are used, select the journal where the bills should be posted in the :guilabel:`Incoming Accounting Journal` field.
   - :guilabel:`Receive as Documents`: Select the :guilabel:`Document Folder` where the Peppol documents should be stored and the :guilabel:`Document Tags`to assign to them.

#. Click :guilabel:`Save`.
```

content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
```diff
@@ -69,37 +69,39 @@ Registration
...
 
 .. note::
-   - To update the :guilabel:`Primary contact email`, click :icon:`oi-arrow-right`
-     :guilabel:`Advanced Configuration`, modify it, and click :guilabel:`Save`.
+   - If you wish to :ref:`disconnect from Peppol <accounting/e-invoicing/peppol-deregister>`, click
+     :guilabel:`Disconnect Peppol`.
```
auva-odoo: - I feel like this should be a tip
- See my general comment about the Peppol deregistration from Odoo section; can that section be removed entirely since we have this note/tip here and the process is now pretty simple?

content/applications/finance/accounting/customer_invoices/electronic_invoicing.rst
```diff
@@ -69,37 +69,39 @@ Registration
...
+   - To update the :guilabel:`Peppol Contact Email`, modify it and click :guilabel:`Save`.
```
auva-odoo: I don't think this is necessary as it is already mentionned in the instructions?

auva-odoo (changes requested):
@tohor-odoo thank you for taking care of this task so fast. I feel like it might have been a bit too fast though 🙈, since a few things mentioned in the linked task seem to have gotten missed, including:

- Throughout the doc: PEPPOL Electronic Invoicing --> Electronic Invoicing
- Vendor bills reception in Documents section update (check my suggestion to clarify the instructions on lines 88-105 first though, it seems there could be some duplicate info)
- Peppol deregistration from Odoo section update/removal

Also, could you double-check whether the "Electronic invoicing with Peppol" section on the Belgium localization page is still accurate given these changes? I couldn't find the fields it mentions, but it might just be me.
I don't think other localization pages are affected, but it might be worth checking just in case. Thanks!

PR #20036, opened 2026-09-11
[IMP] l10n_fr: udpade EDI Tax return section

afma-odoo:
task-5403851


content/applications/finance/fiscal_localizations/france.rst
```diff
@@ -1043,44 +1043,48 @@ EDI tax return
 The EDI tax return (VAT report) can be submitted directly to |DGFiP| through Odoo. To do so, follow
 these steps:
 
+#. Make sure a valid VAT :guilabel:`ROF` (*Référence d’Obligation Fiscale*) is entered on the
+   :doc:`company record </applications/general/companies>`.
 #. :ref:`Review and validate <accounting/tax-returns/vat-return-review>` the tax return (VAT
```
auva-odoo: I know you didn't write this, but could we add the menuslection first? I had a bit of trouble finding where I needed to go in the first place since the ref sends directly to the "Review" process (which totally makes sense) and then I had to scroll up up up to try and find where I was supposed to be in the first place 🙈 

```suggestion
#. Go to :menuselection:`Accounting --> Accounting --> Tax returns`, then :ref:`review and validate <accounting/tax-returns/vat-return-review>` the tax return (VAT
```

content/applications/finance/fiscal_localizations/france.rst
```diff
@@ -1043,44 +1043,48 @@ EDI tax return
...
-   Fiscale) is entered.
-#. If needed, enable the :guilabel:`Add express mention` option and include a comment.
+#. Click :guilabel:`Submit`.
 #. - If payment to |DGFiP| is required:
 
-     - Click :guilabel:`Add a line` in the :guilabel:`IBAN` column to enter an IBAN account number
-       and BIC for the payment.
-     - In the :guilabel:`Amount` field, enter the total :guilabel:`VAT to pay`; the
-       :guilabel:`Amount to pay` field updates automatically.
+     - If needed, enable the :guilabel:`Partial amount` option and set the desired amount to pay.
```
auva-odoo:
```suggestion
     - If needed, enable :guilabel:`Partial amount` and set the desired amount to pay.
```
nitpick but makes it a bit shorter/more direct?

content/applications/finance/fiscal_localizations/france.rst
```diff
@@ -1043,44 +1043,48 @@ EDI tax return
...
+     - Update the :guilabel:`Bank Partner` account number if needed and make sure the corresponding
+       :guilabel:`BIC` reference is set.
 
    - If reimbursement from |DGFiP| is expected:
 
-     - Enable the :guilabel:`Add reimbursement comment` option and include a comment if needed.
-     - Click :guilabel:`Add the line` in the :guilabel:`IBAN` column to enter an IBAN account number
-       and BIC for the payment.
-     - Select the :guilabel:`Reimbursement type`:
+     - Enable the :guilabel:`Ask reimbursement` option to ask for the reimbursement of the
```
auva-odoo:
```suggestion
     - Enable :guilabel:`Ask reimbursement` to ask for the reimbursement of the
```
same as above, optional of course, but here since you're mentioning *enabling this option* already in the next sentence, I think it makes even more sense?

content/applications/finance/fiscal_localizations/france.rst
```diff
@@ -1043,44 +1043,48 @@ EDI tax return
...
+       :guilabel:`VAT Credit`. Proceeding without enabling this option will carry over the
+       reimbursement to the next period.
+
+       - Enable the :guilabel:`Partial amount` option and set the desired amount to be received.
```
auva-odoo: Is this indentation intended? It's different from the rest and it's the only one in this nested list?

Image

auva-odoo (changes requested):
Thank you for your work @afma-odoo! 🙂 I just have a few comments and also:

- I'd add a *the* before each DGFiP in the file I think
- And I know you didn't change this but I was wondering why the Reporting section is currently under Localization overview, and not Accounting? (I started wondering about it since we're also sending that to the DGFiP?) Don't look at it too closely though, I have a feeling this deserves its own task as it's one of those thing that might "expand" 🙈 (and also the potential change might be valid even before 19.4?) Let's discuss this, maybe?

Thank you!

afma-odoo:
@auva-odoo Thank you for this review!
You're right, it makes sense to move the reporting section to Accounting in the future.
And at the same time, as I'm writing it, I feel like it will be a little lost in there. 🤭🤔
Sure, let's discuss it!

auva-odoo (approved):
Thank you for the changes @afma-odoo! I'm going to approve this one as is and I suggest we take care of the reporting section in a separate task so we can 1) take the time to discuss whether it's a valid suggestion and 2) if it is, make that change from the earliest versin possible. Is that ok for you?

@robodoo delegate+

PR #20063, opened 2026-09-14
[IMP] pos: Download Invoice setting

kop-odoo:
task-6569775


content/applications/sales/point_of_sale/use/pos_invoices.rst
xpl-odoo: I know you didn't modify this, but this heading is a duplicate of the above heading.
So we should either: find a new heading or remove it :)

xpl-odoo (commented):
LGTM @kop-odoo !

I just have a suggestion not directly related to your changes :)

content/applications/sales/point_of_sale/use/pos_invoices.rst
kop-odoo: Yes, good idea! I ended up removing it and only keeping the `From the backend` to mirror the previous section about the frontend. This way, the `Consolidated customer invoice` is the only subheading of the backend, and I think the document flows better.

PR #20077, opened 2026-09-15
[IMP] on_premise: publisher warranty over https

mao-odoo:
Before this commit, opening the port 80 was presented as the only viable
solution to make the ping work. A semi frequent complain the support receives
is that the ping being constrained to a non https connection was unsafe,
and the customer were often told that there was nothing to do about it.

But an undocumented option exists.

opw-6537256


content/administration/on_premise.rst
```diff
@@ -67,6 +67,13 @@ To resolve the issue:
   These ports must be kept open even after registering a database, as the update notification runs
   once a week.
 
+  If your internal policies do not allow outbound connection on port 80,
+  you can alternatively open port 443 **AND** add the following line to your Odoo configuration file:
```
xpl-odoo:
```suggestion
  If your internal policies do not allow outbound connections on port `80`,
  you can alternatively open port `443` **and** add the following line to your Odoo configuration file:
```
I would put port numbers inside backticks as it's done just above

content/administration/on_premise.rst
```diff
@@ -67,6 +67,13 @@ To resolve the issue:
...
+
+.. code-block:: ini
+
+  publisher_warranty_url = https://services.odoo.com/publisher-warranty/
```
xpl-odoo:
```suggestion
   publisher_warranty_url = https://services.odoo.com/publisher-warranty/
```
(to respect the usual alignment)

xpl-odoo (changes requested):
Just some minor suggestions @mao-odoo !

content/administration/on_premise.rst
mao-odoo: good call, done for both suggestions

xpl-odoo:
@robodoo r+

